PR #18, opened 2021-03-18
Feature/m UI c 199 scenes

eigom:
Support for scene-based projects.

### DIFF
--- a/GliaWidgets/Coordinator/RootCoordinator.swift
+++ b/GliaWidgets/Coordinator/RootCoordinator.swift
@@ -20,7 +20,7 @@ class RootCoordinator: SubFlowCoordinator, FlowCoordinator {
     private let interactor: Interactor
     private let viewFactory: ViewFactory
     private weak var gliaDelegate: GliaDelegate?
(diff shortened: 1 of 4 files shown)


GliaWidgets/Coordinator/RootCoordinator.swift
```diff
@@ -20,6 +20,7 @@ class RootCoordinator: SubFlowCoordinator, FlowCoordinator {
     private let interactor: Interactor
     private let viewFactory: ViewFactory
     private weak var gliaDelegate: GliaDelegate?
+    private weak var gliaDataSource: GliaDataSource?
```
gersonnoboa, line 23: Why is it called data source? As far as I see, it's the window scene which is an UI element, so it doesn't deal with data.

gersonnoboa (approved):

GliaWidgets/Coordinator/RootCoordinator.swift
eigom: Not a good naming indeed, renamed to SceneProvider.

VladKupriienko (approved):

PR #25, opened 2021-03-31
File upload/download improvements & fixes

eigom:
- added limit to number of uploads
- added NSPhotoLibraryAddUsageDescription so saving to Photo Library won't crash the app
- minor fixes

### DIFF
--- a/GliaWidgets/Lib/Upload/FileUploader.swift
+++ b/GliaWidgets/Lib/Upload/FileUploader.swift
@@ -46,11 +46,19 @@ class FileUploader {
     }
     var count: Int { return uploads.count }
     let state = ObservableValue<State>(with: .idle)
+    let limitReached = ObservableValue<Bool>(with: false)
 
     private var uploads = [FileUpload]()
     private var storage = FileSystemStorage(directory: .documents)
+    private let maximumUploads: Int
 
-    func addUpload(with url: URL) -> FileUpload {
+    init(maximumUploads: Int) {
+        self.maximumUploads = maximumUploads
+        updateLimitReached()
+    }
+
+    func addUpload(with url: URL) -> FileUpload? {
+        guard !limitReached.value else { return nil }
         let localFile = LocalFile(with: url)
         let upload = FileUpload(with: localFile, storage: storage)
         upload.state.addObserver(self) { _, _ in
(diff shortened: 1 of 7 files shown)


TestingApp/Info.plist
```diff
@@ -24,6 +24,8 @@
 	<string>This app needs camera for audio/video calls and photo sending</string>
 	<key>NSMicrophoneUsageDescription</key>
 	<string>This app needs microphone for audio/video calls</string>
+	<key>NSPhotoLibraryAddUsageDescription</key>
+	<string>Save attachments</string>
```
VladKupriienko, line 28: Maybe we should make a little bit more verbose? e.g. "This app needs access to photo library to save attachments" or smth like this.

GliaWidgets/Lib/Upload/FileUploader.swift
```diff
@@ -104,4 +116,8 @@ class FileUploader {
             state.value = newState
         }
     }
+
+    private func updateLimitReached() {
+        limitReached.value = count == maximumUploads
```
VladKupriienko, line 121: Is there a guarantee that there will be no situation when the counter jumps by 2, 3, etc? In that case, it will continue to be false. Maybe a `>=` would be more secure. Or maybe I'm just paranoid and there is no way that would happen 😀

VladKupriienko (approved):

TestingApp/Info.plist
VladKupriienko: Ah, sorry, you are right. Didn't notice that it's from the testing app.

GliaWidgets/Lib/Upload/FileUploader.swift
eigom: Both `addUpload()` methods have a guard to prevent that.

gersonnoboa: I cannot see a way right now of it happening but I would say that the `>=` suggestion makes sense regardless, just in case.

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -69,7 +70,7 @@ class ChatViewModel: EngagementViewModel, ViewModel {
     private var unreadMessages: UnreadMessagesHandler!
     private let showsCallBubble: Bool
     private let storage = ChatStorage()
-    private let uploader = FileUploader()
+    private let uploader = FileUploader(maximumUploads: 25)
```
gersonnoboa, line 73: I guess the requirements don't mention that this needs to be changed from the integrator side if possible, correct? If there's no mention of that, then this is fine.

gersonnoboa (approved):
Something additional, could you capitalize the `l` in `Photo library`? Thanks!

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
eigom: Correct.

PR #28, opened 2021-04-08
Feature/m UI c 236 fixes

eigom:
- Stretched popover fix for iOS 12
- Audio/video indicator fix
- Out of memory crash fix
- Added audio background mode

### DIFF
--- a/GliaWidgets/Lib/File/LocalFile.swift
+++ b/GliaWidgets/Lib/File/LocalFile.swift
@@ -31,7 +31,7 @@ class LocalFile {
     let url: URL
 
     private var thumbnail: UIImage?
-    private static var thumbnailQueue: OperationQueue = {
+    private static let thumbnailQueue: OperationQueue = {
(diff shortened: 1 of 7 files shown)


GliaWidgets/Lib/File/LocalFile.swift
```diff
@@ -31,6 +31,11 @@ class LocalFile {
     let url: URL
 
     private var thumbnail: UIImage?
+    private static var thumbnailQueue: OperationQueue = {
+        var queue = OperationQueue()
```
gersonnoboa, line 35: It can be let, no?

gersonnoboa (approved):

GliaWidgets/Lib/File/LocalFile.swift
eigom: Indeed.

VladKupriienko (approved):

PR #31, opened 2021-04-15
[MUIC-201] [MUIC-289] Fix video/audio call landscape mode

VladKupriienko:
Fix header and button bar disappearing from call view after returning to portrait mode from landscape

As always, some code style refactor included ;)

### DIFF
--- a/GliaWidgets/Component/Bar/CallButtonBar/CallButtonBar.swift
+++ b/GliaWidgets/Component/Bar/CallButtonBar/CallButtonBar.swift
@@ -24,9 +24,9 @@ class CallButtonBar: UIView {
     private let style: CallButtonBarStyle
     private let stackView = UIStackView()
     private var buttons = [CallButton]()
-    private var effectView = UIVisualEffectView(effect: UIBlurEffect(style: .prominent))
+    private var effectView = UIVisualEffectView(effect: UIBlurEffect(style: .dark))
     private var bottomSpaceConstraint: NSLayoutConstraint!
-    private var kInsets = UIEdgeInsets(top: 3.0, left: 3.0, bottom: 3.0, right: 3.0)
+    private var kInsets = UIEdgeInsets(top: 7.0, left: 3.0, bottom: 7.0, right: 3.0)
 
     init(with style: CallButtonBarStyle) {
         self.style = style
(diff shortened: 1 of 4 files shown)


gersonnoboa:
I will check this on Monday

gersonnoboa (commented):
Shouldn't the default state when in landscape be with the controls on? Right now if I switch to landscape, I have to tap to see the controls. And then if I tap again, the controls never hide anymore.

Actually, is there a reason they hide? They could just change the position when in landscape and not need a tap to be shown.

VladKupriienko:
> Shouldn't the default state when in landscape be with the controls on? Right now if I switch to landscape, I have to tap to see the controls. And then if I tap again, the controls never hide anymore.
> 
> Actually, is there a reason they hide? They could just change the position when in landscape and not need a tap to be shown.

I asked about the expected behaviour from Raigo and he said this is the one. Should we ask for some clarification from Andreas/Rasmus?

gersonnoboa:
@VladKupriienko I would say so, because changing to landscape looks... Broken, I would say.

VladKupriienko:
I have also committed a fix for the wrong blur color in the landscape mode for video calls [MUIC-289].

As a small bonus, I've made the bottom button bar design actually follow the one in Figma (buttons were spread out all over the screen in landscape previously).

Also, I've applied the changes to the call view behavior in landscape mode that we discussed together:

- Audio call now **always** shows the buttons and the header, both in landscape and portrait.
- Video call hides the controls on rotation to landscape. Tapping anywhere on the screen will bring them back for 3.2 seconds (was 3.0 in the code. Not really that noticeable, but if we agreed on 3200 let's do 3200).

GliaWidgets/Component/Bar/CallButtonBar/CallButtonBar.swift
```diff
@@ -24,26 +24,40 @@ class CallButtonBar: UIView {
...
+    func adjustStackConstraints() {
+        if currentOrientation.isPortrait {
+            verticalAlignConstrait.autoRemove()
+            stackView.spacing = 2
+            leftConstraint.autoInstall()
+            rightConstraint.autoInstall()
+        } else {
+            leftConstraint.autoRemove()
+            rightConstraint.autoRemove()
+            stackView.spacing = 12
```
gersonnoboa, line 56: Do these have some kind of descriptive naming that you think might explain these spacings better?

gersonnoboa (approved):
If I recall correctly, the video sometimes not stretching correctly when on landscape was a SDK issue, so I will approve.

PR #33, opened 2021-04-23
[MUIC-300] Fix placeholder image appearing through transparent operator avatar

VladKupriienko:


GliaWidgets/Component/ImageView/User/UserImageView.swift
```diff
@@ -24,22 +25,37 @@ class UserImageView: UIView {
     }
 
     func setImage(_ image: UIImage?, animated: Bool) {
+        if image != nil {
```
gersonnoboa, line 28: What about when the image is not nil? Should this be flipped then?

VladKupriienko: I can't think of a possible scenario where this would be called twice, with `nil` the second time, but one can never be too careful. Thanks for noticing :)

VladKupriienko: Updated the PR. I've played around with different wordings for this function and stopped on this one because I wanted parameter name to be present and named clearly.
...anyway it's not public API and just a small private func in a component 👀

gersonnoboa (approved):

PR #37, opened 2021-05-05
[MUIC-279] New message indicator

VladKupriienko:
Implements a new message indicator in the chat if the visitor is not at the bottom of the chat area when a new message is received.

As usual, includes code style refactor in the affected classes.


GliaWidgets/Coordinator/RootCoordinator.swift
```diff
@@ -1,5 +1,5 @@
-import UIKit
 import SalemoveSDK
+import UIKit
```
gersonnoboa, line 2: How do you order them? Alphabetically?

GliaWidgets/Glia.swift
```diff
@@ -16,7 +16,7 @@ public enum GliaEvent {
     case maximized
 }
 
-public protocol SceneProvider: class {
+public protocol SceneProvider: AnyObject {
```
gersonnoboa, line 19: Good, I was going to mention this at some point cause I saw the warning. Thanks for fixing it.

gersonnoboa (commented):
I don't see the operator is typing indicator on my device. I type and nothing happens on the device until I press send, where the device receives the phone. Also, very importantly, the operator does not see that the visitor is typing.

GliaWidgets/Coordinator/RootCoordinator.swift
VladKupriienko: Ah, this was actually SwiftFormat who moved them around and I didn't notice. [Alphabetically indeed.](https://github.com/nicklockwood/SwiftFormat/blob/master/Rules.md#sortedImports)

GliaWidgets/Glia.swift
VladKupriienko: > I don't see the operator is typing indicator on my device. I type and nothing happens on the device until I press send, where the device receives the phone. Also, very importantly, the operator does not see that the visitor is typing.

Indicating that the operator is typing is out of the scope of this PR. This PR only adds a raindrop-like indicator to the chat to indicate if any new messages have appeared while the visitor was scrolling through the chat (i.e. could not see them right away) - [[MUIC-279]](https://salemove.atlassian.net/jira/software/projects/MUIC/boards/132?label=iOS&selectedIssue=MUIC-279)

VladKupriienko:
> I don't see the operator is typing indicator on my device. I type and nothing happens on the device until I press send, where the device receives the phone. Also, very importantly, the operator does not see that the visitor is typing.

Indicating that the operator is typing is out of the scope of this PR. This PR only adds a raindrop-like indicator to the chat to indicate if any new messages have appeared while the visitor was scrolling through the chat (i.e. could not see them right away) - [[MUIC-279]](https://salemove.atlassian.net/jira/software/projects/MUIC/boards/132?label=iOS&selectedIssue=MUIC-279)

As for the fact that the operator does not see the visitor's message - strange, I don't think I encountered that. I will double-check it right away.

gersonnoboa:
No I meant that the operator cannot see that the visitor is typing, but the message arrives fine. I don't know if that has been implemented in the widgets yet or not. If you see our Example app, when you are typing something from the app, you can see it on the Glia Hub as an operator, even before the visitor has pressed send on the device. That is what I meant.

VladKupriienko:
I see. This has always been the case for widgets as well (all the visitor messages were seen in the Hub before even being sent), it shouldn't have been removed by this PR.

Just checked, for me it works as expected.
Screenshot 2021-05-07 at 13 18 18

gersonnoboa (approved):
Interesting, anyways I will approve because the new message indicator functionality works.

PR #38, opened 2021-05-07
[MUIC-304] Fix keyboard disappearing on operator's message received

VladKupriienko:
Previously the app resigned the first responder from the message entry view on any new message (oversight from choice cards PR). 
Now it does so **only** if the received message is a choice card (to deactivate the message entry until it is answered).

### DIFF
--- a/GliaWidgets/View/Chat/Entry/ChatMessageEntryView.swift
+++ b/GliaWidgets/View/Chat/Entry/ChatMessageEntryView.swift
@@ -14,19 +14,24 @@ public class ChatMessageEntryView: UIView {
             updateTextViewHeight()
         }
     }
+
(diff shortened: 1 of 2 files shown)


gersonnoboa:
I will test this later in the evening

VladKupriienko:
Interesting, thanks for pointing that out. I think I also experienced that once, but couldn't figure how I did it. Will look into that on Tuesday.

kuldar-daniel:
@gersonnoboa  seems like the table view content offset was set incorrectly when keyboard is shown. Added function to scroll to bottom instead once keyboard is visible (I beleive that was the original intent).

gersonnoboa (approved):
I will approve because it solves the intended issue, but I will propose an enhancement in the chat.

PR #42, opened 2021-05-12
MUIC-310: Message sending logic is wrong

kuldar-daniel:
Previously the text was cleared only after the response was received (either success or error). Now we clear the text before sending the message request.


gersonnoboa (approved):
I don't really like the name `messageTextTemp` but it is what it is.

kuldar-daniel:
Maybe in the future we could refactor this a bit to include the message being send in the function parameters.
It needs more refactoring due to the validation would be run on the class variable etc.

PR #47, opened 2021-05-17
[MUIC-307] Show Delivered status only for last message

VladKupriienko:


VladKupriienko:
I should probably add that there is also a small refactor included for the affected files. Furthermore, by taking `animated` into consideration for `reloadSection` I removed the jittering that occurred often while loading the messages.

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -362,12 +362,24 @@ extension ChatViewModel {
...
-        let message = ChatMessage(with: message)
-        let item = ChatItem(kind: .visitorMessage(message, status: status))
-        downloader.addDownloads(for: message.attachment?.files, with: uploads)
+        let deliveredStatus = Strings.Message.Status.delivered
+
+        // Remove previous "Delivered" statuses
+        section.items
+            .enumerated()
+            .forEach { index, element in
+                if case .visitorMessage(let message, let status) = element.kind,
```
henrikix, line 371: `if case let .visitorMessage(message, status) = element.kind` perhaps?

GliaWidgets/View/Chat/ChatView.swift
```diff
@@ -101,8 +101,14 @@ class ChatView: EngagementView {
         updateTableView(animated: animated)
     }
 
-    func refreshSection(_ section: Int) {
-        tableView.reloadSections([section], with: .fade)
+    func refreshSection(_ section: Int, animated: Bool) {
```
henrikix, line 104: I didn't notice any usage of `animated: true` - did I miss something?

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -362,12 +362,24 @@ extension ChatViewModel {
...
+                    section.replaceItem(at: index, with: chatItem)
+                }
+            }
+
+        let deliveredMessage = ChatMessage(with: message)
+        let item = ChatItem(kind: .visitorMessage(deliveredMessage, status: deliveredStatus))
+        downloader.addDownloads(for: deliveredMessage.attachment?.files, with: uploads)
         section.replaceItem(at: index, with: item)
-        action?(.refreshRow(index, in: section.index, animated: false))
+        action?(.refreshSection(section.index, animated: false))
```
henrikix, line 382: Have you considered just reloading the rows that need to be updated?

VladKupriienko: That was the initial solution, yes. However, the animation behaved very weirdly for some reason, that's why I went with the entire section.

GliaWidgets/View/Chat/ChatView.swift
VladKupriienko: Well, the initial implementation of this method behaved as if `animated` was set to `true`. That's why I left in the possibility to do so. But you're correct, currently, it is unused. Do you think I should just remove it altogether?

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
VladKupriienko: AFAIK there's no difference between the two. Your suggestion is a bit shorter, but I've just been following the code style of the project.

henrikix: I was just suggesting since the shorter length, yep. If thats the general code style then all good.

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
henrikix: The previous solution used `func refreshRow(_ row: Int, in section: Int, animated: Bool)` which in turn uses
`tableView.beginUpdates()  tableView.endUpdates()` but does not call `reloadRows` which seems odd

GliaWidgets/View/Chat/ChatView.swift
henrikix: I think you can currently remove it, yes.

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
VladKupriienko: I've changed it to only refresh the affected rows, actually calling the `reloadRows(at:with:)` between starting and ending the updates.

GliaWidgets/View/Chat/ChatView.swift
VladKupriienko: `animated` was removed. The function now always behaves as if it was set to `false`.
Couldn't think of a good use case for animated section reload, so I agree with you that it is not needed.

henrikix (approved):

PR #52, opened 2021-05-26
[MUIC-279] [MUIC-281] New message indicator

VladKupriienko:
This pull request is about 90% identical to the [previous one for this feature](https://github.com/salemove/ios-sdk-widgets/pull/37).

It had to be created again due to a mess in releases and rollback of the said pull request. 

Some things have changed, though; it was previously only reviewed by Gerson, so a fresh look wouldn't hurt.


GliaWidgets/ViewModel/Chat/Unread/UnreadMessagesHandler.swift
```diff
@@ -2,34 +2,47 @@ class UnreadMessagesHandler {
...
         self.isViewVisible = isViewVisible
+        self.isChatScrolledDown = isChatScrolledToBottom
         isWindowVisible.addObserver(self) { _, _ in
             self.checkVisible()
         }
         isViewVisible.addObserver(self) { _, _ in
             self.checkVisible()
         }
+        isChatScrolledDown.addObserver(self) { _, _ in
+            self.checkVisible()
```
dukhovnyi, line 24: I'd like to propose to avoid memory leaks and use weak references to self here and on lines 18, 21.

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -100,10 +103,14 @@ class ChatViewModel: EngagementViewModel, ViewModel {
         self.showsCallBubble = showsCallBubble
         self.startAction = startAction
         super.init(interactor: interactor, alertConfiguration: alertConfiguration)
+        unreadMessages.addObserver(self) { unreadCount, _ in
+            self.action?(.updateNewMessageIndicator(itemCount: unreadCount))
```
dukhovnyi, line 107: Let's avoid memory leak and use weak self here.

dukhovnyi (changes requested):

GliaWidgets/View/Chat/ChatView.swift
```diff
@@ -338,4 +360,13 @@ extension ChatView: UITableViewDelegate {
     public func scrollViewWillBeginDragging(_ scrollView: UIScrollView) {
         endEditing(true)
     }
+
+    func scrollViewDidScroll(_ scrollView: UIScrollView) {
+        let chatEnd = scrollView.contentSize.height - scrollView.frame.size.height
+        if scrollView.contentOffset.y + 10 >= chatEnd {
```
henrikix, line 366: Should probably use `if scrollView.contentOffset.y + scrollView.contentInset.top >= chatEnd`
Would also rename `chatEnd` to something like `chatBottomOffset`

GliaWidgets/View/Chat/ChatView.swift
```diff
@@ -3,21 +3,26 @@ import UIKit
...
+    let newMessageIndicatorView: NewMessageIndicatorView
     var numberOfSections: (() -> Int?)?
     var numberOfRows: ((Int) -> Int?)?
     var itemForRow: ((Int, Int) -> ChatItem?)?
     var fileTapped: ((LocalFile) -> Void)?
     var downloadTapped: ((FileDownload) -> Void)?
     var callBubbleTapped: (() -> Void)?
     var choiceOptionSelected: ((ChatChoiceCardOption, String) -> Void)!
+    var newMessageIndicatorTapped: (() -> Void)?
+    var chatScrolledToTheEnd: ((Bool) -> Void)?
```
henrikix, line 15: `chatScrolledToBottom` or even just `scrolledToBottom` since this is already the `ChatView`?

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -42,6 +43,7 @@ class ChatViewModel: EngagementViewModel, ViewModel {
             declined: () -> Void
         )
         case showCallBubble(imageUrl: String?)
+        case updateNewMessageIndicator(itemCount: Int)
```
henrikix, line 46: Would rename to `updateUnreadMessageCount(Int)`

GliaWidgets/ViewModel/Chat/Unread/UnreadMessagesHandler.swift
```diff
@@ -2,34 +2,47 @@ class UnreadMessagesHandler {
     private let unreadMessages: ObservableValue<Int>
     private let isWindowVisible: ObservableValue<Bool>
     private let isViewVisible: ObservableValue<Bool>
+    private let isChatScrolledDown: ObservableValue<Bool>
```
henrikix, line 5: Here we have a third name for the same thing. So you have `scrolledToEnd`, `scrolledToBottom` and `scrolledDown`. I would recommend using `scrolledToBottom` everywhere

GliaWidgets/Resources/Asset.swift
```diff
@@ -32,6 +32,7 @@ public enum Asset {
   public static let uploadRemove = ImageAsset(name: "uploadRemove")
   public static let chatPickMedia = ImageAsset(name: "chatPickMedia")
   public static let chatSend = ImageAsset(name: "chatSend")
+  public static let newMessageIndicator = ImageAsset(name: "newMessageIndicator")
```
henrikix, line 35: `unreadMessageIndicator`?

GliaWidgets/Component/NewMessageIndicator/NewMessageIndicatorView.swift
```diff
@@ -0,0 +1,75 @@
+import UIKit
+
+final class NewMessageIndicatorView: View {
```
henrikix, line 3: `UnreadMessageIndicatorView`?

GliaWidgets/Component/NewMessageIndicator/NewMessageIndicatorStyle.swift
```diff
@@ -0,0 +1,28 @@
+import UIKit
+
+public struct NewMessageIndicatorStyle {
```
henrikix, line 3: `UnreadMessageIndicatorStyle`?

GliaWidgets/View/Chat/ChatStyle.swift
```diff
@@ -12,24 +12,28 @@ public class ChatStyle: EngagementStyle {
     public var videoUpgrade: ChatCallUpgradeStyle
     public var callBubble: BubbleStyle
     public var pickMedia: ItemListStyle
+    public var newMessageIndicator: NewMessageIndicatorStyle
```
henrikix, line 15: `unreadMessageIndicator`?

GliaWidgets/View/Chat/ChatView.swift
```diff
@@ -3,21 +3,26 @@ import UIKit
 class ChatView: EngagementView {
     let tableView = UITableView()
     let messageEntryView: ChatMessageEntryView
+    let newMessageIndicatorView: NewMessageIndicatorView
```
henrikix, line 6: `unreadMessageIndicatorView`?

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -13,6 +13,7 @@ class ChatViewModel: EngagementViewModel, ViewModel {
         case fileTapped(LocalFile)
         case downloadTapped(FileDownload)
         case choiceOptionSelected(ChatChoiceCardOption, String)
+        case chatScrolled(Bool)
```
henrikix, line 16: This is a bit confusing imo since you're going from `view.chatScrolledToTheEnd` to just `chatScrolled`. Im starting to think `chatScrolledToEnd` (or `chatScrolledToBottom`) should also be renamed to `chatScrolled(reachedBottom: Bool)` everywhere so it reflects what it actually is more clearly.

henrikix (changes requested):

GliaWidgets/Component/NewMessageIndicator/NewMessageIndicatorView.swift
```diff
@@ -0,0 +1,75 @@
...
+
+    private let backgroundView = UIImageView()
+    private let userImageView: UserImageView
+    private let badgeView: BadgeView
+
+    private let kSize = CGSize(width: 58, height: 66) // Shadows included
+    private let kUserImageSize = CGSize(width: 36, height: 36)
+    private let kBadgeInsets = UIEdgeInsets(top: 0, left: 0, bottom: 0, right: 5)
+
+    public init(with style: NewMessageIndicatorStyle) {
```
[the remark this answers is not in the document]
VladKupriienko: Every other custom view in the project uses the `init(with style:)` naming. Changing it here would mean changing it everywhere else.

GliaWidgets/Component/NewMessageIndicator/NewMessageIndicatorView.swift
```diff
@@ -0,0 +1,75 @@
...
+            if newItemCount <= 0 {
+                isHidden = true
+            } else {
+                isHidden = false
+                badgeView.newItemCount = newItemCount
+            }
+        }
+    }
+
+    var tap: (() -> Void)?
```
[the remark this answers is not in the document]
VladKupriienko: Used `onTap` for the objc func and `tapped` for this var. Otherwise why would be both called `tapped`.

GliaWidgets/View/Chat/ChatStyle.swift
```diff
@@ -12,24 +12,28 @@ public class ChatStyle: EngagementStyle {
...
-                closeButton: HeaderButtonStyle,
-                visitorMessage: VisitorChatMessageStyle,
-                operatorMessage: OperatorChatMessageStyle,
-                choiceCard: ChoiceCardStyle,
-                messageEntry: ChatMessageEntryStyle,
-                audioUpgrade: ChatCallUpgradeStyle,
-                videoUpgrade: ChatCallUpgradeStyle,
-                callBubble: BubbleStyle,
-                pickMedia: ItemListStyle) {
+    public init(
```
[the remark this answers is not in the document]
VladKupriienko: 🤷‍♂️ 
Common thing in the Styles part of the project.

GliaWidgets/View/Chat/ChatView.swift
VladKupriienko: Good point about taking `contentInset` into consideration, haven't thought of that. Thanks

VladKupriienko: I'd leave `+10` though as well so it detects not just pixel-perfect scrolling to the bottom, but even a very close one.

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
VladKupriienko: `unreadMessages` refers not only to the new message indicator though but also to a minimized call bubble (if you receive a message while a call is minimized, a small badge with a number is shown). This specific `action` is aimed towards a view specifically and only updates the new message indicator.

Maybe `updateNewMessageIndicator(count: Int)` or `updateNewMessageIndicatorCount(Int)` instead?

GliaWidgets/ViewModel/Chat/Unread/UnreadMessagesHandler.swift
VladKupriienko: Agreed, IDK why it's so inconsistent in naming 😅

GliaWidgets/Resources/Asset.swift
VladKupriienko: I just used the name with which we always refer to that thingy. But this does sound better and has a closer meaning.

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
VladKupriienko: UPD: taking into consideration the renaming of the entire thing into an `unreadMessageIndicator` I suggest

`updateUnreadMessageIndicator(count: Int)` or `updateUnreadMessageIndicatorCount(Int)`, because it still only affects the view and not the unread messages observer / call bubble.

GliaWidgets/View/Chat/ChatView.swift
VladKupriienko: Renamed as per last suggestion

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
VladKupriienko: While `chatScrolled(reachedBottom:)` looks and sounds good, it doesn't really work everywhere sadly.

In `UnreadMessagesHandler` it should still be called `isChatScrolledToBottom` because it refers not to an event, but just the boolean value of the result.

In `ChatView` it is a function variable thus parameters cannot be named. Without that semantics are lost (`chatScrolledToBottom((Bool) -> Void)?` vs `chatScrolled((Bool) -> Void)?`

VladKupriienko:
I've pushed a commit with most of the suggestions taken into consideration. Some of the conversations are still open with provided reasoning.

GliaWidgets/Component/NewMessageIndicator/UnreadMessageIndicatorView.swift
```diff
@@ -0,0 +1,75 @@
...
+
+    private let backgroundView = UIImageView()
+    private let userImageView: UserImageView
+    private let badgeView: BadgeView
+
+    private let kSize = CGSize(width: 58, height: 66) // Shadows included
+    private let kUserImageSize = CGSize(width: 36, height: 36)
+    private let kBadgeInsets = UIEdgeInsets(top: 0, left: 0, bottom: 0, right: 5)
+
+    public init(with style: UnreadMessageIndicatorStyle) {
```
dukhovnyi, line 27: Since the `UnreadMessageIndicatorView` is internal, do we need to have this initializer as public?

dukhovnyi (approved):
Looks good, just a one small suggestion :)

GliaWidgets/Component/NewMessageIndicator/UnreadMessageIndicatorView.swift
VladKupriienko: Must've copied from somewhere and didn't notice. Thanks!

henrikix (approved):

PR #55, opened 2021-06-03
MUIC-314: Chat -Tapping on the screen will hide the keyboard

kuldar-daniel:
Addresses issues where there would be no way to dismiss the keyboard in the Chat view. Main concern being that when user wants to view previous messages, there would not be that much space left to view the messages once the focus is set on input field.

**What is done**:
* Add tap gesture to view to hide the keyboard when user taps tableview. Also make sure that taps go through to tableview views too, eg. opening a photo or selection a choice card option
* Add keyboard dismissal to tableview on drag

### DIFF
--- a/GliaWidgets/View/Chat/ChatView.swift
+++ b/GliaWidgets/View/Chat/ChatView.swift
@@ -146,6 +146,7 @@ class ChatView: EngagementView {
         tableView.estimatedRowHeight = UITableView.automaticDimension
         tableView.separatorStyle = .none
         tableView.contentInset = kChatTableViewInsets
+        tableView.keyboardDismissMode = .onDrag
         tableView.register(cell: ChatItemCell.self)
 
         unreadMessageIndicatorView.tapped = { [weak self] in
@@ -153,6 +154,7 @@ class ChatView: EngagementView {
         }
 
         observeKeyboard()
+        addTapGesture()
     }
 
     private func layout() {
@@ -183,6 +185,24 @@ class ChatView: EngagementView {
             of: messageEntryView,
             withOffset: kUnreadMessageIndicatorInset
         )
+    }
+
+    private func addTapGesture() {
+        let tapGesture = UITapGestureRecognizer(
+            target: self,
+            action: #selector(dismissKeyboard)
+        )
+
+        tapGesture.cancelsTouchesInView = false
+
+        addGestureRecognizer(tapGesture)
+    }
+    
+    @objc
+    private func dismissKeyboard(sender: UITapGestureRecognizer) {
+        if sender.state == .ended {
+            endEditing(true)
+        }
     }
 
     private func content(for item: ChatItem) -> ChatItemCell.Content {
--- a/GliaWidgets/View/Chat/ChatView.swift
+++ b/GliaWidgets/View/Chat/ChatView.swift
@@ -146,7 +146,6 @@ class ChatView: EngagementView {
         tableView.estimatedRowHeight = UITableView.automaticDimension
         tableView.separatorStyle = .none
         tableView.contentInset = kChatTableViewInsets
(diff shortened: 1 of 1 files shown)


GliaWidgets/View/Chat/ChatView.swift
```diff
@@ -146,13 +146,15 @@ class ChatView: EngagementView {
...
         tableView.contentInset = kChatTableViewInsets
+        tableView.keyboardDismissMode = .onDrag
         tableView.register(cell: ChatItemCell.self)
 
         unreadMessageIndicatorView.tapped = { [weak self] in
             self?.scrollToBottom(animated: true)
         }
 
         observeKeyboard()
+        addTapGesture()
```
VladKupriienko, line 157: Maybe we could give this function a more meaningful name, one that includes the purpose of that tap gesture? E.g. `addKeyboardDismissalTapGesture()`. It's a mouthful, but it's very descriptive 😄

GliaWidgets/View/Chat/ChatView.swift
```diff
@@ -146,13 +146,15 @@ class ChatView: EngagementView {
         tableView.estimatedRowHeight = UITableView.automaticDimension
         tableView.separatorStyle = .none
         tableView.contentInset = kChatTableViewInsets
+        tableView.keyboardDismissMode = .onDrag
```
VladKupriienko, line 149: Hm, we already have such functionality achieved through implementing `scrollViewWillBeginDragging(_ scrollView: UIScrollView)` from `UITableViewDelegate` (lines 379-381).

I tried removing that delegate method and leaving just this new line: it works, but the animation looks a bit strange (suggestions strip stays for some time). Just like in [this SO question](https://stackoverflow.com/questions/66180854/why-the-animation-of-keyboarddismissmode-is-weird). Funnily enough, the answer proposes to do what we are already doing 🤷

VladKupriienko (changes requested):
I added my 5 cents, but the biggest issue with this PR is the target branch 😄 
From `feature/MUIC-XXX` we first merge into `qa/MUIC-XXX`, where Eve tests it, and when she moves it to Done, we merge it to `dev` (without the review, if nothing changed).

GliaWidgets/View/Chat/ChatView.swift
```diff
@@ -185,6 +187,24 @@ class ChatView: EngagementView {
...
+        )
+
+        tapGesture.cancelsTouchesInView = false
+
+        addGestureRecognizer(tapGesture)
+    }
+    
+    @objc
+    private func dismissKeyboard(sender: UITapGestureRecognizer) {
+        if sender.state == .ended {
```
dukhovnyi, line 203: do we need to handle not only `ended` but `cancelled` and `failed` cases also?

kuldar-daniel: Not really. Since UITapGesture is discrete gesture, it really doesn't trigger for any other state.
Considering that, there is really no need to specifically check for end state either, however for some reason (maybe future-proofing?) Apple's documentation example suggest we check the state: https://developer.apple.com/documentation/uikit/uitapgesturerecognizer

GliaWidgets/View/Chat/ChatView.swift
kuldar-daniel: Good catch, will remove :)

GliaWidgets/View/Chat/ChatView.swift
kuldar-daniel: Changed, thanks!

dukhovnyi (approved):

VladKupriienko (approved):

VladKupriienko:
Terribly sorry it took so long to re-approve 🙇 Hadn't noticed it on Fri and was very busy on Monday

PR #56, opened 2021-06-09
[MUIC-306] Start Glia in app's main window

VladKupriienko:
Previously, Glia would launch in it's own `UIWindow`. This was quite helpful for a number of reasons, but also introduced a very unwanted behavior. ReplayKit only records the main window of the app, therefore Glia's views were not visible during screen sharing.

For the time being, until the SDK migrates from ReplayKit to Broadcast Upload, this workaround is introduced, which uses `GliaViewController` instead of `GliaWindow` to encapsulate all views.


VladKupriienko:
Ah, forgot to point out something
I was actually working on custom `UIViewController` transitioning animations, but they just won't work properly.
So in order to give the fix to the client ASAP, I temporarily replaced them with the standard `.crossDissolve`. Looks fine IMO, but not as good as it should 😄

GliaWidgets/Coordinator/RootCoordinator.swift
```diff
@@ -219,72 +226,17 @@ class RootCoordinator: SubFlowCoordinator, FlowCoordinator {
...
-                    bubbleView: bubbleView,
-                    delegate: self
-                )
-            }
-        } else {
-            return GliaWindow(
-                bubbleView: bubbleView,
-                delegate: self
-            )
+    private func presentWindow(animated: Bool, completion: (() -> Void)? = nil) {
```
kuldar-daniel, line 229: we're not really presenting anything on window here, maybe change the function name?

GliaWidgets/Coordinator/RootCoordinator.swift
```diff
@@ -219,72 +226,17 @@ class RootCoordinator: SubFlowCoordinator, FlowCoordinator {
...
-    private func windowScene() -> UIWindowScene? {
-        if let sceneProvider = sceneProvider {
-            return sceneProvider.windowScene()
-        } else {
-            let scene = UIApplication.shared
-                .connectedScenes
-                .filter { $0.activationState == .foregroundActive }
-                .first
-            return scene as? UIWindowScene
+    private func dismissWindow(animated: Bool, completion: (() -> Void)? = nil) {
```
kuldar-daniel, line 236: Same as above

GliaWidgets/Window/GliaViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+    var bubbleKind: BubbleKind = .userImage(url: nil) {
+        didSet { bubbleWindow?.bubbleKind = bubbleKind }
+    }
+
+    private enum State {
+        case maximized
+        case minimized
+    }
+
+    private var state: State = .maximized
```
kuldar-daniel, line 22: why do we need to retain the state? It's not being passed anywhere

kuldar-daniel (approved):
I'm not huge fan of the "topMostViewController" part of it but I can't really suggest something better from top of my head either, and since this is critical, I'll approve it anyway.

GliaWidgets/Coordinator/RootCoordinator.swift
```diff
@@ -36,7 +36,11 @@ class RootCoordinator: SubFlowCoordinator, FlowCoordinator {
     private let isWindowVisible = ObservableValue<Bool>(with: false)
     private let navigationController = NavigationController()
     private let navigationPresenter: NavigationPresenter
-    private var window: GliaWindow?
+    private let presentingWindow = UIApplication.shared.windows.filter { $0.isKeyWindow }.first
+    private var presentingViewController: UIViewController? {
+        return presentingWindow?.topMostViewController()
+    }
+    private var gliaViewController: GliaViewController!
```
dukhovnyi, line 43: Using implicitly unwrapped optional could make a side effect and emits crash, so I'd like to propose to avoid using it (if possible). Probably we could use optional or lazy variable instead. Does it make sense?

GliaWidgets/Window/GliaViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+    private weak var delegate: GliaViewControllerDelegate?
+    private var gliaNavigationController: UINavigationController!
```
dukhovnyi, line 24: Let's avoid using implicitly unwrapped optional (if possible). Probably we could use optional or lazy variable instead.
Looks like we are using `gliaNavigationController` just to retain the `navigationController`, I could be wrong, but does `addChild` doesn't retain it? If so, we could delete this var at all.

GliaWidgets/Window/GliaViewController.swift
```diff
@@ -0,0 +1,107 @@
...
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
+
+    func insertNavigationController(_ navigationController: UINavigationController) {
+        self.gliaNavigationController = navigationController
+        addChild(navigationController)
+        view.addSubview(navigationController.view)
+        navigationController.didMove(toParent: self)
+    }
```
dukhovnyi, line 50: Probably we could extend `UIViewController` class and move this method to separated extension file like this:

```suggestion
    func add(_ viewController: UIViewController) {
        viewController.willMove(toParent: self)
        addChild(viewController)
        view.addSubview(viewController.view)
        viewController.didMove(toParent: self)
    }
```

One more question about layout, does it use resizing mask or layout constraints?

dukhovnyi (changes requested):
Since the framework doesn't use a window at all, let's remove `window` word from the code and rename the folder `GliaWidgets/**Window**`.

GliaWidgets/Coordinator/RootCoordinator.swift
VladKupriienko: Good catch. Leftover name

GliaWidgets/Window/GliaViewController.swift
VladKupriienko: Most possibly also a leftover which I forgot to clean up

VladKupriienko: Interestingly enough, there doesn't seem to be a use for that field in the earlier versions of the code either. Idk what's up with that

GliaWidgets/Window/GliaViewController.swift
VladKupriienko: I think indeed the best way would be to remove it, you're right

GliaWidgets/Window/GliaViewController.swift
VladKupriienko: The thought of moving it into an extension hasn't crossed my mind, to be honest, but why not.
As for the layout, it's using constraints.

VladKupriienko: Named the extension function `insertChild`, because just `add` seemed not verbose enough to me.

dukhovnyi (approved):
Looks good 👍

PR #60, opened 2021-06-11
[MUIC-339] Fix green border staying after screen share stopped in audio/video call

VladKupriienko:
**Situation**: the green border is controlled by an instance of the `VisitorScreenSharingState` class. When it is `nil`, a green border that indicates that the screen is being recorded is removed.

**Previously**, such an instance was just a field of `EngagementViewModel`, which both `ChatViewModel` and `CallViewModel` inherit. When an audio/video call was launched, the app would push both `ChatViewController` and `CallViewController` into the view hierarchy, each of them holding a reference to their `ViewModel`, which in return held a reference to `VisitorScreenSharingState`. This created a situation where to fully "stop" screen sharing one would have to access both of these ViewModels and set the value to `nil`.

**Instead**, I've introduced a shared handler that is created in `RootCoordinator` and injected from there.

_my favourite ✨ :_ Also includes code style refactor for affected files.


GliaWidgets/Coordinator/Call/CallCoordinator.swift
```diff
@@ -63,7 +73,6 @@ class CallCoordinator: SubFlowCoordinator, FlowCoordinator {
                 self?.delegate?(.minimize)
             }
         }
-        return CallViewController(viewModel: viewModel,
-                                  viewFactory: viewFactory)
+        return CallViewController(viewModel: viewModel, viewFactory: viewFactory)
```
dukhovnyi, line 76: 39-42 lines have another alignment. Let's use the same formatting in both places.

GliaWidgets/Lib/Screensharing/ScreenshareHandler.swift
```diff
@@ -0,0 +1,19 @@
+import SalemoveSDK
+
+class ScreenshareHandler {
+    private var visitorState: VisitorScreenSharingState?
+
+    func updateState(to state: VisitorScreenSharingState?) {
+        visitorState = state
```
dukhovnyi, line 7: do we need to stop sharing before assign a new state?

kuldar-daniel (approved):

GliaWidgets.xcodeproj/project.pbxproj
```diff
@@ -181,6 +181,7 @@
 		1AFB1E7425F8B00B00CA460D /* ChatTextContentView.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1AFB1E7325F8B00B00CA460D /* ChatTextContentView.swift */; };
 		1AFB1E7825F8B26800CA460D /* ChatTextContentStyle.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1AFB1E7725F8B26800CA460D /* ChatTextContentStyle.swift */; };
 		534EEAAA3FCDDD78AAAEB33E /* Pods_TestingApp.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = 9C45ADF8988F719DA1AC8444 /* Pods_TestingApp.framework */; };
+		C42463742673ABE10082C135 /* ScreenshareHandler.swift in Sources */ = {isa = PBXBuildFile; fileRef = C42463732673ABE10082C135 /* ScreenshareHandler.swift */; };
```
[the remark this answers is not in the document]
VladKupriienko: Yes, agreed. This is a mistake that often plagues me - writing screen sharing as one word 😅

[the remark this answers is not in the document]
VladKupriienko: In SDK there is a class named `ScreenshareOfferBlock`. Maybe I was actually trying to follow that naming at the time.
There is also a `ScreenSharingVisitorState` though :D

GliaWidgets/Coordinator/Call/CallCoordinator.swift
VladKupriienko: Lines 39-42 is how it should be, yes. I just thought that this line wasn't long enough to insert linebreaks. Corrected

GliaWidgets/Lib/Screensharing/ScreenshareHandler.swift
VladKupriienko: The SDK doesn't actually fire an event of screen sharing being ended (I would've assumed it would trigger an `onVisitorScreenSharingStateChange` with the status of `.notSharing`, but that's not the case). This is only used to signal the screen sharing session actually starting to show the "End screen share" button.

So yeah, I don't think it's needed.

VladKupriienko:
This feature has already passed the QA, so I'm changing the target branch to `dev`, skipping the `qa/MUIC-339`.
Such is life when dealing with timeline-critical features 😶

PR #67, opened 2021-06-15
[MUIC-352] Fix "Stop screen share" button not disappearing in chat view

VladKupriienko:
Previously the "Stop screen share" button would stay in the chat header if screen sharing was stopped from the upgraded call view.

I've introduced a new observable variable to `ScreenShareHandler` that reflects the changes in screen sharing status. Views subscribe to that observable to modify header buttons as needed.

I've also modified the code for setting item in the header as it sometimes caused some leftover views.

Additionally, moved the screen-share-related code to `EngagementViewModel` and `EngagementViewController`, because previously this code was basically duplicated in Call and Chat VM & VC, which are both subclasses of Engagement VM & VC correspondingly.

### DIFF
--- a/GliaWidgets/ViewModel/EngagementViewModel.swift
+++ b/GliaWidgets/ViewModel/EngagementViewModel.swift
@@ -56,8 +56,8 @@ class EngagementViewModel {
         self.alertConfiguration = alertConfiguration
         self.screenShareHandler = screenShareHandler
         interactor.addObserver(self, handler: interactorEvent)
-        screenShareHandler.status.addObserver(self) { status, _ in
-            self.onScreenSharingStatusChange(status)
+        screenShareHandler.status.addObserver(self) { [weak self] status, _ in
+            self?.onScreenSharingStatusChange(status)
         }
     }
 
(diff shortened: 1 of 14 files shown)


kuldar-daniel (approved):

GliaWidgets/ViewModel/EngagementViewModel.swift
```diff
@@ -54,10 +56,14 @@ class EngagementViewModel {
         self.alertConfiguration = alertConfiguration
         self.screenShareHandler = screenShareHandler
         interactor.addObserver(self, handler: interactorEvent)
+        screenShareHandler.status.addObserver(self) { status, _ in
+            self.onScreenSharingStatusChange(status)
```
dukhovnyi, line 60: Let's use weak reference to self to avoid memory leak.

dukhovnyi (changes requested):

dukhovnyi:
Looks good, just one small fix is needed.

GliaWidgets/ViewModel/EngagementViewModel.swift
VladKupriienko: Ah, the same mistake again. Thanks

VladKupriienko: Fixed this in other places too

VladKupriienko:
Changed target to `dev` as the QA has already greenlit this fix.

dukhovnyi (approved):

PR #82, opened 2021-07-06
[MUIC-383] Fix unresponsive screen after engagement ends

VladKupriienko:
Implemented a suggestion from @kuldar-daniel, cheers for that!

He suggested that computing the topmost view controller might be unreliable when dismissing the Glia view, which proved to be the case in this issue.

Implementing Kuldar's suggestion fixed that problem and potentially made the Widgets more stable towards any UI changes in the Integrator's app.


kuldar-daniel (approved):

GliaWidgets/Coordinator/RootCoordinator.swift
```diff
@@ -38,10 +38,7 @@ class RootCoordinator: SubFlowCoordinator, FlowCoordinator {
 
     private let navigationController = NavigationController()
     private let navigationPresenter: NavigationPresenter
-    private let presentingWindow = UIApplication.shared.windows.filter { $0.isKeyWindow }.first
-    private var presentingViewController: UIViewController? {
-        return presentingWindow?.topMostViewController()
-    }
+    private let gliaPresenter = GliaPresenter()
```
dukhovnyi, line 41: imho, the property name could be `presenter` since it has the type `GliaPresenter`.

GliaWidgets/Lib/Presenter/GliaPresenter.swift
```diff
@@ -0,0 +1,34 @@
+import UIKit
```
dukhovnyi, line 1: Let's add a new line between import and class definition.

GliaWidgets/Lib/Presenter/GliaPresenter.swift
```diff
@@ -0,0 +1,34 @@
...
+internal final class GliaPresenter {
```
dukhovnyi, line 2: `internal` keywords here and below is unnecessary since it is internal by default. Or do I missed something?

GliaWidgets/Lib/Presenter/GliaPresenter.swift
```diff
@@ -0,0 +1,34 @@
...
+    private var window: UIWindow? {
+        if #available(iOS 13, *) {
+            return UIApplication.shared.windows.first { $0.isKeyWindow }
+        } else {
+            return UIApplication.shared.keyWindow
+        }
+    }
+
+    internal var root: UIViewController {
```
dukhovnyi, line 11: Let's improve the naming for this property. I think the `root` is not the best name for it because of
- the class name is `***Presenter`, so, at least for me, this property name looks like it should be the `root` (or parent) presenter or smth like that. If you would like, let's name it `rootViewController` at least.
- it has `root` name but returns the opposite view controller - the most presented view controller. So, perhaps we could put the previous name, smth like `topMostViewController`

What do you think, does my concern make sense?

GliaWidgets/Lib/Presenter/GliaPresenter.swift
```diff
@@ -0,0 +1,34 @@
...
+
+        return presenter
+    }
+
+    internal func present(_ viewController: UIViewController, animated: Bool, completion: (() -> Void)? = nil) {
+        root.present(viewController, animated: animated, completion: completion)
+    }
+
+    internal func dismiss(_ viewController: UIViewController, animated: Bool, completion: (() -> Void)?) {
+        viewController.presentingViewController?.dismiss(animated: animated, completion: completion)
```
dukhovnyi, line 32: Do you expect of changing logic here?
Looks like before this PR the dismiss method called for `topMostViewController` ([here](https://github.com/salemove/ios-sdk-widgets/pull/82/files#diff-c55b54fda28a347f2fdd0c742537e17ab4335a1e4f9f4a0ff227c5441e11bef3L261))

GliaWidgets/Lib/Presenter/GliaPresenter.swift
kuldar-daniel: I like to think of it as root for presenter, so root that the presenter will present on, basically. But it's a good argument

GliaWidgets/Lib/Presenter/GliaPresenter.swift
kuldar-daniel: Yes, this is the actual fix. Previously, topMostViewController was a computed property, so by the time someone calls dismisis on it, the current topMostViewController could be something totally different. So the right approach is to get the vc, that presented Glia, and call dismiss on that.

GliaWidgets/Lib/Presenter/GliaPresenter.swift
kuldar-daniel: I think we can make it private also

GliaWidgets/Lib/Presenter/GliaPresenter.swift
VladKupriienko: As I understood judging by the 👍 , you agree with the explanation provided. Resolving :)

GliaWidgets/Coordinator/RootCoordinator.swift
VladKupriienko: Seemed more natural just below the `navigationPresenter`. Tbh I'd leave it as is if that doesn't bother you too much 😅

GliaWidgets/Lib/Presenter/GliaPresenter.swift
VladKupriienko: Whoops. Fixed

GliaWidgets/Lib/Presenter/GliaPresenter.swift
VladKupriienko: Yes, you are correct. Removed

dukhovnyi: Please delete `internal` keyword for the class definition also.

VladKupriienko: Sorry, a bit unfocused today. Removed there as well

VladKupriienko:
I have changed the logic a bit. We provide a possibility for the Integrator to provide their own SceneProvider, which is a protocol with the sole purpose of getting a `UIWindowScene`. I've used it in the GliaPresenter too. AFAIK, there is no reliable way of getting the currently active and focused scene from a model class, so SceneProvider might come in handy for the rare cases of the Integrator app having multiple scenes. Otherwise, the current approach should work just fine.

dukhovnyi (approved):

PR #86, opened 2021-07-09
[MUIC-376] Fix sent message canceling active uploads

VladKupriienko:
Previously sending a message would remove all uploads regardless of their status.

Now it only removes (and attaches) the files that have been successfully uploaded. All the other files stay in the file upload area and get attached to some new message as soon as they are uploaded.

### DIFF
--- a/GliaWidgets/Lib/Upload/FileUploader.swift
+++ b/GliaWidgets/Lib/Upload/FileUploader.swift
@@ -95,8 +95,15 @@ class FileUploader {
         updateLimitReached()
     }
 
-    func removeAllUploads() {
-        uploads.removeAll()
+    func removeSucceededUploads() {
+        uploads.removeAll(where: { upload in
+            switch upload.state.value {
+            case .uploaded:
+                return true
+            default:
+                return false
+            }
+        })
         updateState()
         updateLimitReached()
     }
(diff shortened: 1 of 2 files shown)


GliaWidgets/Lib/Upload/FileUploader.swift
```diff
@@ -95,8 +95,15 @@ class FileUploader {
...
-        uploads.removeAll()
+    func removeSucceededUploads() {
+        uploads.removeAll(where: { upload in
+            switch upload.state.value {
+            case .uploaded:
+                return true
+            default:
+                return false
+            }
+        })
```
dukhovnyi, line 106: Probably we could simplify this closure to 
```suggestion
        uploads.removeAll(where: { $0..state.value == .uploaded })
```

dukhovnyi (approved):
LGTM, just a small suggestion.

kuldar-daniel (approved):

GliaWidgets/Lib/Upload/FileUploader.swift
VladKupriienko: Actually no, this wouldn't compile. 
I remember trying to make it a one-liner but I was unable to do so. Tried so again after coming back from vacation but still nothing. I don't think it's worth my time anymore or is critical in any regard, so I'll leave it as is.

kuldar-daniel: It can't be done because FileUpload.State has parameters for enum cases.

VladKupriienko: Yes, exactly

PR #88, opened 2021-07-19
Add support for Cocoapods

gersonnoboa:
The same XCFramework released as 0.4.8 has been added to the root of the project for Cocoapods support, as well as the necessary podspec. The necessary podspec for SalemoveSDK has also been added to [glia-ios-podpsecs](https://github.com/salemove/glia-ios-podspecs/blob/master/SalemoveSDK/0.29.6/SalemoveSDK.podspec), so that Cocoapods knows where to find all necessary dependencies.

MOB-811


dukhovnyi (commented):
Since we are adding support of Cocoapods, I'd like to propose changing the Xcode project version and distributing it with the correct version (1.0 as of now). 
@gersonnoboa  @kuldar-daniel  What do you think guys, does it make sense?

dukhovnyi (approved):

PR #101, opened 2021-09-01
[MUIC-359] Fix views hiding in landscape call view

VladKupriienko:
There was an issue where the header and buttons bar would break due to the CallView in the background not receiving orientation change events.

Thus I introduced a fix that adjusts the constraints on `viewWillAppear` call.

### DIFF
--- a/GliaWidgets/View/Call/CallView.swift
+++ b/GliaWidgets/View/Call/CallView.swift
@@ -32,7 +32,7 @@ class CallView: EngagementView {
     private let kRemoteVideoViewPortraitHeightMultiplier: CGFloat = 0.3
     private let kRemoteVideoViewLandscapeHeightMultiplier: CGFloat = 1.0
     private let kBarsHideDelay: TimeInterval = 3.2
-    private let barsAreHidden = ObservableValue<Bool>(with: false)
+    private var barsAreHidden: Bool = false
 
     init(with style: CallStyle) {
         self.style = style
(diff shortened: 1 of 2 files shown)


kuldar-daniel (approved):

GliaWidgets/View/Call/CallView.swift
```diff
@@ -32,6 +32,7 @@ class CallView: EngagementView {
     private let kRemoteVideoViewPortraitHeightMultiplier: CGFloat = 0.3
     private let kRemoteVideoViewLandscapeHeightMultiplier: CGFloat = 1.0
     private let kBarsHideDelay: TimeInterval = 3.2
+    private let barsAreHidden = ObservableValue<Bool>(with: false)
```
kuldar-daniel, line 35: should we dispose it on deinit to prevent leak?

kuldar-daniel: Wait, why do we need ObservableValue in the first place

VladKupriienko: hm, to be on the safe side - probably yes. will fix shortly
Unfortunately, there are _a lot_ other `ObservableValue`s in the project that are not being disposed.

VladKupriienko: > Wait, why do we need ObservableValue in the first place

now that you are telling this, I understand that this could've been just a Bool. Whoops

GliaWidgets/ViewController/Call/CallViewController.swift
```diff
@@ -27,6 +27,11 @@ class CallViewController: EngagementViewController, MediaUpgradePresenter {
         view.willRotate(to: orientation, duration: duration)
     }
 
+    override func viewWillAppear(_ animated: Bool) {
+        guard let view = view as? CallView else { return }
```
dukhovnyi, line 31: Is it expected to don't call `super` here?

VladKupriienko: There was no intention of fully replacing `viewWillAppear`. Fixed, thanks for noticing.

dukhovnyi (approved):

PR #103, opened 2021-09-05
Remove using of fatalError to avoid crash when font can't be registered

dukhovnyi:


GliaWidgets/Resources/FontProvider.swift
```diff
@@ -16,30 +16,35 @@ internal class FontProvider {
...
+        return UIFont(name: named, size: size) ?? .systemFont(ofSize: size)
     }
 
     private func loadFonts() {
-        fonts.forEach { loadFont(named: $0) }
+        _ = fonts.first(where: { !loadFont(named: $0) })
     }
 
-    private func loadFont(named name: String) {
+    private func loadFont(named name: String) -> Bool {
```
gersonnoboa, line 26: if this method returns a boolean then the name is not descriptive about that.

gersonnoboa (commented):
What is it that we are trying to fix? Maybe a little bit of context in the description would have been good because I am out of the loop, not sure if the other reviewers are also aware of what the situation is.

If I understand correctly, now if the font is not loaded, it uses the system font instead? Is this something that is desired or someone has asked for it? I think it has to do with integration with React Native, right? If that is the case, then it should be documented somewhere. If you are trying to load a font but it doesn't work, maybe it's not the worst idea to crash it because it is a development issue.

dukhovnyi:
This is code for loading custom font. The reason for unsuccessful font loading could be pretty different: the font is not found in bundle (should be found on development step), font manager is broken (OS issue, a reboot could help with this situation), font is not allowed to use by some OS restrictions or any reason that Apple think is applicable for this case.

`fatalError` throws runtime error and should be applied for the critical situations when the application can't or shouldn't work. In this situation, I believe it better to have a system font instead of a custom Roboto font, but not crash the application.

@gersonnoboa Does it make sense?

gersonnoboa:
@dukhovnyi ahhh okay, makes sense. Thanks for the context.

gersonnoboa (approved):

PR #106, opened 2021-09-08
MUIC-455 [iOS] Audio/Video - Upgrading the audio to video call in Chat view it does not upgrade to Video call

kuldar-daniel:
This solves an issue where if you start with audio call, then open chat, and operator tries to upgrade to video, the media upgrade dialog get's handled incorrectly, since the engagement is actually call, which was not handled. So nothing would happen. Properly upgrade to correct offer.

The solution is not ideal, but for better solution, we should refactor RootCoordinator quite a bit.

### DIFF
--- a/GliaWidgets/Coordinator/RootCoordinator.swift
+++ b/GliaWidgets/Coordinator/RootCoordinator.swift
@@ -13,7 +13,7 @@ class RootCoordinator: SubFlowCoordinator, FlowCoordinator {
     private enum Engagement {
         case none
         case chat(ChatViewController)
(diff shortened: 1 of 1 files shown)


VladKupriienko:
I'm a bit confused about the target/desitnation branches. 
We're usually merging from bugfix[feature]/MUIC-XXX into qa/MUIC-XXX for Eve to test, and only then into dev when the QA passes.

kuldar-daniel:
Sorry, messed up when I created branch, see https://github.com/salemove/ios-sdk-widgets/pull/107

VladKupriienko:
No biggie :)
Checking the new one

PR #114, opened 2021-09-20
[MUIC-285] Fix choice card border for iOS 12-13

KarimovaKarina:


GliaWidgets/View/Chat/Message/ChoiceCardView.swift
```diff
@@ -1,70 +1,61 @@
...
         default:
             break
         }
     }
-
-    private func contentViews(for choiceCard: ChoiceCard) -> [UIView] {
-        var views = [UIView]()
-
+    private func contentView(for choiceCard: ChoiceCard) -> UIView {
+        let view = UIView()
```
VladKupriienko, line 25: Probably a good idea to give it a more descriptive name - `containerView` for example.

GliaWidgets/View/Chat/Message/ChoiceCardView.swift
```diff
@@ -1,70 +1,61 @@
...
+
+        view.layer.cornerRadius = 8
+        view.layer.borderWidth = 1
+        view.layer.borderColor = viewStyle.frameColor.cgColor
+        let stackView = UIStackView()
```
VladKupriienko, line 30: I'd add a newline between these two lines for readability: `view`-related stuff is finished for now, moving onto `stackView`.

GliaWidgets/View/Chat/Message/ChoiceCardView.swift
```diff
@@ -1,70 +1,61 @@
...
+        stackView.axis = .vertical
+        stackView.spacing = 12
+        stackView.isLayoutMarginsRelativeArrangement = true
+        stackView.layoutMargins = kLayoutMargins
+
+        view.addSubview(stackView)
+        stackView.autoPinEdgesToSuperviewEdges()
         if let imageUrl = choiceCard.imageUrl {
```
VladKupriienko, line 38: Same as above - a newline would improve readability.

GliaWidgets/View/Chat/Message/ChoiceCardView.swift
```diff
@@ -1,70 +1,61 @@
...
             contentAlignment: .left,
             insets: .zero
         )
         textView.text = choiceCard.text
-        views.append(textView)
-
-        guard let options = choiceCard.options else { return views }
-
+        stackView.addArrangedSubview(textView)
+        guard let options = choiceCard.options else {
```
VladKupriienko, line 54: Same here about the newline. Also I think we should keep the `guard ... else { return view }` as a single line as it is quite short and makes the code look a bit cleaner.

VladKupriienko (changes requested):
I don't know if it was accidental, but basically every empty newline was removed from this file for some reason. Please bring them back, having code in a single chunk really hurts the readability of said file :)

VladKupriienko (approved):

PR #115, opened 2021-09-20
[MUIC-390] Fix blinking header button

KarimovaKarina:
Additionally fixed end engagement button appearing after closing engagement

### DIFF
--- a/GliaWidgets/Component/Header/Header.swift
+++ b/GliaWidgets/Component/Header/Header.swift
@@ -24,7 +24,8 @@ class Header: UIView {
     var backButton: HeaderButton
     var closeButton: HeaderButton
     var endButton: ActionButton
-    var endScreenSharingButton: HeaderButton
+    var endScreenShareButton: HeaderButton
+
     private let style: HeaderStyle
     private let leftItemContainer = UIView()
     private let rightItemContainer = UIStackView()
@@ -41,7 +42,7 @@ class Header: UIView {
         self.backButton = HeaderButton(with: style.backButton)
         self.closeButton = HeaderButton(with: style.closeButton)
         self.endButton = ActionButton(with: style.endButton)
-        self.endScreenSharingButton = HeaderButton(with: style.endScreenShareButton)
+        self.endScreenShareButton = HeaderButton(with: style.endScreenShareButton)
         super.init(frame: .zero)
         setup()
         layout()
@@ -57,47 +58,26 @@ class Header: UIView {
         updateHeight()
     }
 
-    public func setLeftItem(animated: Bool) {
-        setItem(backButton, to: leftItemContainer, animated: animated)
+    func showBackButton() {
+        backButton.isHidden = false
     }
 
-    public func showCloseButton() {
+    func showCloseButton() {
         endButton.isHidden = true
-        endScreenSharingButton.isHidden = true
+        endScreenShareButton.isHidden = true
         closeButton.isHidden = false
-    }
-
-    func setEndButtons() {
-        endButton.isHidden = true
-        endScreenSharingButton.isHidden = true
     }
 
     func showEndButton() {
         endButton.isHidden = false
         closeButton.isHidden = true
-        endScreenSharingButton.isHidden = true
+        endScreenShareButton.isHidden = true
     }
 
     func showEndScreenSharingButton() {
         endButton.isHidden = false
-        endScreenSharingButton.isHidden = false
+        endScreenShareButton.isHidden = false
         closeButton.isHidden = true
-    }
-
-    private func setItem(_ item: UIView?, to container: UIView, animated: Bool) {
-        container.subviews.forEach { $0.removeFromSuperview() }
-
-        guard let item = item else { return }
-        item.alpha = 0.0
-        container.addSubview(item)
-        item.autoPinEdge(toSuperviewEdge: .left, withInset: 0, relation: .greaterThanOrEqual)
-        item.autoPinEdge(toSuperviewEdge: .top, withInset: 0, relation: .greaterThanOrEqual)
-        item.autoPinEdge(toSuperviewEdge: .right, withInset: 0, relation: .greaterThanOrEqual)
-        item.autoPinEdge(toSuperviewEdge: .bottom, withInset: 0, relation: .greaterThanOrEqual)
-        item.autoCenterInSuperview()
-        UIView.animate(withDuration: animated ? 0.2 : 0.0) {
-            item.alpha = 1.0
-        }
     }
 
     private func setup() {
@@ -128,11 +108,11 @@ class Header: UIView {
         titleLabel.autoPinEdge(toSuperviewEdge: .right)
         titleLabel.autoAlignAxis(toSuperviewAxis: .horizontal)
 
-        contentView.addSubview(leftItemContainer)
-        leftItemContainer.autoPinEdge(toSuperviewEdge: .left)
-        leftItemContainer.autoAlignAxis(toSuperviewAxis: .horizontal)
+        contentView.addSubview(backButton)
+        backButton.autoPinEdge(toSuperviewEdge: .left)
+        backButton.autoAlignAxis(toSuperviewAxis: .horizontal)
 
-        rightItemContainer.addArrangedSubviews([endScreenSharingButton, endButton, closeButton])
+        rightItemContainer.addArrangedSubviews([endScreenShareButton, endButton, closeButton])
         contentView.addSubview(rightItemContainer)
         rightItemContainer.autoPinEdge(toSuperviewEdge: .right)
         rightItemContainer.autoAlignAxis(toSuperviewAxis: .horizontal)
--- a/GliaWidgets/Component/Header/Header.swift
+++ b/GliaWidgets/Component/Header/Header.swift
@@ -21,10 +21,13 @@ class Header: UIView {
             }
         }
     }
-
+    var backButton: HeaderButton
+    var closeButton: HeaderButton
+    var endButton: ActionButton
+    var endScreenSharingButton: HeaderButton
     private let style: HeaderStyle
     private let leftItemContainer = UIView()
-    private let rightItemContainer = UIView()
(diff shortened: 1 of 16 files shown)


GliaWidgets/View/Chat/Entry/ChatMessageEntryView.swift
```diff
@@ -21,11 +21,11 @@ class ChatMessageEntryView: UIView {
             if isChoiceCardModeEnabled {
                 textView.resignFirstResponder()
             }
-            
+
```
VladKupriienko, line 24: Just to explain why it is good to avoid trailing whitespace - as you can see, this file has no actual changes apart from these whitespaces, but it still shows up as 24 changed lines in the pull request, making it unnecessarily big - "false changes", basically. You can read more on that in this [SO post.](https://stackoverflow.com/questions/300489/why-is-it-bad-to-commit-lines-with-trailing-whitespace-into-source-control)

GliaWidgets/Component/Header/HeaderStyle.swift
```diff
@@ -10,6 +10,10 @@ public struct HeaderStyle {
 
     /// Background color of the view.
     public var backgroundColor: UIColor
+    public var backButton: HeaderButtonStyle
+    public var closeButton: HeaderButtonStyle
+    public var endButton: ActionButtonStyle
+    public var endScreenShareButton: HeaderButtonStyle
 
     ///
     /// - Parameters:
```
VladKupriienko, line 19: ...and don't forget to add them there as well :)

GliaWidgets/Component/Header/Header.swift
```diff
@@ -21,10 +21,13 @@ class Header: UIView {
             }
         }
     }
-
+    var backButton: HeaderButton
+    var closeButton: HeaderButton
+    var endButton: ActionButton
+    var endScreenSharingButton: HeaderButton
     private let style: HeaderStyle
```
VladKupriienko, line 28: A newline here probably wouldn't hurt :)

GliaWidgets/Component/Header/Header.swift
```diff
@@ -50,23 +57,31 @@ class Header: UIView {
         updateHeight()
     }
 
-    func setLeftItem(_ item: UIView?, animated: Bool) {
-        setItem(item, to: leftItemContainer, animated: animated)
+    public func setLeftItem(animated: Bool) {
+        setItem(backButton, to: leftItemContainer, animated: animated)
     }
```
VladKupriienko, line 62: 1. No need to make it public. `Header` itself is an `internal` class, so everything in it can only be `internal` or stricter (`fileprivate` and `private`). Even though it compiles without warnings, it does not make sense to declare methods or fields as `public`.
2. If this method always sets only the back button, the name of the function should reflect that - it's not `setLeftItem` anymore, rather `addBackButton`.

GliaWidgets/Component/Header/Header.swift
```diff
@@ -50,23 +57,31 @@ class Header: UIView {
...
 
-    func setRightItem(_ item: UIView?, animated: Bool) {
-        setItem(item, to: rightItemContainer, animated: animated)
+    public func showCloseButton() {
```
VladKupriienko, line 64: Same here about the `public` modifier.

GliaWidgets/Component/Header/Header.swift
VladKupriienko: Actually, looking through the other bits of code, I'm not even sure this function should behave as it does. 

If the left container only ever contains a back button, why not add it straight away, just as you do with the `endButton`, `endScreenShareButton` and `closeButton` in `layout()`? And then create a function `showBackButton` which only affects the `isHidden` attribute and call it?

GliaWidgets/Component/Header/Header.swift
```diff
@@ -50,23 +57,31 @@ class Header: UIView {
...
-    func setRightItems(_ items: [UIView]?, animated: Bool) {
-        var stack: UIStackView?
-        if let items = items {
-            stack = UIStackView(arrangedSubviews: items)
-            stack?.axis = .horizontal
-            stack?.spacing = 8
-            stack?.alignment = .center
-        }
-        setRightItem(stack, animated: animated)
+    func setEndButtons() {
```
VladKupriienko, line 70: All your other functions that operate on the `isHidden` attribute are called `show***Button`, why is this one different then? :)
Obviously, something like `showEndEngagementAndEndScreenShareButtons` is way too long, but we can make it `showBothEndButtons` or something like this. I honestly don't like the "both" variant very much, but I wasn't able to produce anything more better-looking 😄

GliaWidgets/ViewController/EngagementViewController.swift
```diff
@@ -56,15 +56,9 @@ class EngagementViewController: ViewController, AlertPresenter {
...
-                endEngagementButton.tap = { viewModel.event(.closeTapped) }
-                view.header.setRightItem(endEngagementButton, animated: true)
+                view.header.showEndButton()
             case .showEndScreenShareButton:
-                let endEngagementButton = ActionButton(with: self.viewFactory.theme.chat.endButton)
-                endEngagementButton.tap = { viewModel.event(.closeTapped) }
-                let endScreenShareButton = HeaderButton(with: self.viewFactory.theme.chat.endScreenShareButton)
-                endScreenShareButton.tap = { viewModel.event(.endScreenSharingTapped) }
-                view.header.setRightItems([endScreenShareButton, endEngagementButton], animated: true)
+                view.header.showEndScreenSharingButton()
```
VladKupriienko, line 61: It's very good that we removed the creation of buttons from the `bind(...)` method! It was really sticking out :)

GliaWidgets/Component/Header/Header.swift
```diff
@@ -21,10 +21,13 @@ class Header: UIView {
             }
         }
     }
-
+    var backButton: HeaderButton
+    var closeButton: HeaderButton
+    var endButton: ActionButton
+    var endScreenSharingButton: HeaderButton
```
VladKupriienko, line 27: I'm pretty sure it was called `endScreenShareButton` previously. Not a big deal, just a continuity/consistency issue.

However, a quick search of "screensharing" reveals a bunch of other places in the code where such a naming scheme is used. Even in the SalemoveSDK events 😄  But let's at least call the button the same way as it was called previously.

GliaWidgets/ViewController/EngagementViewController.swift
```diff
@@ -32,19 +32,19 @@ class EngagementViewController: ViewController, AlertPresenter {
...
     }
 
-    func showCloseButton(with style: HeaderButtonStyle, in header: Header) {
-        let back = HeaderButton(with: style, tap: { self.viewModel.event(.closeTapped) })
-        back.contentHorizontalAlignment = .right
-        header.setRightItem(back, animated: false)
+    func showCloseButton(in header: Header) {
+        header.closeButton.tap = { self.viewModel.event(.closeTapped) }
+        header.showCloseButton()
     }
```
VladKupriienko, line 43: I think we can remove both of these methods entirely. Just move the `header.***button.tap` line to `bind()`, just as you did with the others, and call the `header.showCloseButton()` and `header.showBackButton()` (or however you will rename it) in `bind(...)` methods of `CallViewController` and `ChatViewController`.

GliaWidgets/ViewModel/EngagementViewModel.swift
```diff
@@ -1,6 +1,12 @@
 import SalemoveSDK
 
 class EngagementViewModel {
+    enum RightButtons {
+        case closeButton
+        case endButton
+        case endSharingButton
+    }
+
```
VladKupriienko, line 9: Did I miss it or this enum was left unused?
If so, then please remove it :)

VladKupriienko (changes requested):
Very good in general, I saw a lot of welcome changes here. Please go through my comments to make the final touches and we're good to go :)

GliaWidgets/Component/Header/Header.swift
KarimovaKarina: I found out that I don't use this function at all😄

GliaWidgets/Component/Header/HeaderStyle.swift
```diff
@@ -11,19 +11,43 @@ public struct HeaderStyle {
     /// Background color of the view.
     public var backgroundColor: UIColor
 
+    /// Back button style in the header.
```
VladKupriienko, line 14: I should've read these more carefully before suggesting taking them directly from other styles 😅 
"in the header" is of course redundant here - it's `HeaderStyle` after all.
Let's make these consistent with `endButton` and `endScreenShareButton`:
`Style of the back button.`

GliaWidgets/Component/Header/HeaderStyle.swift
```diff
@@ -11,19 +11,43 @@ public struct HeaderStyle {
...
+    public var backButton: HeaderButtonStyle
+
+    /// Close button style in the header.
```
VladKupriienko, line 17: `Style of the queue closing button.`

GliaWidgets/Component/Header/HeaderStyle.swift
```diff
@@ -11,19 +11,43 @@ public struct HeaderStyle {
...
+    /// Style of the screen sharing ending button.
+    public var endScreenShareButton: HeaderButtonStyle
+
     ///
     /// - Parameters:
     ///   - titleFont: Font of the title text.
     ///   - titleColor: Color of the title text.
     ///   - backgroundColor: Background color of the view.
+    ///   - backButton: Back button style in the header.
+    ///   - closeButton: Close button style in the header.
```
VladKupriienko, line 32: same as above

VladKupriienko (approved):
All of the issues are fixed, very nice :)
The last bit would be to improve the descriptions a bit (please see the new comments), and after that, we can finally merge 👍

VladKupriienko (approved):

PR #116, opened 2021-09-20
[MUIC-358] Bring back custom transition animations for Glia view

KarimovaKarina:


GliaWidgets/View/Transition.swift
```diff
@@ -0,0 +1,95 @@
...
+extension Transition: UIViewControllerAnimatedTransitioning {
+    func transitionDuration(using transitionContext: UIViewControllerContextTransitioning?) -> TimeInterval {
+        return duration
+    }
+    
+    func animateTransition(using transitionContext: UIViewControllerContextTransitioning) {
+        let containerView = transitionContext.containerView
+        
+        if transitionMode == .present {
+            if let presentedView = transitionContext.view(forKey: UITransitionContextViewKey.to) {
```
VladKupriienko, line 30: So currently this function is responsible for both presenting and dismissing the view, which makes it kinda bulky. I suggest splitting it in two, basically creating `presentView(...)` and `dismissView(...)` private methods and making the only responsibility of `animateTransition(...)` to choose which one to call depending on `transitionMode`. You can pass `transitionContext` parameter along to them, of course.

I feel like that would go in line with single-responsibility principle of SOLID.

As a bonus, you would get rid of huge chunks of code being indented to the right from being thrown into an if/else statement :)

GliaWidgets/ViewController/GliaViewController.swift
```diff
@@ -48,7 +50,7 @@ class GliaViewController: UIViewController {
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
-
```
VladKupriienko, line 51: Hey, found another trailing whitespace 😄 

I think it would be very beneficial if you tried to build _all_ of the branches we created on Monday and look through the SwiftLint warnings, now that you have it installed. You can then run `swiftlint --fix` if needed.

VladKupriienko (changes requested):
Very good that you got it working!
A few tweaks (mostly cosmetic) and we can put it to QA :)

GliaWidgets/View/Transition.swift
```diff
@@ -0,0 +1,95 @@
+import UIKit
+
+class Transition: NSObject {
```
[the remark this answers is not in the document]
VladKupriienko: Also, let's make it `final`. I don't expect that we would have child classes inheriting from this one.

VladKupriienko (approved):

PR #118, opened 2021-09-24
[MUIC-287] Add borders to choice cards options

KarimovaKarina:
Borders for options appear only when choice card is disabled.

### DIFF
--- a/GliaWidgets/Theme/Theme+Chat.swift
+++ b/GliaWidgets/Theme/Theme+Chat.swift
@@ -93,15 +93,28 @@ extension Theme {
         let choiceCardImageFile = ChatImageFileContentStyle(
             backgroundColor: color.baseLight
         )
-        let choiceCardOption = ChoiceCardOptionStyle(
+        let choiceCardOptionNormalState = ChoiceCardOptionStateStyle(
             textFont: font.bodyText,
-            normalTextColor: color.baseDark,
-            normalBackgroundColor: Color.lightGrey,
-            selectedTextColor: color.baseLight,
-            selectedBackgroundColor: color.primary,
-            disabledTextColor: Color.grey,
-            disabledBackgroundColor: Color.lightGrey,
-            disabledBorderColor: Color.baseShade
+            textColor: color.baseDark,
+            backgroundColor: Color.lightGrey,
+            borderColor: nil
+        )
+        let choiceCardOptionSelectedState = ChoiceCardOptionStateStyle(
+            textFont: font.bodyText,
+            textColor: color.baseLight,
+            backgroundColor: color.primary,
+            borderColor: nil
+        )
+        let choiceCardOptionDisabledState = ChoiceCardOptionStateStyle(
+            textFont: font.bodyText,
+            textColor: Color.grey,
+            backgroundColor: Color.lightGrey,
+            borderColor: Color.baseShade
+        )
+        let choiceCardOption = ChoiceCardOptionStyle(
+            normal: choiceCardOptionNormalState,
+            selected: choiceCardOptionSelectedState,
+            disabled: choiceCardOptionDisabledState
         )
         let choiceCard = ChoiceCardStyle(
             mainText: choiceCardText,
--- a/GliaWidgets/Theme/Theme+Chat.swift
+++ b/GliaWidgets/Theme/Theme+Chat.swift
@@ -97,11 +97,11 @@ extension Theme {
             textFont: font.bodyText,
             normalTextColor: color.baseDark,
             normalBackgroundColor: Color.lightGrey,
-            highlightedTextColor: color.baseLight,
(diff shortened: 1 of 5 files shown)


GliaWidgets/Theme/Theme+Chat.swift
```diff
@@ -99,6 +99,7 @@ extension Theme {
             normalBackgroundColor: Color.lightGrey,
             highlightedTextColor: color.baseLight,
             highlightedBackgroundColor: color.primary,
+            disabledTabsBorderColor: Color.baseShade,
```
VladKupriienko, line 102: "Tabs" is not used anywhere else in the code. They are instead called "options" :)
Also, this is an initializer for the choice card options, thus it is obvious that every parameter is related to it, so even "options" in the name is redundant.

Therefore I suggest renaming it to `disabledBorderColor`. It falls in line nicely with other naming schemes - state (normal/highlighted/disabled, what element it is related to (text/border/background) and what property (color).

**Also**, I suggest moving it to the end of the initializer to have a set order of text - background - border (more in-depth reasons in next comments).

GliaWidgets/View/Chat/Message/Content/Choice/ChoiceCardOptionStyle.swift
```diff
@@ -10,6 +10,8 @@ public final class ChoiceCardOptionStyle {
 
     /// Style of an option in a disabled state - choice card has already been answered by the visitor or the choice card became inactive (e.g. engagement ended).
     public var disabled: ChatTextContentStyle
+    
+    public var disabledTabsBorderColor: UIColor
```
VladKupriienko, line 14: I don't really like how it sticks out from the general structure. This style only has three properties, with each of them only describing their corresponding state. I feel like that should stay this way. 

So, I suggest subclassing the `ChatTextContentStyle` in, say `ChoiceCardOptionStateStyle` - which will only have an additional property `borderColor`. Maybe the integrator would like to change border colors of all choice card options' states, who knows - the more customization, the better. 

You can have this property be `nil` by default in the initializer or instead be explicitly set to `nil` from each state initializer (I also suggest breaking them up, more on that in later comments). This would make more sense semantically as well - `ChoiceCardOptionStyle` is a collection of three separate `ChoiceCardOptionStateStyle`s.

VladKupriienko: ...and don't forget to add description (documentation) to the new class, its property and init 😉 
You can contact me if you struggle with how it should be worded.

GliaWidgets/View/Chat/Message/Content/Choice/ChoiceCardOptionStyle.swift
```diff
@@ -26,6 +28,7 @@ public final class ChoiceCardOptionStyle {
         normalBackgroundColor: UIColor,
         highlightedTextColor: UIColor,
         highlightedBackgroundColor: UIColor,
+        disabledTabsBorderColor: UIColor,
         disabledTextColor: UIColor,
         disabledBackgroundColor: UIColor
```
VladKupriienko, line 33: I honestly don't remember why I decided to have this initializer instead of one that accepts three separate state styles - maybe due to common `textFont`. But now that I look at it, I feel like we should instead have an init that accepts `normal`, `highlighted` and `disabled`.

GliaWidgets/View/Chat/Message/Content/Choice/ChoiceCardOptionView.swift
```diff
@@ -64,6 +64,7 @@ class ChoiceCardOptionView: UIView {
             applyStyle(style.selected)
         case .disabled:
             applyStyle(style.disabled)
+            setBorderLinesForTabs()
```
VladKupriienko, line 67: With the small refactoring I suggested, we should include the code for setting the border in `applyStyle`. If it's nil - do nothing, otherwise do what you are doing in `setBorderLinesForTabs()`

GliaWidgets/View/Chat/Message/Content/Choice/ChoiceCardOptionView.swift
```diff
@@ -74,6 +75,11 @@ class ChoiceCardOptionView: UIView {
         }
     }
 
+    private func setBorderLinesForTabs() {
+        layer.borderColor = style.disabledTabsBorderColor.cgColor
+        layer.borderWidth = 1
+    }
```
VladKupriienko, line 81: Please refer to the comment above :)

VladKupriienko (commented):
Your solution works, but I suggest refactoring the choice card styles a bit to:
1. Make them more customizable
2. Have cleaner and more logical properties and initializers

VladKupriienko (approved):

PR #123, opened 2021-09-28
[MUIC-462] Enable download files with the same name

KarimovaKarina:


GliaWidgets/Lib/Download/FileDownload.swift
```diff
@@ -33,10 +33,8 @@ class FileDownload {
     let file: ChatEngagementFile
 
     private var storageID: String? {
-        if let fileName = file.name {
-            return fileName
-        } else if let fileID = file.id {
-            return fileID
+        if let fileID = file.id, let fileName = file.name {
+            return fileID + "CODE:" + fileName
```
VladKupriienko, line 37: Let's use another separator. First of all, it would look better. Second, iOS doesn't allow for a semicolon in a filename (if you wish to save it to a device).
I suggest using `#` (or even better just `_`). The fileID seems to be just lowercase letters, numbers and dashes, so you won't have any interference if you'll just search for the first occurrence of this symbol to split the name.

GliaWidgets/Lib/File/LocalFile.swift
```diff
@@ -2,7 +2,15 @@ import UIKit
...
     lazy var fileExtension: String = { return url.pathExtension }()
-    lazy var fileName: String = { return url.lastPathComponent }()
+    lazy var fileName: String = {
+        let snippet = url.lastPathComponent
+        var name = ""
+        if let range = snippet.range(of: "CODE:") {
+            name = String(snippet[range.upperBound...])
+        }
+        return name
+
```
VladKupriienko, line 12: Redundant newline + trailing whitespace

GliaWidgets/Lib/File/LocalFile.swift
```diff
@@ -2,7 +2,15 @@ import UIKit
 
 class LocalFile {
     lazy var fileExtension: String = { return url.pathExtension }()
-    lazy var fileName: String = { return url.lastPathComponent }()
+    lazy var fileName: String = {
+        let snippet = url.lastPathComponent
+        var name = ""
+        if let range = snippet.range(of: "CODE:") {
```
VladKupriienko, line 8: Obviously, if you change the separator, do change the code here as well.

GliaWidgets/Lib/File/LocalFile.swift
```diff
@@ -2,7 +2,15 @@ import UIKit
 
 class LocalFile {
```
VladKupriienko, line 3: I don't know why it is the way it is, but there are no newlines between all the multiline `lazy var`s in this file. Please insert them for readability purposes (except for after `fileExtension`, as it fits in a single line).

VladKupriienko (commented):
Sorry for taking so long, only found some time to play with it recently.

In general, while it has its shortcomings, I think it is fine in the current state except for some minor changes I asked for. It gets the job done, fixes the bug it was intended to fix and goes one step further to at least show the real file name in the preview. 

Good job!

VladKupriienko:
Oh, I have also rebased the `qa/MUIC-462` branch on current `dev` because it's changed a bit since this PR was opened. So you'll also have fix the conflict that appeared :)

dukhovnyi:
I'd like to discuss this with you all about this solution. It looks works well and reverses compatible which is great 👍
But probably we could avoid creating this kind of hack with `CODE:` and create a directory with a file identifier, inside of this directory save the file with the filename. In this case, we will not need to worry about two files with the same name and could use the usual URL instead of `QuickLookPreviewItem`.
I made a small proposal, so if you are interested, please take a look at my idea. Just apply `proposal.diff` as a patch for this branch and review changes.

[proposal.patch.zip](https://github.com/salemove/ios-sdk-widgets/files/7296543/proposal.patch.zip)

kuldar-daniel:
I will take a look at this issue now. Sorry for the delay

GliaWidgets/Lib/Storage/FileSystemStorage.swift
```diff
@@ -33,7 +33,14 @@ class FileSystemStorage: DataStorage {
 
     func store(_ data: Data, for key: String) {
         let url = storageURL(for: key)
-        try? data.write(to: url)
+
+        if !fileManager.fileExists(atPath: url.deletingPathExtension().path) {
+            try! fileManager.createDirectory(at: url, withIntermediateDirectories: true, attributes: nil)
```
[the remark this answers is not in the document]
kuldar-daniel: Right, fixed

VladKupriienko (approved):

PR #128, opened 2021-09-30
[MUIC-508] Fix transition issues

VladKupriienko:
Widgets minimizing/maximizing animation that was implemented in #121 was working fine except for an issue where the maximizing view would not react to orientation changes that might have happened while the view was minimized.

This issue was resolved with the usage of `finalFrame(for:)` method of `UIViewControllerContextTransitioning`. Additionally, the transition controller code received a major overhaul that addressed the issues overlooked in the initial code review.

Also includes the removal of unused pieces of old animation from `GliaViewController` (and the entire `UIView+Screenshot.swift` file as a result).

As a very, very minor side change, SwiftGen's output of changing the chat history placeholder string is also included, as it was erroneously excluded from the original #124 PR. This should be in `dev` and I honestly don't want to create a separate PR or even commit to `dev` just for that 😄


VladKupriienko:
Changed the base branch directly to `dev` as this has already passed the QA.

dukhovnyi (approved):
Looks great 👍

PR #137, opened 2021-10-07
MUIC-379: Going from chat screen to audio screen shows weird animation on the background

kuldar-daniel:
This PR fixes weird default UINavigationController animation, when pushed ViewController background is transparent.
Basically we just override the default behaviour and add our own simple transition to combat the issue.

### DIFF
--- a/GliaWidgets/Lib/ViewController/NavigationController/NavigationController.swift
+++ b/GliaWidgets/Lib/ViewController/NavigationController/NavigationController.swift
@@ -1,3 +1,47 @@
 import UIKit
 
-class NavigationController: UINavigationController {}
+class NavigationController: UINavigationController {
+    private let transitionDuration: CFTimeInterval = 0.3
+
+    override func pushViewController(_ viewController: UIViewController, animated: Bool) {
+        if animated {
+            let transition = CATransition()
+            transition.duration = transitionDuration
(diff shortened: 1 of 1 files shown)


GliaWidgets/Lib/ViewController/NavigationController/NavigationController.swift
```diff
@@ -1,3 +1,47 @@
...
+        if animated {
+            let transition = CATransition()
+            transition.duration = transitionDuration
+            transition.timingFunction = CAMediaTimingFunction(name: .easeInEaseOut)
+            transition.type = .push
+            transition.subtype = .fromRight
+            view.layer.add(transition, forKey: nil)
+        }
+        
+        super.pushViewController(viewController, animated: false)
```
dukhovnyi, line 16: probably we could pass animated and don't pass manually `false` for the animated parameter. (the same for other methods in this class)
```suggestion
        super.pushViewController(viewController, animated: animated)
```

dukhovnyi (approved):
Looks good, just one minor suggestion.

GliaWidgets/Lib/ViewController/NavigationController/NavigationController.swift
dukhovnyi: oh, good catch. You are right. Thank you for clarification.

VladKupriienko (approved):

PR #141, opened 2021-10-11
[MUIC-493] Show alert when operator ended engagement

KarimovaKarina:
Before it was like this: if operator ends engagement, visitor doesn't see any notification about that, and engagement just ended. I created new structure SingleActionAlertConfiguration for alert, which has only one action button. When operator ends engagement, visitor gets the alert "Engagement ended", clicks button "OK" and only then engagement is being closed.

### DIFF
--- a/GliaWidgets/ViewController/Common/Alert/AlertViewController+SingleAction.swift
+++ b/GliaWidgets/ViewController/Common/Alert/AlertViewController+SingleAction.swift
@@ -1,6 +1,8 @@
 extension AlertViewController {
-    func makeSingleActionAlertView(with conf: SingleActionAlertConfiguration,
-                                   confirmed: @escaping () -> Void) -> AlertView {
+    func makeSingleActionAlertView(
+        with conf: SingleActionAlertConfiguration,
+        actionTapped: @escaping () -> Void
+    ) -> AlertView {
         let alertView = viewFactory.makeAlertView()
         alertView.title = conf.title
         alertView.message = conf.message
@@ -12,7 +14,7 @@ extension AlertViewController {
         button.title = conf.buttonTitle
         button.tap = { [weak self] in
             self?.dismiss(animated: true)
-            confirmed()
+            actionTapped()
         }
         alertView.addActionView(button)
 
--- a/GliaWidgets/ViewModel/EngagementViewModel.swift
+++ b/GliaWidgets/ViewModel/EngagementViewModel.swift
@@ -16,7 +16,7 @@ class EngagementViewModel {
         )
         case showSingleActionAlert(
             SingleActionAlertConfiguration,
-            confirmed: (() -> Void)?
+            actionTapped: (() -> Void)?
         )
         case showAlert(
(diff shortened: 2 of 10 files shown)


GliaWidgets/Theme/Alert/SingleActionAlertConfiguration.swift
```diff
@@ -0,0 +1,11 @@
+/// Configuration of the operator ending the engagement alert.
+public struct SingleActionAlertConfiguration {
+    /// Title of the alert.
+    public var title: String?
```
dukhovnyi, line 4: I'd like to clarify, do we need to have properties as `var` instead of `let`?
If we don't expect changing of alert configuration in runtime, let's make it as `let`.

GliaWidgets/ViewController/Common/Alert/AlertViewController+SingleAction.swift
```diff
@@ -0,0 +1,21 @@
...
+        alertView.message = conf.message
+        alertView.actionsAxis = .horizontal
+
+        let buttonStyle = viewFactory.theme.alert.positiveAction
+
+        let button = ActionButton(with: buttonStyle)
+        button.title = conf.buttonTitle
+        button.tap = { [weak self] in
+            self?.dismiss(animated: true)
+            confirmed()
```
dukhovnyi, line 15: imho, `confirmed` looks a little bit confused for this kind of action, since this is a one-action alert.
@kuldar-daniel, @VladKupriienko what do you think guys about this naming?

GliaWidgets/ViewModel/EngagementViewModel.swift
```diff
@@ -128,9 +132,13 @@ class EngagementViewModel {
...
             )
         case .ended:
-            if EngagementViewModel.alertPresenters.isEmpty {
-                engagementDelegate?(.finished)
-            }
+            EngagementViewModel.alertPresenters.insert(self)
+            engagementAction?(
+                .showSingleActionAlert(
+                    alertConfiguration.operatorEndedEngagement,
+                    confirmed: { self.endSession() }
```
dukhovnyi, line 139: Looks like `[weak self]` is required. Let's use weak reference here to avoid retain cycle.

GliaWidgets/ViewModel/EngagementViewModel.swift
```diff
@@ -50,6 +50,7 @@ class EngagementViewModel {
     private let screenShareHandler: ScreenShareHandler
     private(set) var isViewActive = ObservableValue<Bool>(with: false)
     private static var alertPresenters = Set<EngagementViewModel>()
+    private var isEngagementEnded = false
```
kuldar-daniel, line 53: Maybe just to clarify that engagement ended on operator side?
isEngagementEndedByOperator ?

KarimovaKarina: this var for the case when Visitor ends Engagement. so that we don't need to notify that operator ended engagement

PR #150, opened 2021-10-18
[MUIC-461] Integration guide + docs update

VladKupriienko:
This PR introduces the integration guide intended for the Wiki page and Glia Integration Documentation later on.

Additionally, it includes additional documentation for the parts where it was missing and minor updates to the existing docs.

The `populate_wiki` script can be used to automatically update the Wiki whenever anything in the `docs` folder is changed. This allows for better management of code and documentation, as all of the changes (code + relevant parts of the integration guide) will be included in the same commit and then automatically pushed to Wiki. 

The script also takes care of generating a sidebar for the Wiki. It supports 1 layer of sub-folders, making them into categories. It also supports ordering the pages in these categories via specific comments at the beginning of the markdown file:
`[order]: # (5)`

Currently, this script is triggered in Bitrise on each push to a `docs/*` branch (for testing purposes), but I will move this step into the master-build workflow when this gets merged. You can see the properly rendered guide in the Wiki already, as this branch had a `docs` prefix :)


GliaWidgets/Component/CallButtonBar/Button/CallButtonStyle.swift
```diff
@@ -2,7 +2,7 @@ import UIKit
 
 /// Style of a button shown in call view bottom button bar (i.e. "Chat", "Video", "Mute", "Speaker" and "Minimize").
 public struct CallButtonStyle {
-    /// Style of a call button ("Chat", "Video", "Mute", "Speaker" and "Minimize") in a specific state - active or inactive.
+    /// Style of a call button ("Chat", "Video", "Mute", "Speaker" and "Minimize") in a specific state - activated or not activated.
```
VladKupriienko, line 5: There was a confusion when we first wrote the documentation. Inactive does not mean that the button is disabled. To indicate that, the app takes an inactive state and makes it more transparent.
Instead it means whether the button is toggled on or off. This only relates to the Mute button and the Speaker button - if you are muted, the "Mute" button is in "active" state, i.e. it is toggled on; same for the speaker.

GliaWidgets/Glia.swift
```diff
@@ -34,84 +34,6 @@ public protocol SceneProvider: AnyObject {
 }
 
 /// Glia's engagement interface.
-///
```
VladKupriienko, line 37: I removed this from here because it was moved to the Integration guide. I think we have agreed on that previously, @kadriiiii  :)

docs/Customization/AlertStyle.md
```diff
@@ -0,0 +1,43 @@
...
+```
+
+It can then be assigned to the `alert` property of the `Theme`:
+```swift
+let theme = Theme()
+
+theme.alert = alertStyle
+```
+
+A more in-depth description of each individual property of the `AlertStyle` can be found in the documentation comments in Xcode itself.
```
dukhovnyi, line 29: probably `in source code` would be more correct than `in Xcode`

docs/Customization/ThemeColor.md
```diff
@@ -0,0 +1,60 @@
...
+theme.chat.unreadMessageIndicator.badge.backgroundColor = .brown
+theme.call.header.backgroundColor = .red
+theme.alert.positiveAction.backgroundColor = .green
+```
+
+**NB!** It is not possible to directly change the values of the `color` property of the `Theme`. To change the colors across the entire app, `ThemeColor`'s initializer should be used. If there is still a need to change only some specific colors, other parameters can be omitted. The omitted parameters will use the default values.
+ ```swift
+var theme = Theme()
+
+theme.color.baseNormal = .black // Not possible!
```
dukhovnyi, line 47: Do I understand correctly that the setter for `baseNormal` is accessible and it is possible to assign? But it is not expected from our side?
If so, probably we could change `Not possible!` to `Incorrect way!`?
```swift
theme.color.baseNormal = .black // Incorrect!
```

docs/General-notes.md
```diff
@@ -0,0 +1,21 @@
...
+For example, chat default look:
+
+<p align="center">
+  <img width="500" src="./images/chat_general_look.png">
+</p>
+
+To read more about customization, navigate to the [Creating a Theme](creating-a-theme) page and other pages in the Customization section.
+
+### UIImage Tint
+To be able to choose any icon's color (tint), the asset used should be set to "Render as: Template Image". That applies to every icon, color of which can be changed.
```
dukhovnyi, line 21: imho, we don't need to use brackets here, since we a talking about tint color only
```suggestion
To be able to choose any icon's tint color, the asset used should be set to "Render as: Template Image". That applies to every icon, color of which can be changed.
```

docs/Getting-started.md
```diff
@@ -0,0 +1,119 @@
...
+#### CocoaPods
+To integrate the GliaWidgets iOS SDK into your project via [CocoaPods](https://cocoapods.org/) modify your `Podfile`:
+```
+source 'https://github.com/CocoaPods/Specs.git'
+source 'https://github.com/salemove/glia-ios-podspecs.git'
+
+user_frameworks!
+
+target 'YOUR_TARGET_NAME' do
+    pod 'GliaWidgets', git: 'https://github.com/salemove/ios-sdk-widgets', tag: 'SDK_VERSION'
```
dukhovnyi, line 37: The user does not need to use the direct git link anymore :)
So we could update this line to
```suggestion
    pod 'GliaWidgets', 'SDK_VERSION'
```

docs/Getting-started.md
```diff
@@ -0,0 +1,119 @@
...
+        configuration: configuration,
+        queueID: "queueID",
+        visitorContext: visitorContext,
+        theme: theme
+    )
+} catch {
+    // Handle the configuration error
+}
+```
+
```
dukhovnyi, line 118: One more thing is setting up the Push Notification feature, do we need to add additional info about it?

kadriiiii (commented):
I have no good suggestions regarding the code comments - the developers have argued there and found the solutions.
But - it's now safe to remove the .md and .png files from this PR, as the content from those is moved to a Glia's integrator-docs repository PR https://github.com/salemove/integrator-docs/pull/256, @kuldar-daniel @raigo-pm

GliaWidgets/Component/MediaUpgrade/MediaUpgradeActionStyle.swift
```diff
@@ -1,18 +1,50 @@
 import UIKit
 
-// TODO: documentation
+/// Style of the media upgrade action in a multiple media type alert window. Currently unused.
```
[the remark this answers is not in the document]
kuldar-daniel: fixed

kuldar-daniel:
Removed docs

kadriiiii (approved):

PR #151, opened 2021-10-19
[MUIC-529] Show end engagement alert when view is maximized

KarimovaKarina:
In case when engagement view is minized , visitor couldn't see alert that operator ended engagement, even when returning back to engagement view (maximize). In this PR the "End engagemant" alert is shown when visitior maximize engagement view.

### DIFF
--- a/GliaWidgets/ViewModel/EngagementViewModel.swift
+++ b/GliaWidgets/ViewModel/EngagementViewModel.swift
@@ -106,7 +106,7 @@ class EngagementViewModel {
                 .showSingleActionAlert(
                     alertConfiguration.operatorEndedEngagement,
                     actionTapped: { [weak self] in
-                            self?.endSession()
+                        self?.endSession()
                     }
                 )
             )
@@ -161,7 +161,7 @@ class EngagementViewModel {
                     .showSingleActionAlert(
                         alertConfiguration.operatorEndedEngagement,
                         actionTapped: { [weak self] in
-                                self?.endSession()
+                            self?.endSession()
                         }
                     )
                 )


GliaWidgets/ViewModel/EngagementViewModel.swift
```diff
@@ -95,6 +99,20 @@ class EngagementViewModel {
...
     func viewDidAppear() {}
 
+    func viewWillAppear() {
+        if isEngagementEndedByOperator {
+            EngagementViewModel.alertPresenters.insert(self)
+            engagementAction?(
+                .showSingleActionAlert(
+                    alertConfiguration.operatorEndedEngagement,
+                    actionTapped: { [weak self] in
+                            self?.endSession()
```
dukhovnyi, line 109: probably I could be wrong but looks like an incorrect alignment of `self?.endSession()`.

GliaWidgets/ViewModel/EngagementViewModel.swift
```diff
@@ -147,6 +165,7 @@ class EngagementViewModel {
                         }
                     )
                 )
+                isEngagementEndedByOperator = true
```
dukhovnyi, line 168: Q: We change the `isEngagementEndedByVisitor` on the confirmation step after the session is ended, but here is before the alert has been presented. Is it correct behavior or do we need to restore consistency and do `isEngagementEndedByOperator = true` after `self?.endSession()` also?

```swift
confirmed: { [weak self] in
    self?.endSession()
    self?.isEngagementEndedByVisitor = true
}
```

KarimovaKarina: when visiotr taps on button "X" (close) and confirms that he wants to end engagement (=taps on "YES") then isEngagementEndedByVisitor = true. But in that case I thought that i don't need to stick isEngagementEndedByOperator variable to action when visitor taps OK button in alert. I just need to know when engagement is ended by operator. And if it was ended by operator i let View Controller to know that it was ended by operator

PR #166, opened 2021-11-02
Clear visitor session

dukhovnyi:
### The problem
In case of different sign-in in the final application, visitor history is shared between all application users.

### Hotfix Proposal
The SalemoveSDK has the feature for clearing visitor sessions, so we need additionally to clean up local sqlite3 db to avoid sharing chat history between sessions.


kuldar-daniel (approved):

dukhovnyi:
The proper implementation for this feature will be introduced in the future.

PR #171, opened 2021-11-10
MUIC-527: Starting an audio/video engagement as a visitor shows "Incoming Chat Engagement" in Glia Hub

kuldar-daniel:


KarimovaKarina (approved):

GliaWidgets/ViewModel/EngagementViewModel.swift
```diff
@@ -115,10 +115,14 @@ class EngagementViewModel {
...
-    func enqueue() {
-        interactor.enqueueForEngagement {} failure: { error in
-            self.handleError(error)
-        }
+    func enqueue(mediaType: MediaType) {
+        interactor.enqueueForEngagement(
+            mediaType: mediaType,
+            success: {},
+            failure: { error in
+                self.handleError(error)
```
dukhovnyi, line 123: since we pass self into closure, let's avoid memory leaking and use weak self here.

kuldar-daniel: good catch 😄

dukhovnyi (approved):

PR #173, opened 2021-11-10
MUIC-502: Remove "print" calls from the Widgets

kuldar-daniel:


KarimovaKarina (approved):

dukhovnyi (commented):
1. Do we have any follow-up ticket for introducing any kind of logs? I agree with removing `print` from the production code, but I believe we must have logs in the application at least for debug purposes
2. You have introduced and a new `GliaError` and I'd like to ask you do we have any requirements to throw an error in case of SDK has ongoing engagement already?

kuldar-daniel:
I'll create a ticket for logging in debug runs
No requirement per se for the engagement existing, but seeing as this is not currently handled and if we wan't to handle it would be very difficult with this approach (probably easier in the poc), I think it makes sense. Of course in ideal world, we'd be able to resume, maybe?

dukhovnyi:
> I'll create a ticket for logging in debug runs No requirement per se for the engagement existing, but seeing as this is not currently handled and if we wan't to handle it would be very difficult with this approach (probably easier in the poc), I think it makes sense. Of course in ideal world, we'd be able to resume, maybe?

yep, looks like we could resume the engagement just to call `rootCoordinator?.event(.maximized)`. But we could take a look at this more deeply in the future.

dukhovnyi (approved):

kuldar-daniel:
Maybe better to discuss it at refinement session before commiting changes to possibly not introduce new error we would have to deprecate later

dukhovnyi:
@kuldar-daniel  sorry for a long read under this simple PR, but I did some investigation and push one additional commit with swiftgen template that we could apply to this case. Please pull the latest commit and build the project to check the results. Perhaps it will be a better solution than the addition optionality and execution fatalError. Please ping me directly for some additional information/discussion. 
Thank you.

PR #196, opened 2021-11-30
Introduce interfaces for fetching and updating visitor info

igorkravchenko:
Currently widgets allow fetching and updating visitor info only after engagement has been started. This PR aims to resolve this by introducing interfaces that will setup the configuration but avoid starting the engagement.

MOB-1007

### DIFF
--- a/GliaWidgets/Glia.swift
+++ b/GliaWidgets/Glia.swift
@@ -122,16 +122,64 @@ public class Glia {
     public var onEvent: ((GliaEvent) -> Void)?
 
     private var rootCoordinator: RootCoordinator?
+    private var interactor: Interactor?
 
     private init() {}
 
-
     /// Setup SDK using specific engagement configuration without starting the engagement.
-    /// - Parameter configuration: Engagement configuration.
-    public func configure(with configuration: Configuration) throws {
-        try Salemove.sharedInstance.configure(appToken: configuration.appToken)
-        try Salemove.sharedInstance.configure(environment: configuration.environment.url)
-        try Salemove.sharedInstance.configure(site: configuration.site)
+    /// - Parameters:
+    ///   - configuration: Engagement configuration.
+    ///   - queueID: Queue identifier.
+    ///   - visitorContext: Visitor context.
+    public func configure(
+        with configuration: Configuration,
+        queueID: String,
+        visitorContext: VisitorContext
+    ) throws {
+        interactor = try Interactor(
+            with: configuration,
+            queueID: queueID,
+            visitorContext: visitorContext
+        )
+    }
+
+    /// Starts the engagement.
+    ///
+    /// - Parameters:
+    ///   - engagementKind: Engagement media type.
+    ///   - theme: A custom theme to use with the engagement.
+    ///   - visitorContext: Visitor context.
+    ///   - features: Set of features to be enabled in the SDK.
+    ///   - sceneProvider: Used to provide `UIWindowScene` to the framework. Defaults to the first active foreground scene.
+    ///
+    /// - throws:
+    ///   - `ConfigurationError.invalidSite`
+    ///   - `ConfigurationError.invalidEnvironment`
+    ///   - `ConfigurationError.invalidAppToken`
+    ///   - `GliaError.engagementExists`
+    ///
+    public func startEngagement(
+        engagementKind: EngagementKind,
+        theme: Theme = Theme(),
+        features: Features = .all,
+        sceneProvider: SceneProvider? = nil
+    ) throws {
+        guard engagement == .none else {
+            throw GliaError.engagementExists
+        }
+
+        guard let interactor = self.interactor else {
+            throw GliaError.sdkIsNotConfigured
+        }
+
+        let viewFactory = ViewFactory(with: theme)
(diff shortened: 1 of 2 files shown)


GliaWidgets/Glia.swift
```diff
@@ -223,4 +280,58 @@ public class Glia {
...
+    ///
+    /// If the request is unsuccessful for any reason then the completion will have an Error.
+    /// The Error may have one of the following causes:
+    ///
+    /// - `GeneralError.internalError`
+    /// - `GeneralError.networkError`
+    /// - `ConfigurationError.invalidSite`
+    /// - `ConfigurationError.invalidEnvironment`
+    /// - `ConfigurationError.invalidAppToken`
+    /// - `ConfigurationError.invalidApiToken`
```
dukhovnyi, line 301: Applicable for both interfaces:
1. Since these errors are part of Salemove SDK, we need to add Salemove namespace to avoid misunderstanding with integrators
```suggestion
    /// - `Salemove.GeneralError.internalError`
    /// - `Salemove.GeneralError.networkError`
    /// - `Salemove.ConfigurationError.invalidSite`
    /// - `Salemove.ConfigurationError.invalidEnvironment`
    /// - `Salemove.ConfigurationError.invalidAppToken`
    /// - `Salemove.ConfigurationError.invalidApiToken`
```

2. To be more consistent, I'd like to propose to handle case if `interactor` is not initialized yet. Does it make sense?
```swift
public func fetchVisitorInfo(completion: @escaping (Result<Salemove.VisitorInfo, Error>) -> Void) {
    guard let interactor = self.interactor else { 
        return completion(GliaError.sdkIsNotConfigured)
    }
    Salemove.sharedInstance.fetchVisitorInfo(completion)
}
```

GliaWidgets/Glia.swift
```diff
@@ -122,9 +122,66 @@ public class Glia {
...
     private init() {}
 
+    /// Setup SDK using specific engagement configuration without starting the engagement.
+    /// - Parameters:
+    ///   - configuration: Engagement configuration.
+    ///   - queueID: Queue identifier.
+    ///   - visitorContext: Visitor context.
+    public func configure(
+        with configuration: Configuration,
+        queueID: String,
```
dukhovnyi, line 136: I'd like to propose to use camelcase properly and rename `queueID` property to `queueId`.

dukhovnyi (commented):
One more thing, perhaps we need to deprecate the original `public func start(` method and/or replace the body. To be consistent and avoid code duplication.
```swift
public func start(
    _ engagementKind: EngagementKind,
    configuration: Configuration,
    queueID: String,
    visitorContext: VisitorContext,
    theme: Theme = Theme(),
    features: Features = .all,
) throws {
    try configure(
        with: configuration,
        queueID: queueID,
        visitorContext: visitorContext
    )

    try startEngagement(
        engagementKind: engagementKind,
        theme: theme,
        features: features,
        sceneProvider: sceneProvider
    )
}
```

igorkravchenko:
> One more thing, perhaps we need to deprecate the original `public func start(` method and/or replace the body. To be consistent and avoid code duplication.
> 
> ```swift
> public func start(
>     _ engagementKind: EngagementKind,
>     configuration: Configuration,
>     queueID: String,
>     visitorContext: VisitorContext,
>     theme: Theme = Theme(),
>     features: Features = .all,
>     sceneProvider: SceneProvider? = nil
> ) throws {
>     try configure(
>         with: configuration,
>         queueID: queueID,
>         visitorContext: visitorContext
>     )
> 
>     try startEngagement(
>         engagementKind: engagementKind,
>         theme: theme,
>         features: features,
>         sceneProvider: sceneProvider
>     )
> }
> ```
Good point. I went with the latter, reused methods without deprecation 43a2f2b600b254b64db08a672532a65d2d8a4936

GliaWidgets/Glia.swift
igorkravchenko: Totally makes sense 4e55cb9419abeedae735bfd35de63fdc889a57eb.

GliaWidgets/Glia.swift
igorkravchenko: Renamed 8b8964479107402cfc1109e51405c49891bc43d5.

GliaWidgets/Glia.swift
```diff
@@ -122,9 +122,70 @@ public class Glia {
...
 
     private var rootCoordinator: RootCoordinator?
+    private var interactor: Interactor?
 
     private init() {}
 
+    /// Setup SDK using specific engagement configuration without starting the engagement.
+    /// - Parameters:
+    ///   - configuration: Engagement configuration.
+    ///   - queueID: Queue identifier.
```
dukhovnyi, line 132:
```suggestion
    ///   - queueId: Queue identifier.
```

dukhovnyi (approved):
Looks good

GliaWidgets/Glia.swift
igorkravchenko: Updated f00262af5ec7092252388566b085767e924638d4.

kuldar-daniel (approved):

PR #199, opened 2021-12-08
Fix video not shown correctly on landscape orientation

gersonnoboa:
The video view behavior has changed for aspectFit scale, and now automatically stretches itself to cover the size of its container when rotating, while maintaining the aspect ratio. Thus, the only needed part from the widgets side is to have the video view cover the whole view bounds, center it, and change the scale to aspectFit.

Please note that the changes in the SDK haven't been published, so the correct behavior will not be visible yet. It will behave the same way as it has been so far. But I have tested these changes myself manually on an iPhone 8 Plus and an iPhone 12 Pro. Once the SDK is published to the new version you can confirm that it is working fine.

MUIC-320


dukhovnyi (approved):

kuldar-daniel (approved):

gersonnoboa:
!merge

PR #206, opened 2021-12-23
Introduce interface for supporting Site API Key

dukhovnyi:
1. Updated Core SDK dependency to 0.32.0
2. Introduced a new public interface for configuration SDK with Site API Key
3. Extended TestingApp with a switcher that can change authorization method (appToken or siteApiKey)

How to obtain [Site API Key](https://docs.glia.com/glia-support/docs/site-api-keys)

### DIFF
--- a/GliaWidgets/Configuration.swift
+++ b/GliaWidgets/Configuration.swift
@@ -1,3 +1,5 @@
+import SalemoveSDK
+
 /// Glia's environment. Use the one that our account manager has assigned to you.
 public enum Environment {
     /// Europe
@@ -7,14 +9,14 @@ public enum Environment {
     /// Beta environment. For development use.
     case beta
 
-    var url: String {
+    var region: Salemove.Region {
         switch self {
-        case .europe:
-            return "https://api.salemove.eu"
         case .usa:
-            return "https://api.salemove.com"
+            return .us
+        case .europe:
+            return .eu
         case .beta:
-            return "https://api.beta.salemove.com/"
+            return .custom(URL(string: "https://api.beta.salemove.com/")!)
         }
     }
 }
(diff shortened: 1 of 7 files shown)


GliaWidgets/Glia.swift
```diff
@@ -59,9 +126,9 @@ public class Glia {
     ///   - sceneProvider: Used to provide `UIWindowScene` to the framework. Defaults to the first active foreground scene.
     ///
     /// - throws:
-    ///   - `ConfigurationError.invalidSite`
-    ///   - `ConfigurationError.invalidEnvironment`
-    ///   - `ConfigurationError.invalidAppToken`
+    ///   - `SalemoveSDK.ConfigurationError.invalidSite`
+    ///   - `SalemoveSDK.ConfigurationError.invalidEnvironment`
+    ///   - `SalemoveSDK.ConfigurationError.invalidAppToken`
```
kuldar-daniel, line 131: I think it can also throw SalemoveSDK.invalidSiteApiKey & perhaps even SalemoveSDK.invalidRegionEndpoint? (same comment for other start method)

dukhovnyi: you are right - fixed in 0a19888

kuldar-daniel (approved):

GliaWidgets/Configuration.swift
```diff
@@ -7,26 +9,35 @@ public enum Environment {
...
+        } else {
+            return ""
+        }
+    }
     /// Deprecated.
     /// The current provided api token
     @available(*, deprecated, message: "Api token is not supported.")
     public let apiToken: String = ""
+
+    public let authorizationMethod: AuthorizationMethod
```
igorkravchenko, line 40: I suppose this needs to have documentation as well.

igorkravchenko (approved):
Looks good. Left one comment regarding documentation.

GliaWidgets/Configuration.swift
dukhovnyi: fixed - 7e6540f52dd9be35c623a4f2b0b34c922e67d49d

PR #207, opened 2022-01-03
Add SwiftPackage support

dukhovnyi:
https://glia.atlassian.net/browse/MUIC-642


GliaWidgets/BundleManaging/BundleManaging.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
+struct BundleManaging {
```
igorkravchenko, line 3: I would add documentation here stating why we introduce this struct, because this approach may be foreign to some of the maintainers.

igorkravchenko: Good move by the way 🙌

igorkravchenko (approved):
Looks good. Added suggestion regarding documentation.

GliaWidgets/BundleManaging/BundleManaging.swift
```diff
@@ -0,0 +1,15 @@
+import Foundation
+
```
kuldar-daniel, line 2: Would you not prefer the name BundleToken as SwiftGen uses?

kuldar-daniel (approved):

Package.swift
```diff
@@ -48,9 +76,14 @@ let package = Package(
             name: "GliaWidgetsSDK",
             dependencies: [
                 "SalemoveSDK",
-                "GliaWidgets",
+                "ReactiveSwift",
```
gersonnoboa, line 79: I think that we should explicitly mention this in our docs that integrators have the possibility to use these dependencies. As we rarely update these dependencies versions so maybe even the version would be useful and it shouldn't be a pain to maintain. Also, do they need to do something to use their own version of the same dependency without any issues? Both previous versions and more recent versions.

gersonnoboa (approved):
I think it woul

PR #212, opened 2022-01-17
Multiple dialogs

kuldar-daniel:


GliaWidgets/Coordinator/Engagement/EngagementViewModel.swift
```diff
@@ -0,0 +1,268 @@
...
+                engagementDelegate?(
+                    .alert(.unexpectedError)
+                )
+            }
+        default:
+            engagementDelegate?(
+                .alert(.unexpectedError)
+            )
+        }
+    }
```
dukhovnyi, line 257: perhaps we could improve this two switches and use only one like this
```swift
        switch error.error {
        case let queueError as QueueError where queueError == .queueClosed,
            let queueError as QueueError where queueError == .queueFull:
            engagementDelegate?(
                .alert(.operatorsUnavailable)
            )
        default:
            engagementDelegate?(
                .alert(.unexpectedError)
            )
        }
```

GliaWidgets/Interactor/MessageDispatcher.swift
```diff
@@ -0,0 +1,83 @@
...
+                default:
+                    break
+                }
+            })
+            .add(to: &disposables)
+    }
+
+    private func numberOfUnreadMessages() -> Int {
+        return [MessageState](messages.values)
+            .filter({ $0 == .received })
```
dukhovnyi, line 80:
```suggestion
            .filter { $0 == .received }
```

GliaWidgets/Lib/Extensions/UIViewController+Extensions.swift
```diff
@@ -8,3 +8,47 @@ extension UIViewController {
         viewController.didMove(toParent: self)
     }
 }
+
+extension UIViewController {
+    private final class DeallocatingProxy {
+        var deallocating: (() -> Void)?
```
dukhovnyi, line 14: since this is closure and it executes on deinit, I'd like to propose to name it `onDeinit`

GliaWidgets/Lib/Extensions/UIViewController+Extensions.swift
```diff
@@ -8,3 +8,47 @@ extension UIViewController {
...
+        deinit {
+            deallocating?()
+        }
+    }
+
+    private enum AssociatedKeys {
+        static var deallocatingProxy = "deallocatingProxy"
+    }
+
+    var deallocating: (() -> Void)? {
```
dukhovnyi, line 25: I'm sorry, but I missed a little bit the reason of using associated objects. Could you clarify this a little bit? Do we have any alternatives for it?

GliaWidgets/Theme/Alert/AlertConfiguration.swift
```diff
@@ -13,6 +13,7 @@ public struct AlertConfiguration {
     public var operatorsUnavailable: MessageAlertConfiguration
 
     /// Configuration of the media upgrade confirmation alert.
+    @available(*, deprecated, message: "Not used")
```
dukhovnyi, line 16: do we have another interface for setting up UI for mediaUpgrade alert?

GliaWidgets/Theme/Alert/ConfirmationAlertConfiguration.swift
```diff
@@ -13,6 +13,7 @@ public struct ConfirmationAlertConfiguration {
     public var positiveTitle: String?
 
     /// Indicates whether the action button's colors should be switched.
+    @available(*, deprecated, message: "Not used")
```
dukhovnyi, line 16: do we have another interface for setting up color for switchButtonBackgroundColors? is it about tint color?

GliaWidgets/Coordinator/Engagement/Alert/Properties/Item/AlertAction.swift
```diff
@@ -0,0 +1,11 @@
+struct AlertAction {
+    enum Style {
+        case regular
+        case destructive
+    }
+
+    let title: String
+    let style: Style
+
+    var handler: (() -> Void)?
```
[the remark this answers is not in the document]
kuldar-daniel: No, but this allows us to handle it straight in coordinator instead of referencing back the whole way through vm + delegates. Or do you have something else in mind?

GliaWidgets/Coordinator/Engagement/EngagementCoordinator.swift
```diff
@@ -0,0 +1,616 @@
...
+
+        case .videoCall:
+            return createCall(
+                ofKind: .video,
+                mediaDirection: .oneWay,
+                startAction: .startEngagement(mediaType: .video)
+            )
+
+        case .none:
+            fatalError("EngagementKind cannot be none.")
```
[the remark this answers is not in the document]
kuldar-daniel: what would I return as uiviewcontroller 🤔

GliaWidgets/Theme/Alert/AlertConfiguration.swift
kuldar-daniel: multiple media upgrade was never implemented, it's basically reduntant

GliaWidgets/Theme/Alert/ConfirmationAlertConfiguration.swift
kuldar-daniel: this is a one odd property, it was to hack positive action into negative themed one (end engagement confirmation yes is positive in the sense we should take action, but negative in the style, since it's destructive). Not sure if any of the integrators would ever use this. The new alerts handle it differently from the get go, eg. action can have style per whetherr it's a destrucive or normal. Hence this property will render useless. Or shuld we preserve the functionality?

GliaWidgets/Lib/Extensions/UIViewController+Extensions.swift
kuldar-daniel: Explained in Slack

PR #218, opened 2022-01-19
The video is not correctly shown in landscape mode

kuldar-daniel:
## Context

**Jira issue URL:** https://glia.atlassian.net/browse/MUIC-320
**Jira issue URL:** https://glia.atlassian.net/browse/MUIC-290

**Summary:**
Fixes video frame in landscape mode. Also allows local video to float around and be dragged across the screen.

### DIFF
--- a/GliaWidgets/View/Call/Video/VideoStreamView.swift
+++ b/GliaWidgets/View/Call/Video/VideoStreamView.swift
@@ -56,7 +56,7 @@ class VideoStreamView: UIView {
     }
 
     @objc
-    func pan(_ gesture: UIPanGestureRecognizer) {
+    private func pan(_ gesture: UIPanGestureRecognizer) {
         guard gesture.view != nil else { return }
 
         let translation = gesture.translation(in: self)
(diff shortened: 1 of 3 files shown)


GliaWidgets/View/Call/Video/VideoStreamView.swift
```diff
@@ -17,24 +20,47 @@ class VideoStreamView: View {
...
+        }
+        streamView.scale = kind == .remote ? .aspectFit : .aspectFill
         addSubview(streamView)
         streamView.autoPinEdgesToSuperviewEdges()
         streamView.layoutIfNeeded()
+        show?(true)
+    }
+
+    @objc
+    func pan(_ gesture: UIPanGestureRecognizer) {
```
dukhovnyi, line 59: perhaps we could make this mehod private since it is not necessary to expose it publically, or do we need it?

dukhovnyi (approved):

GliaWidgets/View/Call/Video/VideoStreamView.swift
kuldar-daniel: true, good catch

PR #224, opened 2022-02-08
Introduce DI instead of singleton and side-effects for testability

igorkravchenko:
This PR introduces a way to inject and control dependencies instead of using them directly to facilitate testability of the SDK.
Note that not all dependencies are covered: specifically ChatStorage still needs to be reworked in order to have dependencies injected. However since there are quite a lot of changes already, it makes sense to review them first and decide if this approach is clear and appropriate. If it is, then other dependencies will be managed in the upcoming PR(s).

MOB-1120


dukhovnyi (approved):
Looks amazing 🎉

kuldar-daniel (approved):

PR #225, opened 2022-02-09
Prepare ChatStorage for using inside of unit-tests

dukhovnyi:


kuldar-daniel:
Would it not have made more sense to make ChatStorage protocol + implementation?

dukhovnyi:
> Would it not have made more sense to make ChatStorage protocol + implementation?

to keep it consistent I decided to follow the same approach as other implementations. One more thing is that we will remove local storage from Widgets SDK at all, so this mechanism is some kind of temporary solution.

igorkravchenko (approved):
Looks good - ChatStorage will be removed soon, so we just treat it as external dependency for now: we will be able to mock it, but we will not write tests to check its internal logic.

kuldar-daniel (approved):

PR #232, opened 2022-02-17
Fix incorrect message in chat after response on the choice card

dukhovnyi:
MOB-1147

### DIFF
--- a/GliaWidgets/ViewModel/Chat/ChatViewModel.swift
+++ b/GliaWidgets/ViewModel/Chat/ChatViewModel.swift
@@ -743,9 +743,10 @@ extension ChatViewModel {
 // MARK: Choice Card
 
 extension ChatViewModel {
-    private func sendChoiceCardResponse(_ option: ChatChoiceCardOption, to messageId: String) {
-        guard let value = option.value else { return }
-        environment.sendSelectedOptionValue(value) { [weak self] result in
+    func sendChoiceCardResponse(_ option: ChatChoiceCardOption, to messageId: String) {
+
+        guard let option = option.asSingleChoiceOption() else { return }
+        environment.sendSelectedOptionValue(option) { [weak self] result in
             guard let self = self else { return }
 
             switch result {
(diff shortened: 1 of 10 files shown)


kuldar-daniel (approved):

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -744,8 +744,9 @@ extension ChatViewModel {
 
 extension ChatViewModel {
     private func sendChoiceCardResponse(_ option: ChatChoiceCardOption, to messageId: String) {
-        guard let value = option.value else { return }
-        environment.sendSelectedOptionValue(value) { [weak self] result in
+
+        guard let option = option.asSingleChoiceOption() else { return }
+        environment.sendSelectedOptionValue(option) { [weak self] result in
```
igorkravchenko, line 749: Let's add test ensuring that new `environment.sendSelectedOptionValue` is called. I suppose the easiest way to test it is to make `sendChoiceCardResponse` internal instead of `private`.

dukhovnyi: I added a lot of mocks, but now this case covered by test :)
fixed - 5f91fc53ba4437584fe7446ecf1f42ecc2dd4ca2

GliaWidgetsTests/Sources/ChatViewModelTests.swift
```diff
@@ -0,0 +1,65 @@
...
+            environment: .init(
+                chatStorage: .failing,
+                fetchFile: { _, _, _ in },
+                sendSelectedOptionValue: { _, _ in
+                    calls.append(.sendSelectedOptionValue)
+                },
+                uploadFileToEngagement: { _, _, _ in },
+                fileManager: fileManager,
+                data: .failing,
+                date: { Date() }
```
igorkravchenko, line 43: I would like to suggest to use mock, like Date.mock or something similar if possible. This is because date is a side-effect, and if for example we are to have some snapshot-based tests, it can cause problems.

dukhovnyi: fixed - 26afd5001506dff4d7261fb2783e8d0ba19df698

PR #233, opened 2022-02-17
Introduce mechanism for ending engagement

dukhovnyi:
MOB-1031


GliaWidgets/Glia.swift
```diff
@@ -285,4 +285,18 @@ public class Glia {
...
+    public func endEngagement(_ completion: @escaping (Result<Void, SalemoveError>) -> Void) {
+
+        interactor?.endSession(
+            success: {
+                completion(.success(()))
+            }, failure: {
+                completion(.failure($0))
+            }
+        )
+        onEvent?(.ended)
```
igorkravchenko, line 299: While Interactor needs to be reworked in order to be mocked (this will be address some other time), for  now we can make sure that onEvent?(.ended) is called using unit test. Also in test we can make sure that nil is set to rootCoordinator (it has to become `internal` instead of `private`).

igorkravchenko (changes requested):
One note regarding unit-tests, other than that - looks good.

GliaWidgets/Glia.swift
dukhovnyi: fixed 8743899cc91de4628974e20a922b41759ba1590c

igorkravchenko (approved):
Looks good! 👍

PR #234, opened 2022-02-21
File Upload - Visitor can upload more than 25 files after removing of last uploaded file

kuldar-daniel:
This PR fixes issue, where removing the file would remove all of the uploads where the local file is the same, which messes up the file count. Add UUID to each individual upload for them to be different.

### DIFF
--- a/GliaWidgets/Lib/Upload/FileUpload.swift
+++ b/GliaWidgets/Lib/Upload/FileUpload.swift
@@ -1,3 +1,5 @@
+import Foundation
+
 class FileUpload {
     enum Error {
         case fileTooBig
@@ -53,6 +55,7 @@ class FileUpload {
 
     private let storage: DataStorage
     private let environment: Environment
+    private let uuid = UUID()
 
     init(with localFile: LocalFile, storage: DataStorage, environment: Environment) {
         self.localFile = localFile
@@ -94,7 +97,7 @@ class FileUpload {
 
 extension FileUpload: Equatable {
     static func == (lhs: FileUpload, rhs: FileUpload) -> Bool {
-        return lhs.localFile == rhs.localFile
+        return lhs.uuid == rhs.uuid
     }
 }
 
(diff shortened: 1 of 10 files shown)


dukhovnyi (commented):
Sorry, I didn't catch how local file comparison affects file count limitation.
Perhaps we need to analyze the mechanism that counts the allowed file amount and rework it.
And in sense of adding unit-tests, I'd like to propose adding tests for this issue also.

kuldar-daniel:
> Sorry, I didn't catch how local file comparison affects file count limitation. Perhaps we need to analyze the mechanism that counts the allowed file amount and rework it. And in sense of adding unit-tests, I'd like to propose adding tests for this issue also.

`uploads.removeAll(where: { $0 == upload })`

when tapping close, all of the ones with same local file would be removed, but the UI would still display them.

GliaWidgets/Lib/Upload/FileUpload.swift
```diff
@@ -53,6 +55,7 @@ class FileUpload {
 
     private let storage: DataStorage
     private let environment: Environment
+    private let uuid = UUID()
```
igorkravchenko, line 58: Let's use DI for bringing generation of UUID here.

igorkravchenko (commented):
I support @dukhovnyi on this, that it looks incorrect to assume equitability based only on UUID and ignoring other fields, even if it solves current issue for now. Also it will be beneficial if we add unit test(s) that will prove that this issue is no longer exists.

kuldar-daniel:
Added test + removed the fileupload equatable protocol alltogether

igorkravchenko (approved):
Great job! Looks good 👍

GliaWidgetsTests/Lib/Upload/FileUploader.Environment.Failing.swift
```diff
@@ -0,0 +1,16 @@
+@testable import GliaWidgets
+extension FileUploader.Environment {
```
dukhovnyi, line 2:
```suggestion

extension FileUploader.Environment {
```

dukhovnyi (approved):

PR #240, opened 2022-02-25
Introduce separate target for snapshot based testing

igorkravchenko:
This PR aims to introduce another target to be used for image based snapshot unit tests. It also adds GIt LFS attributes for tracking PNG images located in `__Snapshots__` and its child directories. The reason for this target to be used separately is to be able to run image based snapshot tests independently from regular tests and text based snapshot tests, because image based snapshots require specific simulator(s). In addition it removes host application setting from GliaWidgetsTests target.


Podfile
```diff
@@ -21,7 +21,7 @@ target 'GliaWidgets' do
   pod 'lottie-ios', '3.2.3'
 end
 
-target 'GliaWidgetsTests' do
+target 'SnapshotTests' do
```
kuldar-daniel, line 24: We shouldn't lose regular tests either, right?

igorkravchenko: Yes, regular tests still run in GliaWidgetsTests. It's even possible to use text based snapshot testing for GliaWidgetsTests, but for now I decided to avoid including it in GliaWidgetsTests to keep things simple.

kuldar-daniel (approved):

dukhovnyi (approved):
👍

PR #241, opened 2022-02-28
Multiple dialogs #1 - Navigation refactor

kuldar-daniel:
This pull request attempts to split multiple dialogs into multiple smaller pieces. This is the first one. Main objectives of these changes:
* Simply and automatically memory managed Coordinator
* Build something above chat and call that could be used to present Alerts on in the future (EngagementNavigationController). Currently, alerts are shown on Call/Chat vc, but in case of multiple dialog handling we would like to show them regardless of whats currently on the screen.
* Simplify navigation logic and memory management. No point in creating call + chat together when only call should be displayed on the screen.


GliaWidgets/Lib/Coordinator/UIViewControllerCoordinator.swift
```diff
@@ -0,0 +1,75 @@
+import UIKit
+
+class UIViewControllerCoordinator: Coordinator {
```
igorkravchenko, line 3: Since this is the base class for some Coordinators, let's ensure that it functions the way it is expected by adding tests. Then if/when internal logic will change, we will have guaranties it still functions as expected.

kuldar-daniel: Will do

GliaWidgets/Lib/Coordinator/UIViewControllerCoordinator.swift
```diff
@@ -0,0 +1,75 @@
...
+    /// Memory management is automatic, meaning, if viewController gets deallocated
+    /// coordinator gets freed from memory as well. This is useful in cases where
+    /// we don't know when UIViewController will be deallocated (UINavigationController back
+    /// button press, dismiss modal controller with swipe down gesture)
+    func coordinate<T: UIViewControllerCoordinator>(to coordinator: T) -> UIViewController {
+        children[coordinator.id] = coordinator
+
+        let viewController = coordinator.start()
+
+        viewController.onDeinit = { [weak self] in
```
igorkravchenko, line 21: Let's move deallocation handling to separate method. Since the main goal of this approach is to detect when view controller has gone off screen, we will be able to iterate on it separately, to improve it if possible.

kuldar-daniel:
Added tests

dukhovnyi (commented):
Just wanted to clarify, would we like to use the associated object?

GliaWidgets/Coordinator/Engagement/Chat/ChatCoordinator.swift
```diff
@@ -114,7 +104,12 @@ class ChatCoordinator: SubFlowCoordinator, FlowCoordinator {
                 self?.presentWebViewController(with: url)
             }
         }
-        return ChatViewController(viewModel: viewModel, viewFactory: viewFactory)
+
+        let viewController = ChatViewController(viewModel: viewModel, viewFactory: viewFactory)
+
```
igorkravchenko, line 109: I suppose newlines here aren't necessary.

GliaWidgets/Lib/Coordinator/UIViewControllerCoordinator.swift
```diff
@@ -0,0 +1,89 @@
+import UIKit
+
+class UIViewControllerCoordinator: Coordinator {
+    let id = UUID()
```
igorkravchenko, line 4: Let's inject UUID.

kuldar-daniel:
@dukhovnyi I still think it is by far the simplest solution to all use cases of memory management. We have used this specific coordinator once before in project and the Rx one which basically does the same in several without any issues.

GliaWidgetsTests/Lib/Coordinator/UIViewControllerCoordinatorTests.swift
```diff
@@ -0,0 +1,19 @@
...
+        let c1: UIViewControllerCoordinator = .mock
+        let c2: UIViewControllerCoordinator = .mock
+
+        var vc = c1.coordinate(to: c2)
+
+        // Store
+        XCTAssertEqual([c2.id: c2], c1.children)
+        // Free on vc deallocation
+        vc = UIViewController()
+        DispatchQueue.main.async {
```
igorkravchenko, line 15: Is this necessary? DispatchQueue.main.async is side effect which should be avoided in unit tests.

GliaWidgetsTests/Lib/Coordinator/UIViewControllerTests.swift
```diff
@@ -0,0 +1,19 @@
...
+        var vc1 = UIViewController()
+        var didDeinit = false
+
+        vc1.onDeinit = {
+            didDeinit = true
+        }
+
+        vc1 = UIViewController()
+
+        DispatchQueue.main.async {
```
igorkravchenko, line 15: Is it possible to get rid of `DispatchQueue.main.async` here?

GliaWidgetsTests/Lib/Coordinator/UIViewControllerCoordinatorTests.swift
kuldar-daniel: Yes. without it, every n-th test would fail. Like 1 in 10 or 20 or something. All debugging I did indicated that the deallocation might not happen straight away (maybe ARC isn't sync?). But even waiting a little bit etc makes the test pass

GliaWidgets/Lib/Coordinator/UIViewControllerCoordinator.swift
kuldar-daniel: Fixed in bed95bc30c5eff03a774fed8a670c891b962a7e9

GliaWidgets/Coordinator/Engagement/Chat/ChatCoordinator.swift
kuldar-daniel: Fixed in 77b0821a9972ef725218919feaf740a2d35434f6

GliaWidgetsTests/Lib/Coordinator/UIViewControllerCoordinatorTests.swift
igorkravchenko: I think we need to look deeper into this. Let me know if you need help to find out what is the reason for this.

GliaWidgetsTests/Mocks/UIViewControllerCoordinator.Mock.swift
```diff
@@ -0,0 +1,20 @@
+@testable import GliaWidgets
+import UIKit
+
+extension UIViewControllerCoordinator {
+    static let mock: UIViewControllerCoordinator = Mock()
```
igorkravchenko, line 5: This will create one instance and will keep it in memory thus providing same instance.
 I think it should be a method that will produce new instance every time.

igorkravchenko (commented):
Left some comments regarding `UIViewControllerCoordinator`

GliaWidgetsTests/Mocks/UIViewControllerCoordinator.Mock.swift
kuldar-daniel: That is very good point that I missed

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -808,6 +807,6 @@ extension ChatViewModel {
 
     enum StartAction {
         case startEngagement
-        case none
+        case none(call: Call?)
```
dukhovnyi, line 810: I'm sorry but I missed the sense/name for this case. StartAction is none with optional Call instance. For what purposes did you create this case? perhaps we could think about a more convenient name for this case?
Based on switch above this case is needed for call... but why did you name it `none`? (just try to understand your thinking).

GliaWidgets/ViewController/Engagement/EngagementNavigationController.swift
```diff
@@ -34,13 +40,29 @@ class NavigationController: UINavigationController {
...
+            let transition = CATransition()
+
+            transition.duration = type(of: self).duration
+            transition.timingFunction = CAMediaTimingFunction(name: .easeInEaseOut)
+            transition.type = .fade
+
+            view.layer.add(transition, forKey: nil)
+        }
+
+        super.setViewControllers(viewControllers, animated: false)
```
dukhovnyi, line 66: do we need to pass `animated` to super or did you specific reasons to don't pass animated to parent?
```suggestion
        super.setViewControllers(viewControllers, animated: animated)
```

GliaWidgets/Coordinator/Engagement/EngagementCoordinator.swift
```diff
@@ -0,0 +1,274 @@
...
+            environment: .init(
+                uuid: environment.uuid
+            )
+        )
+
+        setup()
+    }
+
+    override func start() -> UIViewController {
+        defer {
```
dukhovnyi, line 53: why do we ignore `super.start()` here?

GliaWidgetsTests/Mocks/UIViewControllerCoordinator.Mock.swift
```diff
@@ -0,0 +1,20 @@
...
+        init() {
+            super.init(
+                environment: .init(
+                    uuid: { UUID() }
+                )
+            )
+        }
+
+        override func start() -> UIViewController {
+            UIViewController()
```
dukhovnyi, line 17: do we need to call `super.start` here or would we like to test it without base implementation?

GliaWidgets/ViewController/Engagement/EngagementNavigationController.swift
kuldar-daniel: No. If animated is true we add our own transition as animation and therefore we don't want the uikit's animation that the animated property would provide

GliaWidgets/Coordinator/Engagement/EngagementCoordinator.swift
kuldar-daniel: super start would only call fatalerror, so in this instance the override without super is intentional for all uiviewcontrollercoordinator subclasses.

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
kuldar-daniel: it was named before me, it basically means "don't do anything, since engagement is already ongoing"

GliaWidgetsTests/Mocks/UIViewControllerCoordinator.Mock.swift
kuldar-daniel: Same as above. There was a reason that the coordinator needs to be a class and not a protocol, I forgot exactly what it was.

GliaWidgetsTests/Lib/Coordinator/UIViewControllerTests.swift
kuldar-daniel: Fixed

GliaWidgetsTests/Lib/Coordinator/UIViewControllerCoordinatorTests.swift
kuldar-daniel: Fixed here. Calling onDeinit manually to simulate viewcontroller deallocation call as you suggested

igorkravchenko (approved):
Looks solid at this point. I approve it so we could move on further, but I would like to note that relying on deinit + `objc_setAssociatedObject` etc, is not desirable solution and needs to be reconsidered in separate PR.
@dukhovnyi @EgorovEI  what do you think about it?

dukhovnyi (approved):
I agree with you @igorkravchenko and I believe it is totally ok to approve this for now and have this feature implemented. And in the future to rework this solution and avoid using runtime features like associated objects and swizzling if possible.

PR #248, opened 2022-03-14
Visitor wants to know when they are put on hold

kuldar-daniel:
This feature allows visitor to see when the are put on hold. The cue is:
* In Call view the "on hold" text appears under operator name
* In Call view the bottom text lets visitor know they can continue browsing while on hold
* In Call remote video stream is hidden and operator image is shown again
* In Call view the the operator image displays blurred overlay with on hold icon
* In Minimzed Bubble the same blurred overlay is also displayed

MUIC-241


GliaWidgets/Component/Connect/Operator/ConnectOperatorView.swift
```diff
@@ -88,6 +98,24 @@ class ConnectOperatorView: UIView {
         animationView = nil
     }
 
+    func showOnHoldView() {
+        guard let style = style.onHoldOverlay else { return }
+
+        let onHoldView = OnHoldOverlayView(
+            style: style,
+            imageSize: .init(width: 40, height: 40)
```
igorkravchenko, line 106: Let's move image size to dedicated constant.

GliaWidgets/View/Call/CallView.swift
```diff
@@ -7,8 +7,40 @@ class CallView: EngagementView {
...
+    }
+
+    var isVisitrOnHold: Bool = false {
+        didSet {
+            connectView.operatorView.isVisitorOnHold = isVisitrOnHold
+
+            if isVisitrOnHold {
+                secondLabel.text = style.onHoldTopText
+                connectView.statusView.setSecondText(style.onHoldTopText, animated: false)
+            }
```
igorkravchenko, line 26: What happens for `else` case? Does `secondLabel.text` and  `connectView.statusView.setSecondText` still has  text assigned?

GliaWidgets/View/Call/CallView.swift
```diff
@@ -7,8 +7,40 @@ class CallView: EngagementView {
...
+
+            if case .video = mode {
```
igorkravchenko, line 28: Do we to need address other modes along with `.video`?

GliaWidgets/View/Call/CallView.swift
```diff
@@ -7,8 +7,40 @@ class CallView: EngagementView {
...
+                connectView.isHidden = !isVisitrOnHold
+                topStackView.isHidden = isVisitrOnHold
+                remoteVideoView.isHidden = isVisitrOnHold
```
igorkravchenko, line 31: Is it possible to add test that will ensure that all theses flags are flipped as expected?

GliaWidgets/View/Call/CallView.swift
```diff
@@ -7,8 +7,40 @@ class CallView: EngagementView {
...
+            }
+
+            bottomLabel.text = isVisitrOnHold
+                ? style.onHoldBottomText
+                : nil
+
+            bottomLabel.isHidden = !isVisitrOnHold
```
igorkravchenko, line 38: Also handle this case in tests as well?

GliaWidgets/ViewModel/Call/CallViewModel.swift
```diff
@@ -98,6 +100,12 @@ class CallViewModel: EngagementViewModel, ViewModel {
         call.duration.addObserver(self) { [weak self] duration, _ in
             self?.onDurationChanged(duration)
         }
+        call.isVisitorOnHold.addObserver(self) { [weak self] isOnHold, oldValue in
```
igorkravchenko, line 103: ViewModels seem to be like a good candidate for unit testing. Let's discuss how we can add tests here. Seems like `voice-over` branch has a lot of helpful mocks and Environments ready to be uses for testing.

GliaWidgets/View/Call/CallView.swift
kuldar-daniel: It only applies to video since the video mode has the connectview hidden and remote video stream visible.

GliaWidgets/View/Call/CallView.swift
kuldar-daniel: added else clause. It wasn't strictly needed due to the duration text would then be shown, but for claritys sake I added the duration text on the else

GliaWidgets/Component/Connect/Operator/ConnectOperatorView.swift
kuldar-daniel: the size varies depending whether it is in bubble or in call view though

GliaWidgets/Component/Bubble/BubbleStyle.swift
```diff
@@ -6,16 +6,21 @@ public class BubbleStyle {
     /// Style of a badge shown on the bubble.
     public var badge: BadgeStyle?
 
+    /// Style of a visitor's on hold state overlay.
+    public var onHoldOverlay: OnHoldOverlayStyle?
```
dukhovnyi, line 10: Perhaps I miss something, however I'd like to clarify why `onHoldOverlay` is optional and is variable?
For me, it looks like if this style is nil then `bubble` view will not show on hold state at all - [code](https://github.com/salemove/ios-sdk-widgets/pull/248/files#diff-811f3bc73367f56c237557e63c8e38abab3438cf50662d4dfea7ac21da63c7b7R59).
If we would like to apply default value, I think it should be some default `OnHoldOverlayStyle` and it should not lead to incorrect appearance of Bubble view.
Does it make sense?

GliaWidgets/Component/Bubble/BubbleView.swift
```diff
@@ -44,6 +54,27 @@ class BubbleView: UIView {
         ).cgPath
     }
 
+    func showOnHoldView() {
+        guard
```
dukhovnyi, line 58: Since `show/hide` methods are not synced and we don't have tests for it, I'd like to propose to call `onHoldView?.removeFromSuperview()` on the beginning of this method, to be sure that after accidentally called `showOnHoldView` we will not have an additional `onHoldView` subview.

GliaWidgets/Component/Bubble/BubbleView.swift
```diff
@@ -44,6 +54,27 @@ class BubbleView: UIView {
...
+            let style = style.onHoldOverlay,
+            let userImageView = userImageView
+        else { return }
+
+        let onHoldView = OnHoldOverlayView(
+            style: style,
+            imageSize: .init(width: 26, height: 26)
```
dukhovnyi, line 65: does it make sense to put this size value inside of style?

GliaWidgets/Component/Connect/Operator/ConnectOperatorStyle.swift
```diff
@@ -8,16 +8,22 @@ public struct ConnectOperatorStyle {
...
     public var animationColor: UIColor
 
+    /// Style of the visitor on hold overlay view.
+    public var onHoldOverlay: OnHoldOverlayStyle?
+
     ///
     /// - Parameters:
     ///   - operatorImage: Style of the operator's image.
     ///   - animationColor: Color of the animated concentric circles extending from the operator's image.
+    ///   -
```
dukhovnyi, line 18: looks like the doc is missed here.

GliaWidgets/Component/Connect/Operator/ConnectOperatorView.swift
```diff
@@ -88,6 +98,24 @@ class ConnectOperatorView: UIView {
         animationView = nil
     }
 
+    func showOnHoldView() {
+        guard let style = style.onHoldOverlay else { return }
```
dukhovnyi, line 102: I'd like to propose to call on the beginning of this method `onHoldView?.removeFromSuperview()`. To be sure that accidentally called `showOnHoldView` will not increment amount of onHoldView subviews.

GliaWidgets/Component/OnHoldOverlay/OnHoldOverlayStyle.swift
```diff
@@ -0,0 +1,14 @@
+import UIKit
+
+public struct OnHoldOverlayStyle {
```
dukhovnyi, line 3: Since we introduce a new style, I'd like to propose to avoid extending the global namespace and introduce this structure inside of Theme namespace for example.
What do you think, does it make sense?
cc: @igorkravchenko

GliaWidgets/Component/OnHoldOverlay/OnHoldOverlayStyle.swift
```diff
@@ -0,0 +1,14 @@
+import UIKit
+
```
dukhovnyi, line 2: let's add documentation for this structure.

GliaWidgets/Component/OnHoldOverlay/OnHoldOverlayView.swift
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    private func setup() {
+        imageView.image = style.image.withRenderingMode(.alwaysTemplate)
+        imageView.tintColor = style.imageColor
+    }
+
+    private func layout() {
+        addSubview(blurEffectView)
+        blurEffectView.autoPinEdgesToSuperviewEdges()
```
dukhovnyi, line 35: perhaps I miss something, but for me it looks like `blurEffectView` will be placed under the imageView and I don't see any re-ordering here. What is the benefit of this `blurEffectView` view?

GliaWidgets/Component/OnHoldOverlay/OnHoldOverlayVisualEffectView.swift
```diff
@@ -0,0 +1,23 @@
...
+        effect = nil
+        animator.stopAnimation(true)
+        animator.addAnimations { [weak self] in
+            self?.effect = UIBlurEffect(style: .regular)
+        }
+        animator.fractionComplete = 0.1
+    }
+
+    deinit {
+        animator.stopAnimation(true)
```
dukhovnyi, line 21: why do we need to stop animation explicitly on deinitialization? will it cause any issues if we will deinit it without stopping the animator?

GliaWidgets/View/Call/CallStyle.swift
```diff
@@ -44,6 +44,10 @@ public class CallStyle: EngagementStyle {
     /// Style of the call view bottom button bar (with buttons like "Chat", "Video", "Mute", "Speaker" and "Minimize").
     public var buttonBar: CallButtonBarStyle
 
+    /// The text shown in the top section of the view when visitor is put on hold
+    public var onHoldTopText: String?
+    public var onHoldBottomText: String?
```
dukhovnyi, line 49: Why do we need to have these two strings optional? If these strings are nil visitors can miss context what happened with UI.

GliaWidgets/Component/OnHoldOverlay/OnHoldOverlayStyle.swift
kuldar-daniel: This is consitent to what we have with other components. While I don't mind it being in Theme generally, I think consistency would win here until we would refactor the themeing

GliaWidgets/Component/OnHoldOverlay/OnHoldOverlayView.swift
kuldar-daniel: imageView is for the pause icon asset. The whole overlay view will be placed on top of the user image view, which will blur the user image, but not the pause icon asset on top of it

GliaWidgets/Component/OnHoldOverlay/OnHoldOverlayVisualEffectView.swift
```diff
@@ -0,0 +1,23 @@
...
+    private var animator = UIViewPropertyAnimator(
+        duration: 1,
+        curve: .linear
+    )
+
+    override func draw(_ rect: CGRect) {
+        super.draw(rect)
+
+        effect = nil
+        animator.stopAnimation(true)
```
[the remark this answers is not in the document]
kuldar-daniel: This is used to make the intensity of default iOS blur a little less to achieve the desired effect in the design. Default blureffectview would make the imageview that is blurred very odd and this seemed like a better solution than to make my own blurring of image

GliaWidgets/View/Call/CallStyle.swift
kuldar-daniel: this was done due to backward compatiblity. I saw that similar approach was used previously while developing new features. But default values are actually quite good alternative tht you proposed and I didn't think of it myself. WIll fix

GliaWidgets/Component/Bubble/BubbleView.swift
kuldar-daniel: That could work.

GliaWidgets/Component/Bubble/BubbleStyle.swift
kuldar-daniel: I think the same reason why badge style is optional. It was developed after and we didn't want to force integrators to implement them with errors. But default value is very good idea.

GliaWidgets/Component/OnHoldOverlay/OnHoldOverlayVisualEffectView.swift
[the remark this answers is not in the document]
dukhovnyi: so, it is expected that we create an animator but don't use it, or do we start animation in some other place?

GliaWidgets/View/Call/CallStyle.swift
```diff
@@ -83,7 +91,9 @@ public class CallStyle: EngagementStyle {
         bottomText: String,
         bottomTextFont: UIFont,
         bottomTextColor: UIColor,
-        buttonBar: CallButtonBarStyle
+        buttonBar: CallButtonBarStyle,
+        onHoldTopText: String,
+        onHoldBottomText: String
```
dukhovnyi, line 96: Since we have a pretty big constructor already, perhaps it makes sense to create a separate structure `OnHold` and place there two properties for now?

GliaWidgets/Component/OnHoldOverlay/OnHoldOverlayVisualEffectView.swift
kuldar-daniel: Yes, in case of incorrect memory usage where the view would stay alive (eg. retain cycle), this at least ensures that animation will be ended properly instead of running and wasting resources forever

GliaWidgets/View/Call/CallView.swift
kuldar-daniel: Added tests

GliaWidgets/View/Call/CallView.swift
kuldar-daniel: Added tests in fdeef7bf8b803a5f72352a760db62aa5ef2d6737

GliaWidgets/ViewModel/Call/CallViewModel.swift
kuldar-daniel: Added tests in 80bbe14e21ed66104d3535270a2bbaadb4c11327

GliaWidgets/Component/Bubble/BubbleView.swift
kuldar-daniel: Added

GliaWidgets/Component/OnHoldOverlay/OnHoldOverlayStyle.swift
igorkravchenko: @dukhovnyi @kuldar-daniel I think because most of the component styles are declared globally in the module, it makes sense to do the same for this style.

igorkravchenko (approved):
Good job 👍

GliaWidgets/Component/OnHoldOverlay/OnHoldOverlayStyle.swift
dukhovnyi: ok, let's do this way

GliaWidgets/Component/OnHoldOverlay/OnHoldOverlayVisualEffectView.swift
dukhovnyi: but in case of a memory leak `deinit` method will not execute or do I miss something?

kuldar-daniel: ah, you are correct. I guess I just saw online that it was common practice without really questioning it

dukhovnyi (approved):
Approve in advance, but this [comment](https://github.com/salemove/ios-sdk-widgets/pull/248#discussion_r830530796) has no answer.
Changes in this PR raised Swiftlint issue, so please fix it before merge.
One more suggestion, I'd like to propose to squash all commit in one to have a more flat history.

PR #250, opened 2022-03-16
Survey feature implementation

dukhovnyi:
MOB-1169

| Big state | Small state |
| :-: | :-: |
|  | |


GliaWidgets/Lib/View/View.swift
```diff
@@ -14,10 +14,26 @@ class View: UIView {
...
+    override func updateConstraints() {
+        if needsDefineLayout {
+            needsDefineLayout = false
+            defineLayout()
+        }
+        super.updateConstraints()
+    }
+
+    func setup() {}
+    func defineLayout() {}
```
kuldar-daniel, line 34: I think previously we have named the function just "layout", perhaps it would be better to keep it consistent with other views?

GliaWidgets/Layout/NSLayoutConstraint.Extensions.swift
```diff
@@ -0,0 +1,41 @@
+import Foundation
```
kuldar-daniel, line 1: PureLayout is also available in this project. If the long term goal is to remove the dependency it is fine, but in Cortex we used no dependency-approach also: https://github.com/salemove/ios-cortex-demo/tree/master/Cortex%20Financial/Utility/Constrainable

GliaWidgets/Layout/Makeable.swift
```diff
@@ -0,0 +1,22 @@
+import UIKit
+
+protocol Makeable {}
```
kuldar-daniel, line 3: That is pretty neat

GliaWidgets/Lib/View/View.swift
```diff
@@ -14,10 +14,26 @@ class View: UIView {
...
         super.init(frame: .zero)
+        setup()
     }
 
     @available(*, unavailable)
     required init?(coder: NSCoder) {
         fatalError("init(coder:) has not been implemented")
     }
+
+    override func updateConstraints() {
```
kuldar-daniel, line 25: Why not call defineLayout after setup in initialiser and get rid of the needsDefineLayout variable?

GliaWidgets/ViewController/Survey/Components/Survey.ValidationErrorView.swift
```diff
@@ -0,0 +1,35 @@
...
+        let validationMessage = UILabel().make {
+            $0.numberOfLines = 0
+            $0.textColor = .red
+            $0.font = .boldSystemFont(ofSize: 12)
+            $0.text = "Please provide an answer."
+        }
+        let validationImage = UIImageView().make {
+            $0.image = Asset.surveyValidationError.image
+        }
+        lazy var validationMessageHStack = UIStackView.make(.horizontal, spacing: 8)(
```
kuldar-daniel, line 14: Can't the Makeable protocol be used for stackviews?

GliaWidgets/ViewController/Survey/SurveyViewController.View.swift
```diff
@@ -0,0 +1,116 @@
...
+        // MARK: - Button container
+
+        let buttonContainer = UIView().makeView {
+            $0.backgroundColor = .white
+            $0.layer.masksToBounds = false
+            $0.layer.shadowColor = UIColor.black.cgColor
+            $0.layer.shadowOpacity = 0.15
+            $0.layer.shadowOffset = .zero
+            $0.layer.shadowRadius = 8
+
```
kuldar-daniel, line 30: unneeded empty line

GliaWidgets/ViewController/Survey/SurveyViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+        var props: Props {
+            didSet { render() }
+        }
+
+        required init?(coder: NSCoder) { preconditionFailure() }
+
+        init(props: Props = .init()) {
+            self.props = props
+            super.init(nibName: nil, bundle: nil)
+            if #available(iOS 13.0, *) {
```
kuldar-daniel, line 34: What if the survey only has 1 question, the presentation style will still take a whole screen of space?
I kind of imagined it being more like alert, with custom configurable dimmed background color over the view, and then content being subview on that survey view, only taking up as much space as needed, but not more than screen height * 0.9 or smth. Then disabling the ios 13 modal presentation style alltogether.

GliaWidgets/ViewController/Survey/SurveyViewController.swift
```diff
@@ -0,0 +1,88 @@
...
+                self.submit = submit
+                self.cancell = cancel
+            }
+        }
+
+        var props: Props {
+            didSet { render() }
+        }
+
+        required init?(coder: NSCoder) { preconditionFailure() }
```
kuldar-daniel, line 29: we usually mark the function `@available(*, unavailable)` so it could not be called and would throw compiler error if used

kuldar-daniel (approved):

GliaWidgets/ViewController/Survey/SurveyViewController.View.swift
```diff
@@ -0,0 +1,116 @@
...
+            constraints += contentContainerStackView.widthAnchor.constraint(
+                equalTo: scrollView.safeAreaLayoutGuide.widthAnchor,
+                constant: -2 * contentPadding
+            )
+            constraints += contentContainerStackView.bottomAnchor.constraint(
+                equalTo: scrollView.bottomAnchor,
+                constant: -contentPadding
+            )
+
+            constraints += buttonContainer.bottomAnchor.constraint(equalTo: bottomAnchor)
```
kuldar-daniel, line 94: This doesn't respect safe area

kuldar-daniel (changes requested):

GliaWidgets/Glia.swift
```diff
@@ -321,3 +323,140 @@ public class Glia {
...
     }
 }
+
+extension Glia {
+    // swiftlint:disable function_body_length
+        public func presentSurvey() {
+
+            let decoder = JSONDecoder.gliaApi
+
+            let utf8Data = """
```
igorkravchenko, line 333: This is temporary example of survey specific JSON and it will be removed after proper integration, right?

GliaWidgets/Layout/Makeable.swift
```diff
@@ -0,0 +1,22 @@
...
+
+extension Makeable where Self: UIView {
```
igorkravchenko, line 5: Having specific DSL for creating views while being extremely helpful, it makes support harder for those who are not familiar with this DSL, especially when it's not documented/tested. Is to possible omit usage of Makable? Would it make declaration of UIView more verbose?

GliaWidgets/Layout/NSLayoutConstraint.Extensions.swift
```diff
@@ -0,0 +1,41 @@
...
+        // swiftlint:enable operator_whitespace
+        destination.append(newConstraint)
+    }
+
+    func constraints(with id: NSLayoutConstraint.Identifier) -> [NSLayoutConstraint] {
+        filter { $0.identifier == id.rawValue }
+    }
+}
+
+extension NSLayoutConstraint {
```
igorkravchenko, line 27: Let's document this extension and it's types/methods.

GliaWidgets/Layout/NSLayoutConstraint.Extensions.swift
```diff
@@ -0,0 +1,41 @@
...
+    func activate() {
+        NSLayoutConstraint.activate(self)
+    }
+
+    func deactivate() {
+        NSLayoutConstraint.deactivate(self)
+    }
+
+    // swiftlint:disable operator_whitespace
+    static func +=(
```
igorkravchenko, line 14: I think it may by sufficient to only use
```
NSLayoutConstraint.activate(
    [
        constarint1,
        constarint2,
        ..
    ]
)
```
and avoid using custom extension of Array. What do you think @dukhovnyi @kuldar-daniel ?

GliaWidgets/ViewController/Survey/Components/Survey.ButtonView.swift
```diff
@@ -0,0 +1,75 @@
...
+            constraints += value.leadingAnchor.constraint(equalTo: leadingAnchor, constant: 1)
+            constraints += value.trailingAnchor.constraint(equalTo: trailingAnchor, constant: 1)
+            constraints += value.bottomAnchor.constraint(equalTo: bottomAnchor, constant: 1)
+        }
+
+        func render() {
+            value.text = props.title
+            switch props.state {
+            case .active:
+                value.layer.borderColor = UIColor.blue.cgColor
```
igorkravchenko, line 55: Colors are to be provided for themes, do I get it right?

GliaWidgets/ViewController/Survey/Components/Survey.ValidationErrorView.swift
```diff
@@ -0,0 +1,35 @@
+import UIKit
+
+extension Survey {
+    final class ValidationErrorView: View {
+        let validationMessage = UILabel().make {
+            $0.numberOfLines = 0
+            $0.textColor = .red
+            $0.font = .boldSystemFont(ofSize: 12)
+            $0.text = "Please provide an answer."
```
igorkravchenko, line 9: I suppose there will be a localization part. So in order to avoid missing localizations let's add some kind of `#warning` at least to indicated that and handle it in future.

GliaWidgets/ViewController/Survey/Survey.ViewController.Props.swift
```diff
@@ -0,0 +1,243 @@
...
+            optionProps.selected = option
+            optionProps.showValidationError = false
+            optionProps.answerContainer = .boolean(option.value)
+            setQuestion(sdkQuestion.id.rawValue, optionProps)
+            var newViewControllerProps = props()
+            newViewControllerProps.questionsProps = getQuestions()
+            updateProps(newViewControllerProps)
+        }
+        booleanProps.options = [
+            .init(name: "Yes", value: true, select: handleBooleanOptionSelection),
```
igorkravchenko, line 173: These will need to be localized as well.

GliaWidgets/Lib/View/View.swift
dukhovnyi: the main reason why I named it `defineLayout` is because this method contains mechanisms for defining the layout, so it calls once for defining layout but not for layouting in general. 
Does it make sense? or what do you think about naming @kuldar-daniel  @igorkravchenko ?

GliaWidgets/Layout/NSLayoutConstraint.Extensions.swift
dukhovnyi: yes, we would like to have as less 3rd party as possible, but in sense of PureLayout I have a few ideas how to manage layout better, one of them is that PureLayout activates each constraint separately, which is not good in sense of performance. We have a few extensions on Core SDK side, but since we would like to avoid custom mechanisms/frameworks, so I decided to don't use anything additionally for now and use standard approach for managing constraints :) but it is discussable.

GliaWidgets/Lib/View/View.swift
dukhovnyi: Because of creation constraints, it is part of establishing UI and it is not required to do it on construction/initialization step. Ex. you can just create a view and keep it in memory for a long time.

GliaWidgets/ViewController/Survey/SurveyViewController.View.swift
dukhovnyi: fixed f40675a3bae59f712d7dc991428ffce64843d838

GliaWidgets/ViewController/Survey/Components/Survey.ValidationErrorView.swift
dukhovnyi: in general can be, but in case of Makeable protocol you will need to define manually for each UIStackView this code
```
$0.translatesAutoresizingMaskIntoConstraints = false
$0.axis = axis
$0.spacing = spacing
$0.distribution = distribution
$0.alignment = alignment

... // other customization
```
to avoid this situation, I just create a separate static function :) just useful

GliaWidgets/ViewController/Survey/SurveyViewController.swift
dukhovnyi: fixed - 4c5f7f55d7a12723db9c4721f36e842cd93ad69f

GliaWidgets/ViewController/Survey/SurveyViewController.View.swift
dukhovnyi: good catch 5c5774c9127725a4a9612d39b8e585b3a081d9bc

GliaWidgets/Glia.swift
dukhovnyi: yes, it is just to present it quickly and check that this component works well.

GliaWidgets/Layout/Makeable.swift
dukhovnyi: I've added docs here - 43d3ed6b307908ef0f23ec2e657961394f864243
I just wanted to highlight that this approach is more about creation and setting up objects in one place and avoiding spearing initialization and setting up in different places of application/file.

GliaWidgets/Layout/NSLayoutConstraint.Extensions.swift
dukhovnyi: Added a few docs - d909ae32942ab85c0fe7851a61c36a533f017071
but perhaps I need to add just a comment with a description of the use-case for this function. What do you think, does it make sense?

GliaWidgets/Layout/NSLayoutConstraint.Extensions.swift
dukhovnyi: I think this approach is a little bit better, just examples

```
var constraints = [NSLayoutConstraint](); defer { constraints.activate() }
constraints += buttonContainer.bottomAnchor.constraint(equalTo: bottomAnchor)
constraints += buttonContainer.leadingAnchor.constraint(equalTo: leadingAnchor)
constraints += buttonContainer.trailingAnchor.constraint(equalTo: trailingAnchor)
```

```
let bottomButtonContainer = buttonContainer.bottomAnchor.constraint(equalTo: bottomAnchor)
let leadingButtonContainer = buttonContainer.leadingAnchor.constraint(equalTo: leadingAnchor)
let trailingAnchorContainer = buttonContainer.trailingAnchor.constraint(equalTo: trailingAnchor)
```

GliaWidgets/ViewController/Survey/Components/Survey.ButtonView.swift
dukhovnyi: yes, I'll move out all of them into survey theme.

GliaWidgets/ViewController/Survey/Components/Survey.ValidationErrorView.swift
dukhovnyi: Yes, I'll move this string in the next PR together with theme stuff.

GliaWidgets/ViewController/Survey/Survey.ViewController.Props.swift
dukhovnyi: Yes, I'll move this string in the next PR together with theme stuff.

GliaWidgets/Lib/View/View.swift
igorkravchenko: I think `layout` method is common among all views in SDK. It serves the same purpose as `defineLayout` - defines layout. I think `defineLayout` is more meaningful name, but I think it makes sense to use `layout` for consistency as @kuldar-daniel suggested.

GliaWidgets/ViewController/Survey/SurveyViewController.swift
```diff
@@ -0,0 +1,127 @@
...
+        // MARK: - Private
+
+        // swiftlint:disable force_cast
+        private var contentView: ContentView { view as! ContentView }
+        // swiftlint:enable force_cast
+    }
+}
+
+extension Survey.ViewController {
+    func subscribeToNotification(_ notification: NSNotification.Name, selector: Selector) {
```
kuldar-daniel, line 93: I suggest using KeyboardObserver here

kuldar-daniel (approved):
Looks good otherwise

GliaWidgets/ViewController/Survey/Components/Survey.ValidationErrorView.swift
igorkravchenko: I think it makes sense, but the issue here that `makeView` name does not _reflect_ that `translatesAutoresizingMaskIntoConstraints = true`, basically hides it. In my opinion it is better to be explicit about what method does under the hood. This can be addresses by having dedicated functions that do one thing and then composed into single one that will be applied to the view. Here's some good [examples](https://github.com/pointfreeco/episode-code-samples/blob/main/0003-styling-with-functions/Styling%20with%20Functions.playground/Pages/Complete.xcplaygroundpage/Contents.swift).
One note: since we need to release this feature in the near future we can leave it as is for now and then rework it.

igorkravchenko (approved):
Good job 👍

PR #252, opened 2022-03-30
Survey feature implementation

dukhovnyi:
MOB-1169

### Customization possibilities 
| Default | Dark mode example |
| :-: | :-: |
|  | |

### DIFF
--- a/GliaWidgets/UIColor+Extensions.swift
+++ b/GliaWidgets/UIColor+Extensions.swift
@@ -10,6 +10,19 @@ extension UIColor {
         )
     }
 
+    var hex: String {
+        var r: CGFloat = 0
+        var g: CGFloat = 0
+        var b: CGFloat = 0
+        var a: CGFloat = 0
+
+        getRed(&r, green: &g, blue: &b, alpha: &a)
+
+        let rgb: Int = (Int)(r * 255) << 16 | (Int)(g * 255) << 8 | (Int)(b * 255) << 0
+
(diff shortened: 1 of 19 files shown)


kuldar-daniel (approved):

GliaWidgets/UIColor+Extensions.swift
```diff
@@ -10,6 +10,19 @@ extension UIColor {
...
 
+    var hex: String {
+        var r: CGFloat = 0
+        var g: CGFloat = 0
+        var b: CGFloat = 0
+        var a: CGFloat = 0
+
+        getRed(&r, green: &g, blue: &b, alpha: &a)
+
+        let rgb: Int = (Int)(r * 255) << 16 | (Int)(g * 255) << 8 | (Int)(b * 255) << 0
```
igorkravchenko, line 21: I think in this case it is better to use `UInt` for representation of color hex since `UInt.max` provides twice wider range than `Int.max` on positive side,  and there's no need for negative numbers.

GliaWidgets/ViewController/Survey/SurveyViewController.swift
```diff
@@ -110,15 +117,15 @@ extension Survey {
 }
 
 extension Survey.QuestionPropsProtocol {
```
igorkravchenko, line 119: I think it would be safer to use enum instead of protocol here. But I suppose it can be addressed later at some point.

igorkravchenko (approved):

PR #253, opened 2022-03-31
Add localization for accessibility labels

igorkravchenko:
Add localized strings for VoiceOver.


kuldar-daniel:
@dukhovnyi @igorkravchenko the issue with the approach is that it will not be backwards compatible. Meaning, if existing integrators have defined their own styles for specific views, and now update to where accessibility is required argument, they will face errors with the new sdk

igorkravchenko:
> @dukhovnyi @igorkravchenko the issue with the approach is that it will not be backwards compatible. Meaning, if existing integrators have defined their own styles for specific views, and now update to where accessibility is required argument, they will face errors with the new sdk

@kuldar-daniel I think this is a good point. Do you have ideas on how this can be done differently to preserve backward compatibility and still have ability to localize accessibility labels and hints?

kuldar-daniel:
@igorkravchenko I think we should provide default values in the initialisers

igorkravchenko:
> @igorkravchenko I think we should provide default values in the initialisers

@kuldar-daniel After discussing this point with @dukhovnyi, it seems like it would be necessary to introduce breaking changes. Integrators will need to provide `Accessibility` for every style that requires it. However we added `Accessibility.unsupported` constant to allow integrators to bypass providing concrete strings, using the constant. This will make integrator aware that accessibility properties can be added, but they also can be omitted. But it is on integrator to make the choice.

GliaWidgets/Component/Bubble/BubbleStyle.Accessibility.swift
```diff
@@ -0,0 +1,24 @@
+extension BubbleStyle {
+    /// Accessibility properties for BubbleStyle.
+    public struct Accessibility: Equatable {
```
kuldar-daniel, line 3: Why do they need to conform to Equatable protocol?

kuldar-daniel, line 3: Why do they need to conform to Equatable protocol?

kuldar-daniel (approved):

GliaWidgets/Component/Bubble/BubbleStyle.Accessibility.swift
igorkravchenko: I plan to add unit tests that make sure that every style that has `Accessibility` can have `Accessibility.unsupported` passed to it. So basically for XCTAssertEqual.

igorkravchenko: [For XCTAssertEqual](https://github.com/salemove/ios-sdk-widgets/pull/253#discussion_r845776310).

dukhovnyi (approved):

PR #259, opened 2022-04-07
Fix Build Phases scripts for Apple Silicon Macs

kuldar-daniel:
As per SwiftLint's documentation:

`
That is because Homebrew on Apple Silicon installs the binaries into the /opt/homebrew/bin folder by default. To instruct Xcode where to find SwiftLint, you can either add /opt/homebrew/bin to the PATH environment variable in your build phase
`

This PR adds the `export PATH="$PATH:/opt/homebrew/bin"` before SwiftGen, SwiftLint, SwiftFormat run scripts so that they would run on M1 Macs.

### DIFF
--- a/GliaWidgets.xcodeproj/project.pbxproj
+++ b/GliaWidgets.xcodeproj/project.pbxproj
@@ -2057,7 +2057,7 @@
 			);
 			runOnlyForDeploymentPostprocessing = 0;
 			shellPath = /bin/sh;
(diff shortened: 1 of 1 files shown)


dukhovnyi (commented):
I think you need to edit your `~/.zshrc` or `~/.bashrc` and add at the end of file - `export PATH=/opt/homebrew/bin:$PATH`.
I believe we don't need to export binary into path from Build Phases.

PR #262, opened 2022-04-12
Fix restoration of engagement after app restart

igorkravchenko:
Provide fix by calling `Interactor.withConfiguration` which will restore ongoing engagement if there is one.
Note that I baaed this PR on `master`, but it seems to be mergable as is to `dev`.

MOB-1220


dukhovnyi:
Correct PR is here > https://github.com/salemove/ios-sdk-widgets/pull/263

PR #263, opened 2022-04-12
Fix restoration of engagement after app restart

igorkravchenko:
Provide fix by calling 'Interactor.withConfiguration' which will restore ongoing engagement if there is one.

MOB-1220


dukhovnyi (approved):
Looks good 👍

kuldar-daniel:
Should we add the same logic to Call ViewModel as well?

igorkravchenko:
> Should we add the same logic to Call ViewModel as well?
> 
> EDIT: ahh, since chat and call are created same time anyway, it doesn't matter?

Yeah, it doesn't seem to change anything having the same configuration setup call in `CallViewModel`, at least in terms of fixing current issue.

kuldar-daniel (approved):

PR #264, opened 2022-04-13
Implement engagement transfer in Interactor

kuldar-daniel:
This PR implements engagement transfer callback in Interactor.
Engagement transfer callback changes the state to being engaged with a new first available operator and notifies all listening parties of the change.

MUIC-244

### DIFF
--- a/GliaWidgets/CoreSDKClient/CoreSDKClient.Mock.swift
+++ b/GliaWidgets/CoreSDKClient/CoreSDKClient.Mock.swift
@@ -1,5 +1,6 @@
 #if DEBUG
 import Foundation
+import SalemoveSDK
 
 extension CoreSdkClient {
     static let mock = Self(
@@ -22,6 +23,22 @@ extension CoreSdkClient {
         fetchFile: { _, _, _ in },
         getCurrentEngagement: { return nil }
     )
+}
+
+extension CoreSdkClient.Operator {
+    static func mock(
+        id: String = "mockId",
+        name: String = "Mock Operator",
+        picture: SalemoveSDK.OperatorPicture? = nil,
+        availableMedia: [CoreSdkClient.MediaType]? = [.text, .audio, .video]
+    ) -> CoreSdkClient.Operator {
+        CoreSdkClient.Operator(
+            id: id,
+            name: name,
+            picture: picture,
+            availableMedia: availableMedia
+        )
+    }
 }
 
 extension CoreSdkClient.PushNotifications {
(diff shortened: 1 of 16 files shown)


kuldar-daniel:
If you guys have any idea how to write unit tests for this I would be thankful. It seems like Operator cannot be mocked due to init protection level, so I didn't find any easy to write unit test that state does change when engagement transfer callback gets called

igorkravchenko:
> If you guys have any idea how to write unit tests for this I would be thankful. It seems like Operator cannot be mocked due to init protection level, so I didn't find any easy to write unit test that state does change when engagement transfer callback gets called

Let's wait until `Operator` has public `init`. That will enable mocking for it.

dukhovnyi (commented):
`dev` branch has updated Core SDK dependency, so now Operator constructor is available to  use.

kuldar-daniel:
Added tests

GliaWidgets/Interactor/Interactor.swift
```diff
@@ -92,8 +93,13 @@ class Interactor {
             .filter { $0.0 != nil }
             .forEach {
                 let handler = $0.1
-                DispatchQueue.main.async {
+
```
igorkravchenko, line 96: Let's inject DispatchQueue.main.async. We have this under control I suppose via GCD. Also this check can go to GCD.mainQueue live version.

GliaWidgets/CoreSDKClient/CoreSDKClient.Mock.swift
```diff
@@ -24,6 +24,15 @@ extension CoreSdkClient {
     )
 }
 
+extension CoreSdkClient.Operator {
+    static let mock = CoreSdkClient.Operator(
```
igorkravchenko, line 28: I think it will be more beneficial to make it a function with these default values as arguments. This will allow this mock to be reusable between the tests. Does it make sense?

igorkravchenko (commented):
Overall looks solid. I left a few suggestions.

dukhovnyi (approved):
approved in advance

GliaWidgets/Interactor/Interactor.swift
kuldar-daniel: fixed in 64bfe8bfa8beb92d3bce127f30d4e957b2aab486

GliaWidgets/CoreSDKClient/CoreSDKClient.Mock.swift
kuldar-daniel: fixed in 64bfe8bfa8beb92d3bce127f30d4e957b2aab486

igorkravchenko (approved):
Looks good 👍

PR #265, opened 2022-04-13
Glia iOS Widgets SDK to 0.7.1

dukhovnyi:


kuldar-daniel:
Do we not need to update Package.swift also?

dukhovnyi:
@kuldar-daniel  good catch, thank you. Updated Package.swift.

kuldar-daniel (approved):

PR #266, opened 2022-04-14
Fixed ChatView backgroundColor customization

EgorovEI:
Fixed calling super setup method in ChatView.
Set clear tableView backgroundColor.

MOB-1209

### DIFF
--- a/GliaWidgets/View/Call/CallView.swift
+++ b/GliaWidgets/View/Call/CallView.swift
@@ -81,6 +81,54 @@ class CallView: EngagementView {
         adjustLocalVideoFrameAfterLayout()
     }
 
+    override func setup() {
+        topStackView.axis = .vertical
+        topStackView.spacing = 8
+        topStackView.addArrangedSubviews([operatorNameLabel, durationLabel])
+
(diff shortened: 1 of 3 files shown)


GliaWidgets/View/Call/CallView.swift
```diff
@@ -81,6 +81,54 @@ class CallView: EngagementView {
         adjustLocalVideoFrameAfterLayout()
     }
 
+    override func setup() {
```
kuldar-daniel, line 84: super.setup() missing here. Is this intentional?

dukhovnyi (approved):
Looks good 👍

GliaWidgets/View/Call/CallView.swift
EgorovEI: It's ok. As described in task "Right now, almost everything is configurable, but the background for the middle section of the chat (the white portion) is not exposed", we need to fix only backgroundColor in chat

igorkravchenko (approved):
Looks good to me 👍

kuldar-daniel (approved):

PR #278, opened 2022-05-03
Survey feature implementation

dukhovnyi:
MOB-1169


kuldar-daniel:
I see that Strings are not localisable and customisable, is this expected?

GliaWidgets/Theme/Survey/Theme+Survey.swift
```diff
@@ -0,0 +1,112 @@
+import Foundation
+import UIKit
+
+public extension Theme {
+    /// Text style.
+    struct Text {
```
igorkravchenko, line 6: I suppose you will need to provide public initializers for this and similar structs, since automatically synthesized will have internal access modifier by default, thus unavailable for integrator.

GliaWidgets/Theme/Survey/Theme+Survey.swift
```diff
@@ -0,0 +1,112 @@
+import Foundation
+import UIKit
+
+public extension Theme {
```
igorkravchenko, line 4: I think it is better to be more explicit with access modifiers by avoiding adding it to extension, and add it to declaration of a type instead:
```
 extension Theme {
    ...

    /// Text style.
    public struct Text {
        ...
```

GliaWidgets/Theme/Survey/Theme+Survey.swift
```diff
@@ -0,0 +1,112 @@
...
+    /// Text style.
+    struct Text {
+        /// Foreground hex color.
+        public var color: String
+        /// Font size.
+        public var fontSize: CGFloat
+        /// Font weight.
```
igorkravchenko, line 11: Since we are in the process of making fonts accessible (dynamic type) we need to consider alternative way of providing font style instead or in addition to specifying concrete size, thickness.

igorkravchenko (commented):
Looks solid. I left few comment regarding fonts, access modifiers and initializers.

GliaWidgets/Theme/Survey/Theme+Survey.swift
dukhovnyi: I thought that this is not needed in sense of a big initializer and integrators will change default values instead of implementing the full object separately, but anyway it makes sense to provide an initializer.
fixed - 2720cc15d186e3ab1ba91d1d759f26515e32142b

GliaWidgets/Theme/Survey/Theme+Survey.swift
dukhovnyi: fixed 2720cc15d186e3ab1ba91d1d759f26515e32142b

GliaWidgets/Theme/Survey/Theme+Survey.swift
dukhovnyi: I made to have consistency with existing code and we will introduce changes/improvements with ADA together.

dukhovnyi:
> I see that Strings are not localisable and customisable, is this expected?

good catch @kuldar-daniel , I just forgot about localization. 
Added localization here - 2720cc15d186e3ab1ba91d1d759f26515e32142b
About strings customization it is expected for now, because of thinking do we need to have it in Survey feature.

kuldar-daniel (approved):

igorkravchenko (approved):
Looks good 👍

PR #285, opened 2022-05-09
Visitor transfer in call views

kuldar-daniel:
Implement operator-to-operator engagement transfer for call views. The main idea is to reset Call to state where we wait for video/audio streams to come.

### DIFF
--- a/GliaWidgetsTests/Sources/CallViewModelTests.swift
+++ b/GliaWidgetsTests/Sources/CallViewModelTests.swift
@@ -183,11 +183,7 @@ class CallViewModelTests: XCTestCase {
     }
     
     func test_engagementTransferredReleasesRemoteAndLocalVideoAndShowsConnectingState() throws {
-        enum Calls {
-            case setLocalVideoNil
-            case setRemoteVideoNil
-            case showConnecting
-        }
+        enum Calls { case showConnecting }
         var calls: [Calls] = []
 
         var interactorEnv: Interactor.Environment = .failing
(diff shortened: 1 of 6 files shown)


GliaWidgetsTests/Sources/CallViewModelTests.swift
```diff
@@ -181,4 +181,57 @@ class CallViewModelTests: XCTestCase {
...
+            environment: .mock,
+            call: call,
+            unreadMessages: .init(with: 0),
+            startWith: .engagement(mediaType: .video)
+        )
+
+        viewModel.action = { action in
+            switch action {
+            case .setLocalVideo(let video):
+                if video == nil {
```
igorkravchenko, line 216: Why we need to check for `nil`?

kuldar-daniel: since we expect new streams from core, this just makes sure old ones are released from the view

igorkravchenko: Can we be sure it is the case by using `XCTAssertNill(video)` instead of `if video == nil`?

kuldar-daniel: Updated tests

kuldar-daniel:
> Sorry, for misunderstanding. I think it is great to cover cases like `case .setLocalVideo` and `case .setRemoteVideo`. My question was: can we use `XCTAssertNil(video)` in addition within those cases, instead of using `if video == nil`?
> 
> But also my point is that is not sufficient to use only `if` statement it test, without `else` clause if we can not use `XCTAssertNil(video)`.

Fixed in 4b534bbbe4d5eec8d4cc80f09606349e54a080ea

igorkravchenko (approved):
Looks good to me 👍

EgorovEI (approved):

PR #290, opened 2022-05-11
Fix text view appearance in dark mode

dukhovnyi:
MUIC-774
|Light mode|Dark mode|
|:-:|:-:|
|image|image|


kuldar-daniel:
Out of curiosity, why did you decide to go with Strings and hex for color instead of UIColor?

kuldar-daniel (approved):

dukhovnyi:
> Out of curiosity, why did you decide to go with Strings and hex for color instead of UIColor?

just to have a possibility to serialize color easily. In case integrators need to apply styles from some json/xml/etc. format colors will be serialized as a String and will not require any additional work. This is initial work for further improvements and reworking in this area.

PR #296, opened 2022-05-18
Remove SceneDelegate

dukhovnyi:
The project can't use SceneDelegate and AppDelegate at the same time. Removed SceneDelegate to make AppDelegate work and pass necessary events to SalemoveAppDelegate (example applicationDidBecomeActive).


kuldar-daniel:
Why can't it use both?

dukhovnyi:
> Why can't it use both?

Because if we have an "Application Scene Manifest" in  `Info.plist` and app delegate has a `configurationForConnectingSceneSession` method, the UIApplication doesn't send background and foreground lifecycle messages to our app delegate - `applicationDidBecomeActive` (our case), `applicationWillResignActive`, `applicationDidEnterBackground`, and `applicationWillEnterForeground`.

kuldar-daniel (approved):

EgorovEI (approved):

igorkravchenko (commented):
Not sure if removing it will help especially when we bump up deployment target to iOS 13. I think we need to ensure that SDK works in expected way with SceneDelegate too.
@dukhovnyi

dukhovnyi:
@igorkravchenko  From a short perspective, we should be sure that TestingApp works correctly, however, from a long perspective we should migrate to using SceneDelegate.

igorkravchenko (approved):

PR #297, opened 2022-05-19
Fixed sending messages during engagement establishing process

EgorovEI:
Fixed sending messages during engagement establishing process
Added tests

### DIFF
--- a/GliaWidgets/ViewModel/Chat/ChatViewModel.swift
+++ b/GliaWidgets/ViewModel/Chat/ChatViewModel.swift
@@ -340,7 +340,8 @@ extension ChatViewModel {
             files: files
         )
 
-        if interactor.isEngaged {
+        switch interactor.state {
+        case .engaged:
             let item = ChatItem(with: outgoingMessage)
             appendItem(item, to: messagesSection, animated: true)
             uploader.succeededUploads.forEach { action?(.removeUpload($0)) }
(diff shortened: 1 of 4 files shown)


igorkravchenko (approved):
Looks good 👍
We will update disabled unit tests after next release of CoreSDK.

dukhovnyi (approved):

GliaWidgetsTests/Mocks/CoreSDKClient.Operator.Mock.swift
```diff
@@ -0,0 +1,14 @@
+import Foundation
+import SalemoveSDK
+@testable import GliaWidgets
+
+extension CoreSdkClient.Operator {
+    static var mock: CoreSdkClient.Operator {
```
kuldar-daniel, line 6: I also made this in my visitor transfer PR. As noted there, it should be a function. Perhaps you should change it to function too so the merge conflicts wont be as bad?

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -368,26 +369,37 @@ extension ChatViewModel {
...
             }
-        } else {
-            let messageItem = ChatItem(with: outgoingMessage)
+
+        case .enqueued:
+            handle(pendingMessage: outgoingMessage)
+
+        case .enqueueing, .ended, .none:
+            handle(pendingMessage: outgoingMessage)
+            enqueue(mediaType: .text)
```
kuldar-daniel, line 378: Why do we need to call enqueue again here? what if it is a call?

EgorovEI: According to our previous and current logic we need to establish engagement if user sends message if it isn't already established. There I only replaced `if  interactor.isEngaged` with `switch` and added  separate handling for `enqueued` state.

GliaWidgetsTests/Mocks/CoreSDKClient.Operator.Mock.swift
EgorovEI: 19842e64d7c7c6f21a03c0c2a55f601583483950

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
kuldar-daniel: The previous logic was: until interactor tells us we are not engaged, add messages to pending queue. Once interactor tells us that we are engaged, send all the messages until queue is empty.
Calling enqueue(mediaType: .text) without checking whether the chat was actually supposed to start engagement or not (since Audio/Video call also starts chat) can lead to some odd behaviour I think.

kuldar-daniel: Looks like you cant go to chat anyway when call is in queue, although I don't see why we would not allow it. Might be some case we haven't covered yet but should.

PR #301, opened 2022-05-27
In call view during an audio/video new messages indicator does not show notification about new messages

kuldar-daniel:
MUIC-567


dukhovnyi (approved):
looks straightforward, do we need to add some tests for this?

igorkravchenko (approved):

PR #303, opened 2022-06-02
Add supporting default alert button styles

dukhovnyi:
Introducing applying alert button styles as a default for survey buttons.

### DIFF
--- a/GliaWidgets/ViewController/Survey/SurveyViewController.View.swift
+++ b/GliaWidgets/ViewController/Survey/SurveyViewController.View.swift
@@ -118,21 +118,9 @@ extension Survey {
                 buttonContainer.backgroundColor = .init(hex: hex)
             }
             scrollView.layer.cornerRadius = theme.survey.layer.cornerRadius
-            cancelButton.backgroundColor = .init(hex: theme.survey.cancellButton.background)
-            cancelButton.layer.cornerRadius = theme.survey.cancellButton.cornerRadius
-            cancelButton.titleLabel?.font = .systemFont(
-                ofSize: theme.survey.cancellButton.title.fontSize,
-                weight: .init(rawValue: theme.survey.cancellButton.title.fontWeight)
-            )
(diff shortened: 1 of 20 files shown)


kuldar-daniel (approved):

GliaWidgets/ViewController/Survey/SurveyViewController.View.swift
```diff
@@ -162,3 +150,25 @@ extension NSAttributedString {
         return mutableString
     }
 }
+
+extension UIButton {
+    func update(with style: Theme.Button) {
+        style.background.map { backgroundColor = UIColor(hex: $0) }
```
igorkravchenko, line 156: While it's convenient to use map like that, I think it goes against the purpose of map operation (transforming values).
Is it possible to use it like this?
```
backgroundColor = style.background.map(UIColor.init(hex:))
```
Does it make sense?

GliaWidgets/ViewController/Survey/SurveyViewController.View.swift
```diff
@@ -162,3 +150,25 @@ extension NSAttributedString {
...
+        contentEdgeInsets = .init(top: 6, left: 16, bottom: 6, right: 16)
+        clipsToBounds = true
+        layer.masksToBounds = false
+        layer.cornerRadius = style.cornerRadius
+
+        style.borderColor.map { layer.borderColor = UIColor(hex: $0).cgColor }
```
igorkravchenko, line 162: Same comment as above.

igorkravchenko (commented):
Looks good. Just a few map related questions.

GliaWidgets/ViewController/Survey/SurveyViewController.View.swift
dukhovnyi: image
I thought the same, but for some reason, Xcode throws error...

igorkravchenko (approved):

PR #305, opened 2022-06-06
Chat/Audio/Video - Tapping 15 digits long phone number crashes a call

kuldar-daniel:
This PR fixes issue, where tapping a 15 digit code would crash the app.
15 digit code was detected as a shipping number and therefore the linkTapped closure would be called.
Instead of this, we only track links and phone numbers so 15 digit long text won't be highlighted at all.
We also have an issue where in certain cases Messages app would open instead of phone dialer. 
Instead of letting TextView handle the URL managing we now compose and call the UIApplication to handle URL ourselves to tackle this issue.
MUIC-841

### DIFF
--- a/GliaWidgets/View/Chat/Message/Content/Text/ChatTextContentView.swift
+++ b/GliaWidgets/View/Chat/Message/Content/Text/ChatTextContentView.swift
@@ -104,10 +104,8 @@ extension ChatTextContentView: UITextViewDelegate {
         in characterRange: NSRange,
         interaction: UITextItemInteraction
     ) -> Bool {
-        if URL.scheme == "tel" {
-            openPhoneDialer(url: URL)
-        } else if URL.scheme == "mailto" {
-            openEmailClient(url: URL)
+        if URL.scheme == "tel" || URL.scheme == "mailto" {
+            openUrl(url: URL)
         } else {
             linkTapped?(URL)
         }
@@ -115,20 +113,10 @@ extension ChatTextContentView: UITextViewDelegate {
         return false
     }
 
-    private func openPhoneDialer(url: URL) {
-        let phone = url.absoluteString.replacingOccurrences(of: "tel:", with: "")
+    private func openUrl(url: URL) {
+        guard environment.uiApplication.canOpenURL(url) else { return }
 
-        if let callUrl = URL(string: "tel://\(phone)"), environment.uiApplication.canOpenURL(callUrl) {
-            environment.uiApplication.open(callUrl)
-        }
-    }
-
-    private func openEmailClient(url: URL) {
-        let email = url.absoluteString.replacingOccurrences(of: "mailto:", with: "")
-
-        if let emailUrl = URL(string: "mailto://\(email)"), environment.uiApplication.canOpenURL(emailUrl) {
-            environment.uiApplication.open(emailUrl)
-        }
+        environment.uiApplication.open(url)
     }
 }
 
--- a/GliaWidgets/View/Chat/Message/Content/Text/ChatTextContentView.swift
+++ b/GliaWidgets/View/Chat/Message/Content/Text/ChatTextContentView.swift
@@ -109,12 +109,21 @@ extension ChatTextContentView: UITextViewDelegate {
     }
 
     func handleUrl(url: URL) {
-        if url.scheme == "tel" || url.scheme == "mailto" {
-            guard environment.uiApplication.canOpenURL(url) else { return }
+        switch url.scheme?.lowercased() {
+        case "tel",
+            "mailto":
+            guard
+                environment.uiApplication.canOpenURL(url)
+            else { return }
 
             environment.uiApplication.open(url)
-        } else {
+
+        case "http",
+            "https":
             linkTapped?(url)
+
+        default:
+            return
         }
     }
 }
--- a/GliaWidgets/View/Chat/Message/Content/Text/ChatTextContentView.swift
+++ b/GliaWidgets/View/Chat/Message/Content/Text/ChatTextContentView.swift
@@ -58,7 +58,7 @@ class ChatTextContentView: UIView {
         textView.isEditable = false
         textView.textContainerInset = .zero
         textView.textContainer.lineFragmentPadding = 0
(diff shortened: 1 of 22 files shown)


GliaWidgets/View/Chat/Message/Content/Text/ChatTextContentView.swift
```diff
@@ -101,11 +101,30 @@ extension ChatTextContentView: UITextViewDelegate {
...
+        }
+
+        return false
+    }
+
+    private func openPhoneDialer(url: URL) {
+        let phone = url.absoluteString.replacingOccurrences(of: "tel:", with: "")
+
+        if let callUrl = URL(string: "tel://\(phone)"), UIApplication.shared.canOpenURL(callUrl) {
+            UIApplication.shared.open(callUrl)
```
igorkravchenko, line 119: Let's introduce Environment with `UIApplication` struct injected. It can be introduced under `UIKitBased` namespace.
Also let's make `openPhoneDialer` and `openEmailClient` internal and add unit tests for them using introduced `Environment` and `UIApplication` structs.

igorkravchenko (commented):
Looks good. Left a few comments.

GliaWidgets/View/Chat/Message/Content/Text/ChatTextContentView.swift
```diff
@@ -101,11 +101,30 @@ extension ChatTextContentView: UITextViewDelegate {
         in characterRange: NSRange,
         interaction: UITextItemInteraction
     ) -> Bool {
-        if URL.scheme == "tel" || URL.scheme == "mailto" {
-            return true
+        if URL.scheme == "tel" {
```
dukhovnyi, line 107: looks like ideal place to convert this to switch. what do you think, does it make sense?
```swift
        switch URL.scheme {
        case "tel":
            openPhoneDialer(url: URL)
        case "mailto":
            openEmailClient(url: URL)
        default:
           return false
        }
```

GliaWidgets/View/Chat/Message/Content/Text/ChatTextContentView.swift
```diff
@@ -101,12 +104,37 @@ extension ChatTextContentView: UITextViewDelegate {
...
+
+        return false
+    }
+
+    private func openPhoneDialer(url: URL) {
+        let phone = url.absoluteString.replacingOccurrences(of: "tel:", with: "")
+
+        if let callUrl = URL(string: "tel://\(phone)"), environment.uiApplication.canOpenURL(callUrl) {
+            environment.uiApplication.open(callUrl)
+        }
```
dukhovnyi, line 123: I may be wrong, but iOS should handle properly URL with `tel:` and `tel://`. Since we previously checked the URL scheme we need to check if OS can open URL and open it, or did I missed something?
```swift
private func openPhoneDialer(url: URL) {
    guard environment.uiApplication.canOpenURL(url) else { return }
    environment.uiApplication.open(url)
}
```

kuldar-daniel: I think you are correct, I was just following all the sources online I could find. They seemed to do the slashes thing.
But regarding to old docs it seems it should work OOO: https://developer.apple.com/library/archive/featuredarticles/iPhoneURLScheme_Reference/PhoneLinks/PhoneLinks.html#//apple_ref/doc/uid/TP40007899-CH6-SW1

GliaWidgets/View/Chat/Message/Content/Text/ChatTextContentView.swift
kuldar-daniel: reworked as suggested by yuri. Does it make sense to still write test for openUrl?

dukhovnyi (approved):

GliaWidgets/View/Chat/Message/Content/Text/ChatTextContentView.swift
igorkravchenko: Yes, to make sure that openUrl will receive only `tel:` and `mailto:` schemas.

kuldar-daniel: added tests in cad98bcb8c746c71e72b9cdae0523f4518a16f21

GliaWidgetsTests/Sources/ChatTextContentViewTests.swift
```diff
@@ -0,0 +1,68 @@
...
+        view = .mock(environment: environment)
+        view.linkTapped = {
+            calls.append(.linkTapped($0))
+        }
+
+        let linkUrl = URL(string: "https://mock.mock")!
+        view.handleUrl(url: linkUrl)
+
+        XCTAssertEqual(calls, [.linkTapped(linkUrl)])
+    }
```
igorkravchenko, line 67: What can also be done is adding test that makes sure that others unknown schemas will not work.

kuldar-daniel: added in 555c8eea864e0f2ae0dc0971e820c62f72c87122

igorkravchenko (approved):
Looks good 👍

GliaWidgets/View/Chat/Message/Content/Text/ChatTextContentView.swift
```diff
@@ -101,18 +104,56 @@ extension ChatTextContentView: UITextViewDelegate {
...
-        } else {
-            linkTapped?(URL)
-            return false
+        handleUrl(url: URL)
+        return false
+    }
+
+    func handleUrl(url: URL) {
+        switch url.scheme?.lowercased() {
+        case "tel",
```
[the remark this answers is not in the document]
kuldar-daniel: This piece of logic can be delegated into viewmodel, I agree. I can do it myself once I get to unit testing if it's okay? Or is it blocking something?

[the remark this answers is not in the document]
EgorovEI: No problem, it's not blocker. It will be good if you move it to ViewModel when you want.

PR #314, opened 2022-06-15
Added environment for FileUploadListView

EgorovEI:
Added preferredContentSizeCategory variable to UIKitBased.UIApplication
Added tests for FileUploadListView

MOB-1345

### DIFF
--- a/GliaWidgets/View/Chat/Entry/Upload/FileUploadListView.swift
+++ b/GliaWidgets/View/Chat/Entry/Upload/FileUploadListView.swift
@@ -4,6 +4,14 @@ import SwiftUI
 class FileUploadListView: UIView {
     var removeTapped: ((FileUpload) -> Void)?
 
+    var maxUnscrollableViews: Int {
+        if environment.uiApplication.preferredContentSizeCategory() <= .accessibilityMedium {
+            return 3
+        } else {
+            return 2
+        }
+    }
+
     private var uploadViews: [FileUploadView] {
         return stackView.arrangedSubviews.compactMap { $0 as? FileUploadView }
     }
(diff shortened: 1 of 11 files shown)


GliaWidgets/View/Chat/Entry/Upload/FileUploadListView.swift
```diff
@@ -4,23 +4,29 @@ import SwiftUI
 class FileUploadListView: UIView {
     var removeTapped: ((FileUpload) -> Void)?
 
+    var maxUnscrollableViews: Int {
+        if environment.uiApplication.preferredContentSizeCategory() <= .accessibilityMedium {
+            return 3
```
gersonnoboa, line 9: Not having the context about this specific project, these numbers don't say much when reading them. I don't know if putting them into variables would solve the issue, so maybe having a comment for `maxUnscrollableViews` that says why 3 or 2 according to one or other case would be good.

igorkravchenko: I support @gersonnoboa on this. I would recommend to add comments regarding the logic behind this computed property and the goal it achieves.

igorkravchenko (approved):
Looks good. I approve it in advance taking into account the comment from Gerson will be addressed.

GliaWidgets/View/Chat/Entry/Upload/FileUploadListView.swift
EgorovEI: 7138828f0ce6e4548422b7865f271f1b79d9aa10

gersonnoboa (approved):

PR #319, opened 2022-06-22
Removed app-token usage from TestingApp

EgorovEI:


igorkravchenko (approved):

dukhovnyi (commented):
Looks good, however, I believe we need to make unavailable app-token in Configuration also. My suggestions are
- change "deprecated" to "unavailable" for interfaces in [Configuration.Deprecated.swift](https://github.com/salemove/ios-sdk-widgets/blob/master/GliaWidgets/Configuration.Deprecated.swift). Gist is [here](https://gist.github.com/dukhovnyi/9672c189796a4962861ba74ef9774d14). Since this interfaces will not be available anymore, I think we could use "fatalError" inside of body if needed (for example in init methods). What do you think guys @igorkravchenko @gersonnoboa @EgorovEI about fatalError, does it make sense?
- rename "Configuration.Deprecated.swift" to "Configuration.Unavailable.swift"
- mark case "appToken" as unavailable also in [Configuration.swift](https://github.com/salemove/ios-sdk-widgets/blob/master/GliaWidgets/Configuration.swift#L61)
- remove from documentation `/// - 'SalemoveSDK.ConfigurationError.invalidAppToken'` since   Core SDK will not throw this kind of error also. [Example](https://github.com/salemove/ios-sdk-widgets/blob/master/GliaWidgets/Glia.swift#L91)

gersonnoboa:
@dukhovnyi as far as I know, there is a Glia policy that deprecated methods should work properly for at least one year before making them fully unavailable. So we might not be able to throw fatalError from the widgets yet.

gersonnoboa (approved):
Because of the fact that we might not be able to make methods unavailable yet, then I am going to approve this as it is.

EgorovEI:
The main point is can we mark interfaces unavailable. Some time ago Carlos said we might do this after at least 6 month. 
6 month are gone from deprecating. 

I agree with `fatalError()` usage with wrapping function for `unavailable` cases.

gersonnoboa:
As I have been told that it should be six months and not one year, then the unavailable + fatalError sounds fine for me as well.

EgorovEI:
@igorkravchenko @gersonnoboa @dukhovnyi please check changes discussed above e6df5344b8e23473ca14a2bd6ec71edaee5b9b3a

dukhovnyi:
```swift
func unavaialbe(function: StaticString = #function) -> Never {
    fatalError("This implementation for \(function) is not longer available.")
}
```
@igorkravchenko looks like a good idea, however, I have one question, does this message bring to the integrator any useful information, except the function is not available anymore? :) perhaps we could provide information to the integrator what method can replace the current one. if the function has no analogs then yes, this message is good enough.

dukhovnyi (approved):

gersonnoboa (approved):

PR #326, opened 2022-06-30
Add Interactor tests

kuldar-daniel:
MUIC-819


EgorovEI (approved):

GliaWidgets/Interactor/Interactor.swift
```diff
@@ -160,33 +161,6 @@ extension Interactor {
         }
     }
 
-    func request(
```
igorkravchenko, line 163: This method is no longer needed?

kuldar-daniel: It was never used. I think it might become useful again in the future when visitor can request media upgrade as well, but then it can be implemented again. Redundant for now.

dukhovnyi (approved):

PR #327, opened 2022-06-30
Add accessibility identifiers for using in AT environment

dukhovnyi:


gersonnoboa (commented):
I see that these IDs are a bit different to the ones in `Configuration.storyboard`. The format there seems quite useful because it's like `screen_name_viewtype`. Do you think that those could be useful here as well? Not only to maintain consistency but also because the format itself is useful.

dukhovnyi:
@gersonnoboa yep, agree that style should be consistent, we just need to decide what format is better and unified. Would you like to use the style from Configuration.storyboard? what about you guys @igorkravchenko @EgorovEI ?

igorkravchenko:
> @gersonnoboa yep, agree that style should be consistent, we just need to decide what format is better and unified. Would you like to use the style from Configuration.storyboard? what about you guys @igorkravchenko @EgorovEI ?

I don't have concrete preference on this, but if other views already use this style, I think it makes sense to keep it that way.

igorkravchenko (approved):

dukhovnyi:
@gersonnoboa  your suggestion makes sense, let's follow `screen_name_viewtype` naming for accessibility identifiers. Updated - fd646da068b6d50ec9ca7c126fab422b2a8287ba

gersonnoboa (approved):

gugalo:
Merging this PR as it is required by the https://github.com/salemove/acceptance_tests/pull/2646

gugalo:
!merge

PR #331, opened 2022-07-12
Fix Queue identifier doesn't store

dukhovnyi:


gersonnoboa (approved):
The changes are fine but the commit messages for each one could display more information. Both for the future and for someone who is out of context to understand what is being achieved.

EgorovEI (approved):

PR #339, opened 2022-07-21
Add accessibility label for audio upgrades

gersonnoboa:
In order to detect if an engagement has been upgraded to audio, new
accessibility identifiers have been added to the buttons in the audio
call view, and the alert that appears when the operator requests an
audio call from the visitor. For the audio call view, the
accessibility identifier is only set when the button is enabled.

MOB-1400

### DIFF
--- a/GliaWidgets/Component/CallButtonBar/Button/CallButton.swift
+++ b/GliaWidgets/Component/CallButtonBar/Button/CallButton.swift
@@ -132,6 +132,9 @@ class CallButton: UIView {
         )
         accessibilityValue = properties.value
         accessibilityLabel = properties.label
+
+        let buttonAccessibilityIdentifier = "audio_\(properties.value.lowercased())_button"
+        accessibilityIdentifier = isUserInteractionEnabled ? buttonAccessibilityIdentifier : ""
     }
 
     private func style(for state: State) -> CallButtonStyle.StateStyle {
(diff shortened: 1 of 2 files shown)


GliaWidgets/Component/CallButtonBar/Button/CallButton.swift
```diff
@@ -131,7 +131,9 @@ class CallButton: UIView {
             badgeItemCount: badgeView?.newItemCount
         )
         accessibilityValue = properties.value
-        accessibilityLabel = properties.label
+
+        let buttonAccessibilityLabel = "audio_\(properties.value.lowercased())_button"
+        accessibilityLabel = isUserInteractionEnabled ? buttonAccessibilityLabel : ""
```
EgorovEI, line 136: Can Appium find element by accessibilityLabel? 
Looks a bit strange you use `accessibilityLabel` instead of `accessibilityIdentifier`

dukhovnyi: I'm not sure, I think this is just typo :)

gersonnoboa: I definitely meant identifier, but the test worked with both... Fixed it now

EgorovEI (approved):

dukhovnyi (approved):

GliaWidgets/Component/CallButtonBar/Button/CallButton.swift
dukhovnyi: afaik, Appium when search item starts from identifier > label > if label or text item compares text.

igorkravchenko (approved):

gersonnoboa:
!merge

PR #340, opened 2022-07-25
Change accessibility identifier prefix to `media`

gersonnoboa:
Because the CallButton class is used in both audio and video, then it
makes more sense to have the accessibility identifier's prefix to be
`media` rather than `audio`.

MOB-1400


dukhovnyi (approved):

EgorovEI (approved):

gersonnoboa:
!merge

PR #346, opened 2022-07-26
Add unit tests for default parameters of deprecated Glia.start

igorkravchenko:
Accompanying unit test to make sure that passed parameters from deprecated `start` method are passed further.

MOB-1464

### DIFF
--- a/GliaWidgets/Coordinator/RootCoordinator.Environment.swift
+++ b/GliaWidgets/Coordinator/RootCoordinator.Environment.swift
@@ -1,11 +1,11 @@
-import Foundation
-
-extension EngagementViewModel {
+extension RootCoordinator {
     struct Environment {
         var chatStorage: Glia.Environment.ChatStorage
         var fetchFile: CoreSdkClient.FetchFile
         var sendSelectedOptionValue: CoreSdkClient.SendSelectedOptionValue
         var uploadFileToEngagement: CoreSdkClient.UploadFileToEngagement
+        var audioSession: Glia.Environment.AudioSession
+        var uuid: () -> UUID
         var fileManager: FoundationBased.FileManager
         var data: FoundationBased.Data
         var date: () -> Date
@@ -14,10 +14,10 @@ extension EngagementViewModel {
         var uiImage: UIKitBased.UIImage
         var createFileDownload: FileDownloader.CreateFileDownload
         var loadChatMessagesFromHistory: () -> Bool
+        var timerProviding: FoundationBased.Timer.Providing
         var fetchSiteConfigurations: CoreSdkClient.FetchSiteConfigurations
         var getCurrentEngagement: CoreSdkClient.GetCurrentEngagement
-        var timerProviding: FoundationBased.Timer.Providing
-        var uuid: () -> UUID
+        var submitSurveyAnswer: CoreSdkClient.SubmitSurveyAnswer
         var uiApplication: UIKitBased.UIApplication
     }
 }
--- a/GliaWidgets.xcodeproj/project.pbxproj
+++ b/GliaWidgets.xcodeproj/project.pbxproj
@@ -342,6 +342,17 @@
 		9AE9E4B327E0E60F00BFE239 /* CallViewModel.Mock.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9AE9E4B227E0E60F00BFE239 /* CallViewModel.Mock.swift */; };
 		9AE9E4B527E0EE2E00BFE239 /* CallViewControllerVoiceOverTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9AE9E4B427E0EE2E00BFE239 /* CallViewControllerVoiceOverTests.swift */; };
 		9AE9E4B727E1E30500BFE239 /* MockHelpers.swift in Sources */ = {isa = PBXBuildFile; fileRef = 9AE9E4B627E1E30500BFE239 /* MockHelpers.swift */; };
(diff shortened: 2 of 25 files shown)


dukhovnyi (approved):
looks very good 👍

GliaWidgets/Coordinator/RootCoordinator.Environment.swift
```diff
@@ -0,0 +1,23 @@
...
+    struct Environment {
+        var chatStorage: Glia.Environment.ChatStorage
+        var fetchFile: CoreSdkClient.FetchFile
+        var sendSelectedOptionValue: CoreSdkClient.SendSelectedOptionValue
+        var uploadFileToEngagement: CoreSdkClient.UploadFileToEngagement
+        var audioSession: Glia.Environment.AudioSession
+        var uuid: () -> UUID
+        var fileManager: FoundationBased.FileManager
+        var data: FoundationBased.Data
+        var date: () -> Date
```
gersonnoboa, line 11: Maybe this one as well. I guess it's the current date.

GliaWidgets/Coordinator/RootCoordinator.Environment.swift
```diff
@@ -0,0 +1,23 @@
+extension RootCoordinator {
+    struct Environment {
+        var chatStorage: Glia.Environment.ChatStorage
+        var fetchFile: CoreSdkClient.FetchFile
+        var sendSelectedOptionValue: CoreSdkClient.SendSelectedOptionValue
+        var uploadFileToEngagement: CoreSdkClient.UploadFileToEngagement
+        var audioSession: Glia.Environment.AudioSession
+        var uuid: () -> UUID
+        var fileManager: FoundationBased.FileManager
+        var data: FoundationBased.Data
```
gersonnoboa, line 10: This one also has a very non-descriptive name

GliaWidgets/Coordinator/RootCoordinator.Environment.swift
```diff
@@ -0,0 +1,23 @@
+extension RootCoordinator {
+    struct Environment {
+        var chatStorage: Glia.Environment.ChatStorage
+        var fetchFile: CoreSdkClient.FetchFile
+        var sendSelectedOptionValue: CoreSdkClient.SendSelectedOptionValue
+        var uploadFileToEngagement: CoreSdkClient.UploadFileToEngagement
+        var audioSession: Glia.Environment.AudioSession
+        var uuid: () -> UUID
```
gersonnoboa, line 8: And this one I guess is used for when one wants to generate a UUID, to ensure that you can test it? But the name is very undescriptive.

gersonnoboa (commented):
Don't really have enough context to comment much apart from what I identified as very undescriptive names in some names. But I don't know if it's possible, or advisable, to change them.

GliaWidgets/Coordinator/RootCoordinator.Environment.swift
igorkravchenko: I guess in order to clarify this I need to provide a bit of context about Environment(s) that are widely used in Widgets SDK. Basically Environment is a dependency container that solves the problem exchanging dependencies during testing. 
So regarding particular naming of things like `var date: () -> Date` and `var uuid: () -> UUID` is way of saying is that there is going to be used `Date()` and UUID() accordingly. In case of using some type that has several initializers that expect specific arguments, there will be used wrapper struct instead of simple closure, like `var  data: FoundationBased.Data` fro example. But the idea stays the same - for using 3rd party frameworks like CoreSDK and 1st party like `Foundation.FileManager` we will have closures and structs-wrappers that mimic actual types provided by frameworks. 
Child Environment that is going to use `Foundation.Date` for some very specific case will get this type injected from parent Environment that may not use it explicitly (or use it for some very different case), so there's no sense give to parent Environment property specific name, but very general one like `var date: () -> Date`. Does it make more sense now?

GliaWidgets/Coordinator/RootCoordinator.Environment.swift
igorkravchenko: I added explanation [here](https://github.com/salemove/ios-sdk-widgets/pull/346#discussion_r933312932).

GliaWidgets/Coordinator/RootCoordinator.Environment.swift
igorkravchenko: I added explanation https://github.com/salemove/ios-sdk-widgets/pull/346#discussion_r933312932.

GliaWidgets/Coordinator/RootCoordinator.Environment.swift
```diff
@@ -0,0 +1,23 @@
...
+        var fileManager: FoundationBased.FileManager
+        var data: FoundationBased.Data
+        var date: () -> Date
+        var gcd: GCD
+        var localFileThumbnailQueue: FoundationBased.OperationQueue
+        var uiImage: UIKitBased.UIImage
+        var createFileDownload: FileDownloader.CreateFileDownload
+        var fromHistory: () -> Bool
```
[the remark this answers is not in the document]
igorkravchenko: Fixed a70be0799a23b151a0cf090a8d95bce4d068b918.

GliaWidgets/Coordinator/RootCoordinator.Environment.swift
gersonnoboa: All of it makes sense :) but maybe if you add something like this in the code, then it would be perfect, and then the names can stay as is.

igorkravchenko: Added documentation regarding Environment(s) in the framework and naming conventions 8f8892caf7aa9b548a6619e7a253524f4a858137.

gersonnoboa (approved):

igorkravchenko:
!squash

PR #352, opened 2022-08-02
Removed applying frames to tested controllers

EgorovEI:
After removing SceneDelegate snapshot test were broken, but Testing app appearance hasn't been changed. 
It affected only snapshot tests. So after removing applying frames to tested controllers tests started to pass.

Maybe we will need to roll back this code when add SceneDelegate again in the future.

### DIFF
--- a/GliaWidgets/View/Call/CallView.swift
+++ b/GliaWidgets/View/Call/CallView.swift
@@ -126,6 +126,8 @@ class CallView: EngagementView {
     }
 
     override func setup() {
+        accessibilityIdentifier = "call_root_view"
+
         topStackView.axis = .vertical
         topStackView.spacing = 8
         topStackView.addArrangedSubviews([operatorNameLabel, secondLabel])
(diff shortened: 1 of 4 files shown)


GliaWidgets/View/Call/CallView.swift
```diff
@@ -126,6 +126,8 @@ class CallView: EngagementView {
     }
 
     override func setup() {
+        accessibilityIdentifier = "call_root_view"
```
EgorovEI, line 129: I probably forgot to merge this line :-)

dukhovnyi (commented):
is it the reason of broken snapshot tests?

EgorovEI:
The reason was applying frames

dukhovnyi (approved):

gersonnoboa (approved):
It would be good to add your description of this pull request to the commit description too, so that it's easier to know why this was done.

igorkravchenko (approved):

PR #362, opened 2022-08-24
Bump Core SDK version after release

gersonnoboa:
After the Core SDK is released, the Podfile.lock of the widgets should
be updated to reflect this. Thus, an update is triggered in Bitrise,
and then the `create_pr` lane is run in Fastlane in order to create
a PR with the new version.

MOB-1502

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -262,6 +262,16 @@ workflows:
     description: >-
       Workflow for creating builds of dev branch. Triggered on each push to dev
       branch, notifies over email + Slack (internal)
+  update_dependencies:
+    steps:
+    - activate-ssh-key@4: {}
+    - git-clone@6: {}
+    - cocoapods-install@2:
+        inputs:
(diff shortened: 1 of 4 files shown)


EgorovEI (approved):

scripts/create_pr.sh
```diff
@@ -0,0 +1,7 @@
+BRANCH_NAME=$1
```
dukhovnyi, line 1: I'd like to propose renaming this script to "commit_unstaged_changes" or something like this, what do you think, does it make sense?

dukhovnyi (approved):

scripts/create_pr.sh
gersonnoboa: Yep, that one is better, fixed: 0be5b37b57ae76c182a42f787ec920d9685e56e2

gersonnoboa:
!squash

PR #363, opened 2022-08-29
Set operator image to default on engagement termination

rasmustautsglia:
When operator ended the engagement, the chat bubble, on the client side, was set to generic silhouette placeholder.

This fix keeps the default operator image until the user has ended the engagement on their side.

MOB-1428

### DIFF
--- a/GliaWidgets/ViewModel/EngagementViewModel.swift
+++ b/GliaWidgets/ViewModel/EngagementViewModel.swift
@@ -135,13 +135,6 @@ class EngagementViewModel {
             engagementDelegate?(.finished)
 
         case .ended(let reason) where reason == .byOperator:
-
-            self.engagementDelegate?(
-                .engaged(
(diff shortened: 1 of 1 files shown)


dukhovnyi (approved):
Looks good 👉
Just a few things to improve in this PR
- I'd like to propose to name branch as ticket number, like `MOB-001`
- To easily keep on track of ticket progress in Jira we need to add to commit description ticket number. It can be done on last line in ticket description like this 
```
This is commit title
Here is some details about changes
MOB-001
```

dukhovnyi (commented):
will we show embedded into SDK `placeholder` after ending engagement or `site default image`?

gersonnoboa (approved):

gersonnoboa:
Could you rephrase your question @dukhovnyi ? So we can check that before merging

rasmustautsglia:
!merge

PR #367, opened 2022-09-06
Visitor Opaque Authentication

igorkravchenko:
Add visitor opaque authentication. This is accompanying addition on top of Core SDK, aimed to bring VOA (also known as *Direct ID Token Authentication*) to Widgets:

- Currently authentication is not allowed during ongoing engagement (this restriction may be lifted in future).
- In order to authenticate or deauthenticate SDK configuration must be accomplished.

MOB-1492


GliaWidgetsTests/ViewModel/Chat/Storage/AuthenticatedChatStorageTests.swift
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    func test_isNewMessageReturnsTrueForNotStoredMessage() {
+        let messageId = "messageId"
+        XCTAssertTrue(storage.isNewMessage(.mock(id: messageId)))
+    }
+
+    func test_isNewMessageReturnsFalseForStoredMessage() {
+        let messageId = "messageId"
+        let queueId = "queueID"
```
gersonnoboa, line 63: maybe messageId and queueId can be set as lets at the top? seems like they are the same for all

Podfile
```diff
@@ -23,4 +23,4 @@ end
 
 target 'SnapshotTests' do
   pod 'AccessibilitySnapshot', '0.5.0'
-end
\ No newline at end of file
+end
```
gersonnoboa, line 26: Nice

gersonnoboa (approved):

gersonnoboa:
However, a description in the commit would be useful.

Podfile
igorkravchenko: That is a part of 71208e1c8e0961100c29d77eb65e78dbabe24100 where I remove linked locally Core SDK source via Cocoapods (committed that by mistake).

dukhovnyi (approved):

dukhovnyi:
!squash

PR #374, opened 2022-09-20
Fixed tableView jumping on reloading response card

EgorovEI:
The table view was jumping on refreshRow calling because of inside the method technically the cell content is reloaded twice, by applying content property and calling tablleView.reloadRow method. This commit fixes it by deleting reloadRow call. Also contains inserting the rows inside performBatchUpdates, according Apple documentation.

### DIFF
--- a/GliaWidgets/View/Chat/ChatView.swift
+++ b/GliaWidgets/View/Chat/ChatView.swift
@@ -153,14 +153,19 @@ class ChatView: EngagementView {
     }
 
     func appendRows(_ count: Int, to section: Int, animated: Bool) {
-        guard let rows = numberOfRows?(section) else { return }
+        guard
+            let rows = numberOfRows?(section),
+            rows >= count
+        else { return }
 
         if animated {
-            let indexPaths = (rows - count ..< rows)
-                .map { IndexPath(row: $0, section: section) }
-            tableView.insertRows(at: indexPaths, with: .bottom)
+            tableView.performBatchUpdates {
+                let indexPaths = (rows - count ..< rows)
+                    .map { IndexPath(row: $0, section: section) }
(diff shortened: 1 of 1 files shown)


GliaWidgets/View/Chat/ChatView.swift
```diff
@@ -156,11 +156,13 @@ class ChatView: EngagementView {
         guard let rows = numberOfRows?(section) else { return }
```
dukhovnyi, line 156: perhaps it is overhead, however, I'd like to propose to add to `guard` one more statement
```swift
guard 
    let rows = numberOfRows?(section),
    rows > count
else { return }
```
to avoid `out of range` exception, just in case.

dukhovnyi (approved):

gersonnoboa (commented):
The commit (and the PR) doesn't have any information about this commit accomplishes its result. So I see the code but I can't really say if what you are trying to accomplish is done correctly in this code. Something like `The table view was jumping because [insert here whatever was happening with the animations]. This commit fixes it by inserting the rows inside performBatchUpdates, thus [insert here why that fixes it].`. This will make our reviews more meaningful and also will let future people understand the commit by reading the description.

GliaWidgets/View/Chat/ChatView.swift
EgorovEI: Makes sense, but with `rows >= count`, for cases if it's the first item

EgorovEI: d91e38d332b823cb0ee0785fa35eb9a4c5f7fc1c

gersonnoboa (approved):

rasmustautsglia (approved):

igorkravchenko (approved):

PR #382, opened 2022-09-23
Add alert customization from remote config

rasmustautsglia:
Add alert customization from remote configuration. SurveyButton has been renamed to ActionButton because of wider use case for such struct

MOB-1563

### DIFF
--- a/GliaWidgets/RemoteConfiguration/RemoteConfiguration.swift
+++ b/GliaWidgets/RemoteConfiguration/RemoteConfiguration.swift
@@ -4,6 +4,7 @@ public struct RemoteConfiguration: Codable {
     public let callScreen: Call?
     public let chatScreen: Chat?
     public let surveyScreen: Survey?
+    public let alertScreen: Alert?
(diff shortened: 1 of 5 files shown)


GliaWidgets/RemoteConfiguration/RemoteConfiguration.swift
```diff
@@ -33,6 +34,24 @@ public extension RemoteConfiguration {
         }
     }
 
+    struct Alert: Codable {
+        let titleFont: Font?
```
EgorovEI, line 38: I guess `titleFont` and `titleColor` can be replaced with `title: Text?`
And the same for `message`

rasmustautsglia: Fixed.

dukhovnyi (approved):

EgorovEI (approved):

rasmustautsglia:
!merge

PR #397, opened 2022-10-13
Fix restoring of engagement for enqueued flow

igorkravchenko:
Currently for authenticated visitor resuming of engagement does not happen. This PR attempts to solve this.

MOB-1492

### DIFF
--- a/GliaWidgets/ViewModel/Chat/ChatViewModel.swift
+++ b/GliaWidgets/ViewModel/Chat/ChatViewModel.swift
@@ -157,7 +157,11 @@ class ChatViewModel: EngagementViewModel, ViewModel {
             // messages from operator until message is sent from visitor.
             self?.interactor.withConfiguration { [weak self] in
                 guard let self = self else { return }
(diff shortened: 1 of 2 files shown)


GliaWidgets/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -157,7 +157,9 @@ class ChatViewModel: EngagementViewModel, ViewModel {
             // messages from operator until message is sent from visitor.
             self?.interactor.withConfiguration { [weak self] in
                 guard let self = self else { return }
-                if case .startEngagement = self.startAction, self.chatStorageState().isEmpty() {
+                if
```
gersonnoboa, line 160: Do you think that putting the condition in a `let` and then asking `if condition` would help make this code more readable? Although I don't know if you can do that with `case`. Right now it's not clear to me what scenario we are evaluating.

gersonnoboa (approved):

dukhovnyi:
!squash

dukhovnyi:
!merge

PR #398, opened 2022-10-14
Change version increment type from minor to patch

gersonnoboa:
Currently, the release process will update the project version to minor, but before we were doing it with patch, so now it's changed to be consistent with how we did things before.

MOB-1589


dukhovnyi (approved):

dukhovnyi:
!merge

PR #399, opened 2022-10-18
Add selected state to ButtonBar buttons

rasmustautsglia:
Currently only active and inactive states were present, leaving out selected state. This PR adds the possibility to customize selected state as well.

MOB-1597

### DIFF
--- a/GliaWidgets/Component/CallButtonBar/Button/CallButtonStyle.swift
+++ b/GliaWidgets/Component/CallButtonBar/Button/CallButtonStyle.swift
@@ -26,11 +26,14 @@ public struct CallButtonStyle {
         public var accessibility: Accessibility
     }
 
-    /// Style of active (i.e. toggled "on") state.
+    /// Style of active state.
     public var active: StateStyle
 
-    /// Style of inactive (i.e. toggled "off") state.
+    /// Style of inactive state.
     public var inactive: StateStyle
+
+    /// Style of selected state.
+    public var selected: StateStyle
 
     /// Accessibility related properties.
     public var accessibility: Accessibility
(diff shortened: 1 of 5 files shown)


dukhovnyi (approved):

GliaWidgets/Component/CallButtonBar/Button/CallButtonStyle.swift
```diff
@@ -160,4 +164,64 @@ public struct CallButtonStyle {
...
+
+    private mutating func applySelectedBarButtonConfiguration(button: RemoteConfiguration.BarButtonStates?) {
+        button?.selected?.background?.type.map { backgroundType in
+            switch backgroundType {
+            case .fill:
+                button?.selected?.background?.value.map {
+                    selected.backgroundColor = UIColor(hex: $0[0])
+                }
+            case .gradient:
+
```
gersonnoboa, line 176: NIT: I don't think that the comment needs a space above and below. Also it could be aligned to the indentation that the actual code is going to have rather than the indentation of the case statement. So, like

```
case .gradient:
    // The logic for gradient has not been implemented
    break
```

In case you decide to take the suggestion, then it would apply to all comments. Otherwise leave all as they are right now, as long as they are consistent with each other.

gersonnoboa (approved):

GliaWidgets/Component/CallButtonBar/Button/CallButtonStyle.swift
rasmustautsglia: This comment will be removed in a gradient PR, I just left it as were the other ones so that it would look roughly the same. Once This PR is merged, I will take care of this issue there :)

rasmustautsglia:
!merge

PR #411, opened 2022-11-11
Add possibility to increment the version through GitHub Actions

gersonnoboa:
Add the same work that is now present in all Android repositories, where you can run a GitHub Action that lets you choose the increment type, and Bitrise will create a PR that increments the version number. Also, the same approach is used as in Android where the variable sent through GitHub is replacing the envvar in Bitrise. This is done in a script step in Bitrise, and is the approach recommended by Bitrise themselves.

MOB-1625

### DIFF
--- a/.github/workflows/increment-project-version.yml
+++ b/.github/workflows/increment-project-version.yml
@@ -1,0 +1,21 @@
+name: Increment project version
+
+on:
+  workflow_dispatch:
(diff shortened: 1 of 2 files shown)


dukhovnyi (approved):

gersonnoboa:
!merge

PR #416, opened 2022-11-15
Fix memory leak in Chat and Call views

rasmustautsglia:
Currently when engagement had ended, the following objects were not deallocated from the memory:
1. EngagementViewModel
2. EngagementViewController
3. ChatView
4. ChatViewController
5. ChatViewModel
6. CallView
7. CallViewController
8. CallViewModel
9. RootCoordinator

This PR fixes the memory leaks.

MOB-1559

### DIFF
--- a/GliaWidgets/FoundationBased.Mock.swift
+++ b/GliaWidgets/FoundationBased.Mock.swift
@@ -3,7 +3,7 @@ import Foundation
 
 extension FoundationBased.FileManager {
     static let mock = Self(
-        urlsForDirectoryInDomainMask: { _, _ in [] },
+        urlsForDirectoryInDomainMask: { _, _ in [.mock] },
         fileExistsAtPath: { _ in false },
         createDirectoryAtUrlWithIntermediateDirectories: { _, _, _ in },
         removeItemAtUrl: { _ in },
--- a/GliaWidgets/ViewController/Chat/ChatViewController.swift
+++ b/GliaWidgets/ViewController/Chat/ChatViewController.swift
@@ -52,27 +52,50 @@ class ChatViewController: EngagementViewController, MediaUpgradePresenter,
         view.header.showBackButton()
         view.header.showCloseButton()
 
-        view.numberOfSections = { viewModel.numberOfSections }
-        view.numberOfRows = { viewModel.numberOfItems(in: $0) }
-        view.itemForRow = { viewModel.item(for: $0, in: $1) }
-        view.messageEntryView.textChanged = { viewModel.event(.messageTextChanged($0)) }
-        view.messageEntryView.sendTapped = { viewModel.event(.sendTapped) }
-        view.messageEntryView.pickMediaTapped = { viewModel.event(.pickMediaTapped) }
-        view.messageEntryView.uploadListView.removeTapped = { viewModel.event(.removeUploadTapped($0)) }
-        view.fileTapped = { viewModel.event(.fileTapped($0)) }
-        view.downloadTapped = { viewModel.event(.downloadTapped($0)) }
-        view.callBubbleTapped = { viewModel.event(.callBubbleTapped) }
-        view.choiceOptionSelected = { viewModel.event(.choiceOptionSelected($0, $1)) }
-        view.chatScrolledToBottom = { viewModel.event(.chatScrolled(bottomReached: $0)) }
-        view.linkTapped = { viewModel.event(.linkTapped($0)) }
-        view.selectCustomCardOption = { viewModel.event(
-            .customCardOptionSelected(
-                option: $0,
-                messageId: $1
-            )
-        ) }
+        view.numberOfSections = { [weak viewModel] in
+            viewModel?.numberOfSections
+        }
+        view.numberOfRows = { [weak viewModel] rows in
+            viewModel?.numberOfItems(in: rows)
+        }
+        view.itemForRow = { [weak viewModel] row, section in
+            viewModel?.item(for: row, in: section)
+        }
+        view.messageEntryView.textChanged = { [weak viewModel] text in
+            viewModel?.event(.messageTextChanged(text))
+        }
+        view.messageEntryView.sendTapped = { [weak viewModel] in
+            viewModel?.event(.sendTapped)
+        }
+        view.messageEntryView.pickMediaTapped = { [weak viewModel] in
+            viewModel?.event(.pickMediaTapped)
+        }
+        view.messageEntryView.uploadListView.removeTapped = { [weak viewModel] upload in
+            viewModel?.event(.removeUploadTapped(upload))
+        }
+        view.fileTapped = { [weak viewModel] file in
+            viewModel?.event(.fileTapped(file))
+        }
+        view.downloadTapped = { [weak viewModel] download in
+            viewModel?.event(.downloadTapped(download))
+        }
+        view.callBubbleTapped = { [weak viewModel] in
+            viewModel?.event(.callBubbleTapped)
+        }
+        view.choiceOptionSelected = { [weak viewModel] option, messageId in
+            viewModel?.event(.choiceOptionSelected(option, messageId))
+        }
+        view.chatScrolledToBottom = { [weak viewModel] bottomReached in
+            viewModel?.event(.chatScrolled(bottomReached: bottomReached))
+        }
+        view.linkTapped = { [weak viewModel] url in
+            viewModel?.event(.linkTapped(url))
+        }
+        view.selectCustomCardOption = { [weak viewModel] option, messageId in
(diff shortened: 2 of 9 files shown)


EgorovEI (approved):

GliaWidgets/ViewController/EngagementViewController.swift
```diff
@@ -42,24 +42,24 @@ class EngagementViewController: ViewController, AlertPresenter {
         view.header.backButton.tap = { [weak self] in self?.viewModel.event(.backTapped) }
         view.header.closeButton.tap = { [weak self] in self?.viewModel.event(.closeTapped) }
 
-        viewModel.engagementAction = { action in
+        viewModel.engagementAction = { [weak self] action in
             switch action {
             case .confirm(let conf, let accessibilityIdentifier, confirmed: let confirmed):
-                self.presentConfirmation(
+                self?.presentConfirmation(
```
gersonnoboa, line 48: Maybe here it would be worth it to do a `guard let self = self else { return }` instead to avoid having all the optional selfs? I don't know if that is more preferable.

GliaWidgets/ViewController/Chat/ChatViewController.swift
```diff
@@ -48,31 +48,54 @@ class ChatViewController: EngagementViewController, MediaUpgradePresenter,
     }
 
     // swiftlint:disable function_body_length
-    private func bind(viewModel: ChatViewModel, to view: ChatView) {
+    private func bind(to view: ChatView) {
```
gersonnoboa, line 51: The thing with this solution as i also told you in DMs, is that, while it works, it also means you have to set the view controller's view model before calling this. It's not a contained function, it depends on other factors. Also the name of it loses its meaning, because what is being binded to the view? The previous version was contained and it showed its purpose through the name alone. If others could recommend a way in which we still fix the leak but we keep the previous method signature, it would be ideal. I read that one can capture a parameter weakly, with something like

```
private func bind(viewModel: ChatViewModel, to view: ChatView) {
    weak var weakViewModel = viewModel

    view.numberOfSections = {
        weakViewModel.numberOfSections
    }
    ...
```

What do you guys think?

gersonnoboa: Another option would be to set the view controller's viewModel at first, like it's being done lower with call. That way we don't have to set the view controller's view model before every execution of `bind`. Although that would mean the function has side effects that maybe are not explicit in the name.

rasmustautsglia: 1. I tried out the suggestion you outlined by declaring weakViewModel. Unfortunately, it brought back the memory leak and none of the mentioned objects were deallocated when the engagement ended.
2. I agree that having the name Bind on the method doesn't give any context anymore. What would be a better name for it in the current scenario?
3. I did catch that last comment. Could you please elaborate on that a bit?

rasmustautsglia: `private func bind(viewModel: ChatViewModel, to view: ChatView) {
        view.header.showBackButton()
        view.header.showCloseButton()

        view.numberOfSections = { [weak viewModel] in
            viewModel?.numberOfSections
        }
        view.numberOfRows = { [weak viewModel] rows in
            viewModel?.numberOfItems(in: rows)
`

This however does work.

rasmustautsglia: `private func bind(viewModel: ChatViewModel, to view: ChatView) {
        view.header.showBackButton()
        view.header.showCloseButton()

        view.numberOfSections = { [weak viewModel] in
            viewModel?.numberOfSections
        }
        view.numberOfRows = { [weak viewModel] rows in
            viewModel?.numberOfItems(in: rows)`

This however works

rasmustautsglia: Screenshot 2022-11-16 at 13 19 33

This however works.

gersonnoboa: For 3, I meant that something that we can also do is

```
private func bind(viewModel: ..., to view: ...) {
    self.viewModel = viewModel
    ...
```

That way at least the dependency that this function has on `self.viewModel` is very explicit, and it cannot happen that we call bind and we forget to set `self.viewModel` first. Same thing happens in `CallViewModel.init`, where you initially set `self.call = call` and then just use `self.call` from then onwards.

If my suggestion didn't work, then I guess that it could just be called `bindViewModel(to view:)`. But maybe someone that knows more about memory management than me could suggest if something similar to my original solution would work, because I think that that would be the best way to fix it. To send the view model instead of depending on `self.viewModel`

gersonnoboa: `[weak viewModel]` is also nice. That would be good because we can still use `bind(viewModel: ..., to view:...)` which is a nice name, and we don't depend on `self.viewModel` anymore. What do others think?

igorkravchenko (commented):
Looks good. However unit test are needed here to prove that previously leaked instance for a given scenario gets deallocated now after the fix.

GliaWidgetsTests/Sources/CallViewControllerTests.swift
```diff
@@ -0,0 +1,27 @@
+@testable import GliaWidgets
+import SalemoveSDK
+import XCTest
+
+class CallViewControllerTests: XCTestCase {
+
```
gersonnoboa, line 6: Unneeded space

GliaWidgetsTests/Sources/CallViewControllerTests.swift
```diff
@@ -0,0 +1,27 @@
...
+    func test_call_deinit() {
+
```
gersonnoboa, line 8: Also here. And similar thing happens in the texts below

gersonnoboa (approved):
Although SwiftLint is complaining. But check out [this PR](https://github.com/salemove/ios-sdk-widgets/pull/420) by Egor, which will solve those. So after he merges that to master, you can just rebase and then they will be gone from here as well.

GliaWidgetsTests/Sources/CallViewControllerTests.swift
rasmustautsglia: Fixed [6c181fd](https://github.com/salemove/ios-sdk-widgets/pull/416/commits/6c181fde74064d1185ae3311815fcd2645dc12b3)

GliaWidgets/FoundationBased.Mock.swift
```diff
@@ -3,7 +3,7 @@ import Foundation
 
 extension FoundationBased.FileManager {
     static let mock = Self(
-        urlsForDirectoryInDomainMask: { _, _ in [] },
+        urlsForDirectoryInDomainMask: { _, _ in [.mock] },
```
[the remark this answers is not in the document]
igorkravchenko: I think it is good, just check that other tests still pass after this change. I would also add comment with explanation why this particular mocked method `FileManager. urlsForDirectoryInDomainMask` has to return non empty array.

GliaWidgetsTests/Sources/ChatViewControllerTests.swift
```diff
@@ -0,0 +1,25 @@
...
+import SalemoveSDK
+import XCTest
+
+class ChatViewControllerTests: XCTestCase {
+    func test_chat_deinit() {
+        var testViewController: ChatViewController?
+
+        let autoreleasepoolExpectation = expectation(description: "Autoreleasepool should drain")
+
+        autoreleasepool {
```
igorkravchenko, line 11: See my comment with example of testing deallocation.

GliaWidgets/ViewController/EngagementViewController.swift
```diff
@@ -42,7 +42,8 @@ class EngagementViewController: ViewController, AlertPresenter {
         view.header.backButton.tap = { [weak self] in self?.viewModel.event(.backTapped) }
         view.header.closeButton.tap = { [weak self] in self?.viewModel.event(.closeTapped) }
 
-        viewModel.engagementAction = { action in
+        viewModel.engagementAction = { [weak self] action in
+            guard let self = self else { return }
```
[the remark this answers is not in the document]
rasmustautsglia: This syntax failed in Bitrise due to it being using older version

dukhovnyi (approved):

rasmustautsglia:
!merge

igorkravchenko (changes requested):
@rasmustautsglia I think you still need to update CallViewControllerTests in similar manner as ChatViewControllerTests was updated.

igorkravchenko (approved):

rasmustautsglia:
!merge

PR #419, opened 2022-11-21
Create example integration app

rasmustautsglia:
Currently the integration process took more time than it should. According to recent reports over 20 minutes. To reduce the time for initial setup, a sample app was created to demonstrate the primary use of Glia's WidgetSDK. The sample app allows to start an engagement with text, audio, and video.

MOB-1638


rasmustautsglia:
Ignore the swiftlint comments. It fails in places that are not related to this ticket and Egor has fixed all of them in a separate ticket.

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
```diff
@@ -0,0 +1,117 @@
+import SwiftUI
+import SalemoveSDK
+import GliaWidgets
+
+class ContentViewPresenter: ObservableObject {
+
+    private let config: GliaConfiguration
```
gersonnoboa, line 7: This file has a lot of configurations between GliaConfiguration, Configuration, and Salemove.Configuration, so this variable would benefit from having a descriptive name, like `gliaConfiguration`, even if it's longer.

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
```diff
@@ -0,0 +1,117 @@
...
+                theme: theme
+            )
+        } catch {
+            print("Error starting engagement")
+        }
+    }
+
+    // Configure Glia, for Core SDK.
+    private func initializeCoreSDK(completion: @escaping () -> Void) {
+        let myConfig = config.composeConfiguration()
```
gersonnoboa, line 71: myConfig is not a descriptive name, and here it's particularly not useful because the result of a method executed in `config` is saved in `myConfig`.

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
```diff
@@ -0,0 +1,117 @@
...
+        do {
+            let salemoveConfiguration = try Salemove.Configuration(
+                siteId: myConfig.site,
+                region: myConfig.environment.salemoveRegion,
+                authorizingMethod: .siteApiKey(id: config.apiID, secret: config.apiSecret)
+            )
+
+            Salemove.sharedInstance.configure(
```
gersonnoboa, line 79: Interesting, never realized that if you want to use something from the SDK before the widgets start, you need to configure it separately. The developer experience is not good here :) something for us to think about in the future.

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
```diff
@@ -0,0 +1,117 @@
...
+            )
+        } catch {
+            print("Error has happened: \(error.localizedDescription)")
+        }
+    }
+
+    // Customize and apply Theme
+    private func customizeTheme() -> Theme {
+        let color = ThemeColor(
+            primary: .init(red: 0.52, green: 0.11, blue: 0.92, alpha: 1),
```
gersonnoboa, line 91: Would be good to add here which color this is in plain words, just in case, or just use system colors.

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
```diff
@@ -0,0 +1,117 @@
...
+        theme.minimizedBubble.badge?.fontColor = .white
+        theme.minimizedBubble.badge?.font = .systemFont(ofSize: 14)
+
+        return theme
+    }
+}
+
+// Check if the desired queue is open and available for a certain media.
+private extension SalemoveSDK.Queue {
+    func isSupport(mediaType: MediaType) -> Bool {
```
gersonnoboa, line 110: `func isMediaTypeSupported(_ mediaType: MediaType) -> Bool` maybe?

Example/Glia Sample Integration/ViewModifiers/AvailabilityModifier.swift
```diff
@@ -0,0 +1,16 @@
+import SwiftUI
+
+struct IsAvailable: ViewModifier {
```
gersonnoboa, line 3: I think that this name is a little bit too generic. I see that this is so that if the media type is available, then it's solid, otherwise it's partially opaque. This is generic enough that I agree with not naming it anything related to media types, but maybe something that conveys more what is being done would be good. I see that for example the button also has a modifier called button style, maybe OpacityStyle would be a better name? You are most likely more experienced in SwiftUI than I am, so maybe there's a name that sounds more SwiftUI to this modifier :)

Example/Glia Sample Integration/ViewModifiers/EngagementButtonStyle.swift
```diff
@@ -0,0 +1,22 @@
+import SwiftUI
+
+struct EngagementButtonStyle: ButtonStyle {
+
+    private let backgroundColor = Color(red: 0.52, green: 0.11, blue: 0.92)
```
gersonnoboa, line 5: This color is used two times, so it would be good to have it in a single place. I know that one is for the buttons outside and another one for the widgets. But imagine you are demoing this in D&D and want to say that it's easy to change colors, but in reality you have to change colors in two places. Because we want to convey the message that your current app's color scheme can be easily picked up by the widgets, it makes sense to have them use the same color and just have to change it once.

Example/Glia Sample Integration/Views/ContentView.swift
```diff
@@ -0,0 +1,36 @@
+import SwiftUI
+
+struct ContentView: View {
+
+    @ObservedObject var presenter = ContentViewPresenter()
+
+    var body: some View {
+        VStack(spacing: 16){
```
gersonnoboa, line 8: I see here and in the EngagementButtonStyle that the spacing, paddings, etc are used directly in the code. Is that how normally they are made in SwiftUI? Because normally I would advise to move them to a variable. If you want to create another VStack inside in the future, does that get the parent's spacing or you would have to set it again? In that case, it would be better to add it to a variable.

gersonnoboa (commented):
If there is the possibility of making the image a bit smaller, it would be good. Right now the whole GliaWidgets folder (the one that holds the code for the widgets only, not the whole repository), is 1.5MB. So adding a file of 130KB in the context of that, is a quite significant size increase.

Example/Glia Sample Integration/Extensions/GliaEnvironment.swift
```diff
@@ -0,0 +1,17 @@
+import SalemoveSDK
+import GliaWidgets
+
+extension Environment {
+    var salemoveRegion:  Salemove.Region {
```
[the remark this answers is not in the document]
rasmustautsglia: There are 2 identically structured, yet different types in WidgetSDK, which need the be converted

Example/Glia Sample Integration/ViewModifiers/AvailabilityModifier.swift
rasmustautsglia: That is a good comment and I agree this is very generic. However, I don't think this is necessarily bad. In addition to opacity configuration, this modifier also disables the button entirely, not allowing the user to click on it. This can be used on any button anywhere by just passing in a boolean value

Example/Glia Sample Integration/Views/ContentView.swift
rasmustautsglia: Short answer: Yes in SwiftUI they are used directly in the code.
A bit longer. I use those directly because it is a static modifier. If for instance there was a condition for having padding or spacings dynamic, then most definitely they will be stored in a variable. So in the current situation `let spacing = 16` would be an overkill.

To answer your question about Vstack: VStack within a VStack won't get the spacing of the parent but is rather considered as a single entity(view) within the parent VStack. This would mean that the child VStack will have its own spacing if declared so. This is useful because if your spacing within parent and child is the same, there would be no need for the child in the first place.

Example/Glia Sample Integration/ViewModifiers/EngagementButtonStyle.swift
rasmustautsglia: Fixed [af4a31c](https://github.com/salemove/ios-sdk-widgets/commit/af4a31cbdfdff9ae7acfa5eb81c2eaee841033d5)

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
rasmustautsglia: Fixed [af4a31c](https://github.com/salemove/ios-sdk-widgets/commit/af4a31cbdfdff9ae7acfa5eb81c2eaee841033d5)

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
rasmustautsglia: Fixed [af4a31c](https://github.com/salemove/ios-sdk-widgets/commit/af4a31cbdfdff9ae7acfa5eb81c2eaee841033d5)

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
rasmustautsglia: Fixed [af4a31c](https://github.com/salemove/ios-sdk-widgets/commit/af4a31cbdfdff9ae7acfa5eb81c2eaee841033d5)

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
rasmustautsglia: Fixed [af4a31c](https://github.com/salemove/ios-sdk-widgets/commit/af4a31cbdfdff9ae7acfa5eb81c2eaee841033d5)

Example/Glia Sample Integration/Extensions/GliaEnvironment.swift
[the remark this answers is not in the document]
rasmustautsglia: Fixed [af4a31c](https://github.com/salemove/ios-sdk-widgets/commit/af4a31cbdfdff9ae7acfa5eb81c2eaee841033d5)

Example/Glia Sample Integration/Configuration.swift
```diff
@@ -0,0 +1,25 @@
+import GliaWidgets
+import SalemoveSDK
+
+// Configure the Glia class
+class GliaConfiguration {
+
```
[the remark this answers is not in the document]
rasmustautsglia: Fixed [af4a31c](https://github.com/salemove/ios-sdk-widgets/commit/af4a31cbdfdff9ae7acfa5eb81c2eaee841033d5)

rasmustautsglia:
The size of the image was reduced from 130KB to 29.45KB

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
```diff
@@ -0,0 +1,115 @@
...
+        theme.minimizedBubble.badge?.font = .systemFont(ofSize: 14)
+
+        return theme
+    }
+}
+
+// Check if the desired queue is open and available for a certain media.
+private extension SalemoveSDK.Queue {
+    func isMediaTypeSupported(_ mediaType: MediaType) -> Bool {
+        if state.status == .open && state.media.contains(mediaType) {
```
gersonnoboa, line 109: This can also be changed to just `state.status == .open && state.media.contains(mediaType)` since that will return true or false

Example/Glia Sample Integration/ViewModifiers/AvailabilityModifier.swift
gersonnoboa: As I don't think of a better name that encapsulates both functionalities, then it's fine. If no one else says anything then it should be understandable enough

gersonnoboa (approved):

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
```diff
@@ -0,0 +1,115 @@
+import SwiftUI
+import SalemoveSDK
+import GliaWidgets
+
+class ContentViewPresenter: ObservableObject {
+    private let gliaConfiguration: GliaConfiguration
+    private var theme = Theme()
+    private var queue: SalemoveSDK.Queue?
+
+    @Published var isVideoAvailable: Bool = false
```
EgorovEI, line 10: Why do you need `@Published` attribute?

Example/Glia Sample Integration/ViewModifiers/AvailabilityModifier.swift
```diff
@@ -0,0 +1,15 @@
...
+    let isAvailable: Bool
+
+    init(_ isAvailable: Bool) {
+        self.isAvailable = isAvailable
+    }
+
+    func body(content: Content) -> some View {
+        content
+            .disabled(isAvailable ? false : true)
```
EgorovEI, line 12: will `.disabled(!isAvailable)` work?

Example/Glia Sample Integration.xcodeproj/project.pbxproj
```diff
@@ -0,0 +1,455 @@
...
+		};
+/* End XCConfigurationList section */
+
+/* Begin XCRemoteSwiftPackageReference section */
+		C03A8042292B895B00DDECA6 /* XCRemoteSwiftPackageReference "ios-sdk-widgets" */ = {
+			isa = XCRemoteSwiftPackageReference;
+			repositoryURL = "https://github.com/salemove/ios-sdk-widgets.git";
+			requirement = {
+				branch = master;
+				kind = branch;
```
dukhovnyi, line 441: I'd like to propose using the latest released version but not a `master` branch because it is the guarantee that the example project will use the latest release, but not not-released changes. does it make sense to me guys? @gersonnoboa  @igorkravchenko

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
rasmustautsglia: A view that needs to be rerendered when this value changes, can't do it without @Published property wrapper.

EgorovEI: So `ObservableObject` invalidates View from the box when `@Published` properties are changed and you don't need to subscribe explicitly? 
Something like 
```swift
presenter.$isVideoAvailable
    .sink() {
        reloadViewState()
}
```

Example/Glia Sample Integration.xcodeproj/project.pbxproj
rasmustautsglia: Fixed [d992bda](https://github.com/salemove/ios-sdk-widgets/commit/d992bda8093d149fb48112026a37b81e7267537c)

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
rasmustautsglia: What is the benefit of it?

Example/Glia Sample Integration/ViewModifiers/AvailabilityModifier.swift
rasmustautsglia: Fixed [a4da8d8](https://github.com/salemove/ios-sdk-widgets/commit/a4da8d87adaf8307e7a6192d03a7944cc6346f37)

EgorovEI (approved):

Example/Glia Sample Integration/ViewModifiers/AvailabilityModifier.swift
dukhovnyi: Since this is `ViewModifier` (+ context from SwiftUI) I think it is ok to leave it as is for now.

dukhovnyi (approved):

igorkravchenko (approved):

rasmustautsglia:
!merge

PR #420, opened 2022-11-21
Fixed bottom padding on Survey screen

EgorovEI:
During fixing snapshot tests previously we deleted bottom padding for buttons on Survey screen. It was wrong and I added it back. possiblePrecision usage was removed for almost all tests because it was preventing tests from failing with small changes on UI.

MOB-1645

### DIFF
--- a/GliaWidgets/Coordinator/Call/CallCoordinator.swift
+++ b/GliaWidgets/Coordinator/Call/CallCoordinator.swift
@@ -48,8 +48,12 @@ class CallCoordinator: SubFlowCoordinator, FlowCoordinator {
         )
         return viewController
     }
+}
 
-    private func makeCallViewController(
+// MARK: - Private
+
+private extension CallCoordinator {
+    func makeCallViewController(
         call: Call,
         startAction: CallViewModel.StartAction
     ) -> CallViewController {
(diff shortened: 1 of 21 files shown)


igorkravchenko (commented):
So precision for snapshot tests removed, do tests still pass?

GliaWidgets/Coordinator/Call/CallCoordinator.swift
```diff
@@ -82,30 +86,38 @@ class CallCoordinator: SubFlowCoordinator, FlowCoordinator {
...
-            }
+            self?.handle(event: event)
         }
         return CallViewController(
             viewModel: viewModel,
             viewFactory: viewFactory
         )
     }
+
+    func handle(event: EngagementViewModel.DelegateEvent) {
```
gersonnoboa, line 100: There's something bugging me about having two different methods called `handle` but with different events, but the thing is that it actually makes it easier to write code to handle events because you don't have to call one like `handleEngagementViewModelEvent` and another one `handleCallViewModelEvent`. So If someone else says something about it then we can think about it, otherwise just ignore me :)

GliaWidgets/Glia.swift
```diff
@@ -255,23 +255,27 @@ public class Glia {
...
-                self?.onEvent?(.minimized)
-            case .maximized:
-                self?.onEvent?(.maximized)
-            }
+            self?.handleCoordinatorEvent(event)
         }
         rootCoordinator?.start()
     }
 
+    private func handleCoordinatorEvent(_ event: RootCoordinator.DelegateEvent) {
```
gersonnoboa, line 263: But here there is a `handleCoordinatorEvent`. So perhaps the ones at the top should be called different names, or this one should be called `handle`. Personally I prefer the first one, but as long as both are consistent, it's fine.

gersonnoboa (commented):
There's a lot of SwiftLint stuff in this PR. I guess that it's to reduce warnings in the code, which is fine, but maybe it would be good to have those in a different commit. Let's say something goes wrong with the bottom padding fix and we need to revert it, we would then be reverting also unrelated SwiftLint fixes.

EgorovEI:
> So precision for snapshot tests removed, do tests still pass?

Actually, I found strange behaviour. I run snapshot tests on `master` branch (still works with Roboto fonts) in `Widgets` repo  referring to `master` branch in `Snapshots` repo, which contains references already generated with SF fonts, and many tests didn't fail due to this precision. So I removed it, and tests started to fail. But when I switched to `unified` branch which already works SF, tests continued to fail without any visible diffs. So I added precision back to the tests which continued to fail. 
It seems like the tests are not stable and we might encounter difficulties with this.

EgorovEI:
> There's a lot of SwiftLint stuff in this PR. I guess that it's to reduce warnings in the code

It's not warnings, but errors caused in places I haven't changed.

GliaWidgets/Coordinator/Call/CallCoordinator.swift
EgorovEI: 946e38598032d196fd906015b6d7e2a8e44fd974

GliaWidgets/Glia.swift
EgorovEI: 946e38598032d196fd906015b6d7e2a8e44fd974

dukhovnyi (approved):

igorkravchenko (approved):

rasmustautsglia (approved):

PR #431, opened 2022-11-27
Remove storing chat history locally

dukhovnyi:


GliaWidgets/ViewController/Chat/ChatViewController.Mock.swift
```diff
@@ -55,141 +54,8 @@ extension ChatViewController {
...
         }
 
-        let messageUuid = UUID.incrementing
-        let messageId = { messageUuid().uuidString }
-        let fileUuid = UUID.incrementing
-        let fileId = { fileUuid().uuidString }
-        let queueId = UUID.mock.uuidString
-        let operatorAttachmentURL = URL.mock.appendingPathComponent("image").appendingPathExtension("png")
         chatViewModelEnv.fetchChatHistory = { $0(.success([])) }
-        chatViewModelEnv.chatStorage.messages = { _ in
```
EgorovEI, line 65: Do snapshot tests work after deleting mocked messages?

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -152,20 +149,17 @@ class ChatViewModel: EngagementViewModel, ViewModel {
...
-            self?.interactor.withConfiguration { [weak self] in
-                guard let self = self else { return }
-                // We only proceed to considering enqueue flow if `startAction` is about starting of engagement.
-                guard case .startEngagement = self.startAction else { return }
-                // We enqueue eagerly in case if this is the first engagement for visitor (by  evaluating previous chat history)
-                // or in case if engagement has been restored.
-                if self.chatStorageState().isEmpty() || self.environment.getCurrentEngagement() != nil {
-                    self.enqueue(mediaType: .text)
-                }
+        loadHistory { [weak self] history in
```
EgorovEI, line 152: Does `loadHistory` work for ongoing engagement?

Podfile
```diff
@@ -28,3 +28,11 @@ end
 target 'SnapshotTests' do
   pod 'AccessibilitySnapshot', '0.5.0'
 end
+
+post_install do |installer|
```
gersonnoboa, line 32: Is this part of the changes for this PR as well?

gersonnoboa (commented):
Does this PR mean that now engagements will always start right away? No need to write a message if you have history? If that is the case, we should check if there is documentation about the previous way, and update it. Also, does this affect somehow messages sent by bots? I wouldn't think so but it would be good to check. I will leave it as comment for now, and if it turns out that everything is fine then you can re-request review and I will approve it.

gersonnoboa:
Also a commit description would be nice.

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -526,16 +505,9 @@ extension ChatViewModel {
...
-
-        chatStorageState().storeMessage(
-            message,
-            interactor.queueID,
-            interactor.engagedOperator
-        )
-
         switch message.sender.type {
         case .operator:
+            guard messagesSection.items.first(where: { $0.id == message.id }) == nil else { return }
```
igorkravchenko, line 510: I guess explanation here would be helpful.

GliaWidgets/ViewController/Chat/ChatViewController.Mock.swift
dukhovnyi: I think no, but we will need to update snapshot significantly because lot of changes in UI.

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
dukhovnyi: yes, because `withConfiguration` happens early, so it doesn't require to do it again here.

GliaWidgets/ViewModel/Chat/ChatViewModel.swift
dukhovnyi: it is not needed anymore.

Podfile
dukhovnyi: no more :)

GliaWidgets/ViewController/Chat/ChatViewController.Mock.swift
EgorovEI: Will be good to create ticket for fixing Snapshot test

EgorovEI (approved):

gersonnoboa (approved):

dukhovnyi:
!merge

PR #435, opened 2022-12-08
Add secure conversations welcome screen skeleton

gersonnoboa:
More info in each individual commit.

NOTE: If you want to test the functionality on your simulator/device, you need to go to `Features.swift` and add `.secureConversations` to the `all` array. This is a temporary feature flag that will be removed at the end of the task.


GliaWidgets/View/SecureConversations/SecureConversationsWelcomeView.swift
```diff
@@ -0,0 +1,38 @@
...
+    let header: Header
+
+    init(with style: SecureConversationsWelcomeStyle) {
+        self.style = style
+        self.header = Header(with: style.header)
+
+        super.init()
+
+        setup()
+        layout()
```
dukhovnyi, line 15: `View` class has the necessary methods for settings up and defining UI, so I'd like to propose removing this.
```suggestion
```

GliaWidgets/View/SecureConversations/SecureConversationsWelcomeView.swift
```diff
@@ -0,0 +1,38 @@
...
+    }
+
+    required init?(coder: NSCoder) {
+        fatalError("init(coder:) has not been implemented")
+    }
```
dukhovnyi, line 20: overridden in `View`
```suggestion
```

GliaWidgets/View/SecureConversations/SecureConversationsWelcomeView.swift
```diff
@@ -0,0 +1,38 @@
...
+    override func setup() {
+        super.setup()
+
+        header.title = style.title
+        backgroundColor = style.backgroundColor
+    }
+}
+
+private extension SecureConversationsWelcomeView {
+    func layout() {
```
dukhovnyi, line 31:
```suggestion

    override func defineLayout() {
        super.defineLayout()
```

dukhovnyi (commented):
A few suggestions:
- Let's use feature branch `origin/feature/secure-messaging` until we will not finish secure messaging.
- Does it make sense to create `SecureConversation` namespace and more all related UI stuff there? Or will it be overwork?
```swift
enum SecureConversation {}

extension SecureConversation {
    final class WelcomeViewController: ViewController { ... }
    final class WelcomeView: View { ... }
}
```

EgorovEI (approved):

gersonnoboa:
@dukhovnyi fixed the namespace 16fb9881c81b6f4e04d553a8c6a4538cb93d0ffc but I named it `SecureConversations` because it seems like that is the name of the feature

GliaWidgets/Feature.swift
```diff
@@ -10,6 +10,7 @@ public struct Features: OptionSet {
 
     /// Bubble that is shown in engagement time outside of engagement view.
     public static let bubbleView = Self(rawValue: 1 << 0)
+    public static let secureConversations = Self(rawValue: 2 << 0)
```
dukhovnyi, line 13: This enum was created for optional features that could be managed by the integrator. However, secure messaging is not an optional feature, so I'm not sure that `secureConversations` should be here. Or we can use this approach temporarily and remove it before release.

dukhovnyi (approved):

rasmustautsglia (approved):

GliaWidgets/Feature.swift
gersonnoboa: Indeed I used it as a temporary measure, as this was already here. Once we merge the feature branch, then it will be removed. I have created a ticket for it https://glia.atlassian.net/browse/MOB-1707

igorkravchenko (approved):
Looks good to me. I will base my work on this - I will need to change the internals of  WelcomeViewController and its coordinator.

gersonnoboa:
!merge

PR #437, opened 2022-12-13
Introduce Command - wrapper for closures used in Props for data-driven UI

igorkravchenko:
Add wrapper for closure to be used in `Props` of data-driven UI.
It makes closure conform `Equatable` and `Hashable`, providing info about callee.

MOB-1706

### DIFF
--- a/GliaWidgets/Lib/Command/Command.swift
+++ b/GliaWidgets/Lib/Command/Command.swift
@@ -1,0 +1,67 @@
+/// Developer-friendly wrapper around a closure.
+/// It makes debugging easier by providing callee information.
+/// Note: since closure does not conform to `Equatable`, it will
+/// also be ignored in `Equatable` and `Hashable` implementations
+/// of `Command`.
+///
+/// Example of declaration and execution:
+/// ```
+/// let validateEmail = Command<String> { email in /* email validation goes here */ }
+/// validateEmail("email@example.test")
+/// ```
+struct Command<T>: Hashable {
+    let tag: String
+    let file: String
+    let function: String
+    let line: UInt
+    let closure: (T) -> Void
+
+    init(
+        tag: String = "",
+        file: StaticString = #file,
+        function: StaticString = #function,
+        line: UInt = #line,
+        closure: @escaping (T) -> Void
+    ) {
+        self.tag = tag
+        self.file = "\(file)"
+        self.function = "\(function)"
+        self.line = line
+        self.closure = closure
+    }
+
+    func execute(with value: T) {
+        closure(value)
+    }
+
+    func callAsFunction(_ value: T) {
+        execute(with: value)
+    }
+
+    static func == (lhs: Self, rhs: Self) -> Bool {
+        lhs.tag == rhs.tag &&
+        lhs.file == rhs.file &&
+        lhs.function == rhs.function &&
+        lhs.line == rhs.line
+    }
+
+    func hash(into hasher: inout Hasher) {
+        hasher.combine(tag)
+        hasher.combine(file)
+        hasher.combine(function)
+        hasher.combine(line)
+    }
+}
+
+/// A shorthand for Command for closure with zero parameters.
+typealias Cmd = Command<Void>
+
+extension Cmd {
+    func execute() where T == Void {
+        self.execute(with: ())
+    }
+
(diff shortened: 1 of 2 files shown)


GliaWidgets/Lib/Command/Command.swift
```diff
@@ -0,0 +1,67 @@
...
+
+    func execute(with value: T) {
+        closure(value)
+    }
+
+    func callAsFunction(_ value: T) {
+        execute(with: value)
+    }
+
+    static func == (lhs: Self, rhs: Self) -> Bool {
```
gersonnoboa, line 41: Completely unrelated and of course nothing to do with your code, I have always despised this implementation by Apple. Who thought `lhs` and `rhs` were good names?

gersonnoboa (approved):

GliaWidgets/Lib/Command/Command.swift
igorkravchenko: I think it came from the languages like C/C++. LHS stands for left-hand side. I guess back in the day length of variables did matter. Also in math it is also commonly [used](https://en.wikipedia.org/wiki/Sides_of_an_equation).

GliaWidgets/Lib/Command/Command.swift
```diff
@@ -0,0 +1,67 @@
...
+        self.function = "\(function)"
+        self.line = line
+        self.closure = closure
+    }
+
+    func execute(with value: T) {
+        closure(value)
+    }
+
+    func callAsFunction(_ value: T) {
```
EgorovEI, line 37: I don't actually understand what is the difference between this method and `execute(with value: T)`. Could you please clarify?

GliaWidgets/Lib/Command/Command.swift
gersonnoboa: Ah maybe if you are an English speaker it makes more sense. I don't remember dealing with it in my limited experience with C++ so when I read it the first time I was dumbfounded.

GliaWidgets/Lib/Command/Command.swift
igorkravchenko: Sure: it allows code like this:
```
let validateEmail = Command<String> { email in /* email validation goes here */ }
validateEmail("email@example.test")
```
It makes usage more ergonomic by omitting the call to `execute`. Without it the code would be like this:
```
validateEmail.execute(with: "email@example.test")
```

EgorovEI:
Would be beneficial to add few additional things:
- default values. Something like 
```swift 
static var nop: Command {
    return Command { _ in }
}
```
and 
```swift 
static var nop: Cmd {
    return Cmd { }
}
```
  - Dispatching on queue:
  ```swift
  public func dispatched(on queue: DispatchQueue) -> CommandWith {
    return Command { value in
      queue.async {
        self.execute(with: value)
      }
    }
  }
  ```
  - Chaining:
  ```swift 
  func then(_ another: Command) -> Command {
    return Command { value in
      self.execute(with: value)
      another.execute(with: value)
    }
  }
  ```

igorkravchenko:
> Would be beneficial to add few additional things:
> 
> * default values. Something like
> 
> ```swift
> static var nop: Command {
>     return Command { _ in }
> }
> ```
> 
> and
> 
> ```swift
> static var nop: Cmd {
>     return Cmd { }
> }
> ```
> 
>  * Dispatching on queue:
> 
> ```swift
> public func dispatched(on queue: DispatchQueue) -> CommandWith {
>   return Command { value in
>     queue.async {
>       self.execute(with: value)
>     }
>   }
> }
> ```
> 
>  * Chaining:
> 
> ```swift
> func then(_ another: Command) -> Command {
>   return Command { value in
>     self.execute(with: value)
>     another.execute(with: value)
>   }
> }
> ```

Sure. I'll add those as soon as those are to be used. Also I'd like to see how far I can get without usage of `.nop`.

GliaWidgets/Lib/Command/Command.swift
EgorovEI: Ohh, I see. I totally forgot this `callAsFunction` feature)

EgorovEI (approved):

igorkravchenko:
!merge

PR #452, opened 2023-01-05
Remove storing chat history locally

dukhovnyi:
Secure Conversations feature doesn’t expect to store chat transcript locally and allows fetching it from API for an authenticated visitor. The non-authenticated visitor will see chat history only for ongoing engagement. If an application has been terminated, then SDK will fetch chat history from API after restoring engagement.


gersonnoboa (approved):

EgorovEI (approved):

rasmustautsglia (approved):

dukhovnyi:
!merge

PR #459, opened 2023-01-16
Handle visitor code states and errors

rasmustautsglia:
This PR covers different states and errors related to Visitor Code.
A loading spinner view, generic error text, error debug message, and
refresh functionality was added

MOB-1768


GliaWidgets/CallVisualizer/VisitorCode/VisitorCodeCoordinator.swift
```diff
@@ -54,13 +54,14 @@ extension CallVisualizer {
                     case .closeButtonTap:
                         self?.codeViewController?.presentingViewController?.dismiss(animated: true)
                         self?.delegate?(.closeTap)
+                    case .refreshButtonTap:
```
igorkravchenko, line 57: Why this decision is made in coordinator? I would suggest to move it inside view model, since it is the one responsible for to business logic.

GliaWidgets/Component/Button/Action/ActionButtonStyle.swift
```diff
@@ -18,16 +18,16 @@ public struct ActionButtonStyle: Equatable {
     public var backgroundColor: ColorType
 
     /// Corner radius of the button.
-    public var cornerRaidus: CGFloat?
+    public var cornerRaidus: CGFloat
```
igorkravchenko, line 21: Seems like this a change of **public** API, this is likely to affect integrators.
@dukhovnyi @gersonnoboa do you think that this a breaking change?

igorkravchenko (commented):
Looks solid. Left however a few comments/questions.

GliaWidgets/Component/Button/Action/ActionButtonStyle.swift
gersonnoboa: Yes, it seems like a breaking change, you cannot send `nil` anymore. It doesn't seem that necessary to remove the `?`, or am I missing something? Otherwise it's too much effort to deprecate this. Another issue is that this says `cornerRaidus` and not `cornerRadius`...

GliaWidgets/CallVisualizer/VisitorCode/VisitorCodeViewModel.swift
```diff
@@ -54,22 +55,28 @@ extension CallVisualizer {
         func requestVisitorCode() {
             _ = environment.requestVisitorCode { [weak self] result in
                 switch result {
-                case let .success(code):
-                    self?.receivedVisitorCode = code
+                case let .success(code):                    self?.visitorCodeExpiresAt = code.expiresAt
```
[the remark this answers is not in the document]
rasmustautsglia: fixed [7f7ffb0](https://github.com/salemove/ios-sdk-widgets/pull/459/commits/7f7ffb066e97d3911b01d19d5d1c775aa675a8ff)

GliaWidgets/CallVisualizer/VisitorCode/VisitorCodeViewModel.swift
```diff
@@ -54,22 +55,28 @@ extension CallVisualizer {
...
-            timer = environment.timerProviding.scheduledTimer(
-                withTimeInterval: expDate.timeIntervalSinceNow,
-                repeats: false
-            ) { [weak self] _ in
-                self?.requestVisitorCode()
+            switch viewState {
+            case .error, .loading:
+                break
+            case .success:
+                guard let expDate = visitorCodeExpiresAt else { return }
```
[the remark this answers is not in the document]
rasmustautsglia: fixed [7f7ffb0](https://github.com/salemove/ios-sdk-widgets/pull/459/commits/7f7ffb066e97d3911b01d19d5d1c775aa675a8ff)

GliaWidgets/Component/Button/Action/ActionButtonStyle.swift
rasmustautsglia: I see your reasoning and that's a good point. My thought process was that if it's optional, but we assign a value through init() regardless, there would be no reason for optional. But I understand now. reverted back [7f7ffb0](https://github.com/salemove/ios-sdk-widgets/pull/459/commits/7f7ffb066e97d3911b01d19d5d1c775aa675a8ff)

rasmustautsglia: CornerRaidus -> CornerRadius [c8e833e](https://github.com/salemove/ios-sdk-widgets/pull/459/commits/c8e833e8afdbdc9d37d56da2e33f5685ff45a714)

GliaWidgets/Component/Button/Action/ActionButtonStyle.swift
```diff
@@ -62,7 +62,7 @@ public struct ActionButtonStyle: Equatable {
         titleColor: UIColor,
         textStyle: UIFont.TextStyle = .body,
         backgroundColor: ColorType,
-        cornerRaidus: CGFloat = 4.0,
+        cornerRadius: CGFloat = 4.0,
```
gersonnoboa, line 65: Unfortunately we cannot fix the typo without also doing a breaking change, because the integrator can also have a call to the init and the init requiring `cornerRadius` instead of `cornerRaidus` would break their compilation. I suggest that you leave it like it was before with the typo, and create a ticket that addresses this separately.

GliaWidgets/CallVisualizer/VisitorCode/VisitorCodeCoordinator.swift
rasmustautsglia: fixed [9c8a7dd](https://github.com/salemove/ios-sdk-widgets/pull/459/commits/9c8a7dd8d4bd3e407e572ffcc5e61e33b4868ccc)

GliaWidgets/Component/Button/Action/ActionButtonStyle.swift
rasmustautsglia: Reverted back [90e93ee](https://github.com/salemove/ios-sdk-widgets/pull/459/commits/90e93eeb82019e934ec4faf362d68afee200bb87)

EgorovEI (approved):

igorkravchenko (approved):

GliaWidgets/Component/Button/Action/ActionButtonStyle.swift
```diff
@@ -18,7 +18,7 @@ public struct ActionButtonStyle: Equatable {
     public var backgroundColor: ColorType
 
     /// Corner radius of the button.
-    public var cornerRaidus: CGFloat?
+    public var cornerRaidus: CGFloat
```
gersonnoboa, line 21: As far as I understand, you would still need to put the `?` because integrators, for whatever reason, could have here `nil` as an assignment, and with this change, it would be a breaking change.

GliaWidgets/Component/Button/Action/ActionButtonStyle.swift
```diff
@@ -62,10 +62,10 @@ public struct ActionButtonStyle: Equatable {
         titleColor: UIColor,
         textStyle: UIFont.TextStyle = .body,
         backgroundColor: ColorType,
-        cornerRaidus: CGFloat? = 4.0,
-        shadowOffset: CGSize? = CGSize(width: 0.0, height: 2.0),
-        shadowColor: UIColor? = UIColor.black,
-        shadowRadius: CGFloat? = 2.0,
+        cornerRaidus: CGFloat = 4.0,
```
gersonnoboa, line 65: Same in all of these. If they don't send anything then the default value would be used, but in the previous version, they could send `nil`, which they can't now, so it's a breaking change as well.

gersonnoboa (approved):
I'm gonna approve because the changes are small, but please don't merge it without the changes because otherwise we will create breaking changes.

GliaWidgets/Component/Button/Action/ActionButtonStyle.swift
rasmustautsglia: fixed

GliaWidgets/Component/Button/Action/ActionButtonStyle.swift
rasmustautsglia: fixed

rasmustautsglia:
!merge

PR #461, opened 2023-01-18
Commented after-run integration-spm for pull-request workflow

EgorovEI:
Running integration-spm workflow was commented because of it fails by time out. Example: [build](https://app.bitrise.io/build/fd2eaf0d-bee8-4903-a679-cd75f5191128). Yurii will investigate it later.

MOB-1813

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -268,8 +268,8 @@ workflows:
     description: >-
       Workflow for checking and building pull requests. Does not notify
       anywhere, but shows on the pull request itself.
-    after_run:
-    - integration-spm
+#    after_run:
+#    - integration-spm
   pull-request-snapshot:
     steps:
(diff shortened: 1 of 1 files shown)


EgorovEI:
> I have never done something like this in Bitrise, so I wanted to ask if you are certain that it works as expected? Will Bitrise accept this yml without issues? I don't see why not but I wanted to make sure if it's better to comment rather than to remove the after_run directly and then re-add it through Bitrise's website once it gets fixed.

Oh, sorry, forgot to add that it works, because I tested it with manual running.
https://app.bitrise.io/build/5415699d-6446-42f7-bd25-8b6259408cf8

gersonnoboa (approved):

igorkravchenko (approved):

PR #464, opened 2023-01-23
Add loading and confirmation screen navigation after sending message

gersonnoboa:
When sending a secure message, the button goes to a loading state, and then on success, it navigates to the confirmation view. The error handling is still on TODO, as I don't want to make the PR bigger.

MOB-1698


EgorovEI (approved):

gersonnoboa:
!merge

PR #468, opened 2023-01-31
Rename RootCoordinator

dukhovnyi:
- Renamed RootCoordinator to EngagementCoordinator
- Created dummy CallVisualizer.Coordinator
- Updated Xcode template to avoid automatically adding copyright


Podfile
```diff
@@ -17,7 +17,7 @@ end
 
 target 'GliaWidgets' do
   pod 'PureLayout', '~> 3.1'
-  pod 'SalemoveSDK'
+  pod 'SalemoveSDK', :path => '../ios-sdk/GliaCoreSDK.sources.podspec'
```
rasmustautsglia, line 20: This was changed due to testing purposes, yes? We should be able to revert it now.

dukhovnyi: yes

GliaWidgets/Sources/CallVisualizer/CallVisualizer.Coordinator.swift
```diff
@@ -0,0 +1,49 @@
...
+        ) {
+            let alert = AlertViewController(
+                kind: .screenShareOffer(conf, accepted: accepted, declined: declined),
+                viewFactory: viewFactory
+            )
+            presenter.present(alert, animated: true, completion: nil)
+        }
+
+        func presentCallVisualizerViewController() {
+
```
igorkravchenko, line 26: Seems like unnecessary newline.

GliaWidgets/Sources/CallVisualizer/CallVisualizer.Coordinator.swift
```diff
@@ -0,0 +1,49 @@
...
+        private let viewFactory: ViewFactory
+
+        private var presenter: UIViewController {
+            if var topController = UIApplication.shared.windows.first(where: { $0.isKeyWindow })?.rootViewController {
+                while let presentedViewController = topController.presentedViewController {
+                    topController = presentedViewController
+                }
+
+                return topController
```
igorkravchenko, line 43: I see two problems here:
1. This may crash integrator's application, because we are accessing are controlled by integrator. This means if `topController` will be `UIAlertController`, presentation on it will cause runtime exception that will result in app crash.
2. UIApplication.shared is used directly. This is not testable. Instead Coordinator should have Environment inside which will be UIKitBased.UIApplication that provides access to windows.

GliaWidgets/Sources/CallVisualizer/CallVisualizer.Coordinator.swift
dukhovnyi: fixed - 6d91a2d615cbdacbee4896bad224c116fd07adc4

GliaWidgets/Sources/CallVisualizer/CallVisualizer.Coordinator.swift
dukhovnyi: fixed - 4afa679cdc62918be3038a7caa4b0b3a7f611272

EgorovEI (approved):

GliaWidgets/Sources/CallVisualizer/CallVisualizer.Coordinator.swift
```diff
@@ -0,0 +1,62 @@
...
+        init(presenter: @escaping () -> UIViewController?) {
+            self.getInstance = presenter
+        }
+
+        let getInstance: () -> UIViewController?
+    }
+}
+
+extension CallVisualizer.Presenter {
+    static func topViewController(application: UIKitBased.UIApplication = .live) -> Self {
```
igorkravchenko, line 51: I think we should pass `application` parameter explicitly from Glia.Environment. In general we should avoid implicit (default) parameters for dependencies, because they will slip in during unit testing, triggering live API(s). So I wold suggest to remove default `live`.

GliaWidgets/Public/Glia/Glia.swift
```diff
@@ -100,6 +102,41 @@ public class Glia {
...
+                gcd: environment.gcd,
+                imageViewCache: environment.imageViewCache,
+                timerProviding: environment.timerProviding,
+                uiApplication: environment.uiApplication
+            )
+        )
+
+        callVisualizerCoordinator = .init(
+            viewFactory: viewFactory,
+            presenter: .topViewController(application: .live)
```
igorkravchenko, line 121: think we should pass application parameter explicitly from Glia.Environment. So that resulting line would be:
`presenter: .topViewController(application: environment.uiApplication)`

GliaWidgets/Public/Glia/Glia.swift
```diff
@@ -100,6 +102,41 @@ public class Glia {
...
+            guard let engagement = self?.environment.coreSdk.getCurrentEngagement(),
+                  engagement.source == .callVisualizer,
+                  case .screenShareOffer(let answer) = event else {
+                return
+            }
+
+            self?.callVisualizerCoordinator?.offerScreenShare(
+                with: Theme().alertConfiguration.screenShareOffer,
+                accepted: {
+                    // should show bubble here
```
igorkravchenko, line 133: Is this a hint for future implementation for bubble?

dukhovnyi: that is removed in next PR
This comment fixed here - 6e1fdab26d89b583c47a68413b7683904268724e

GliaWidgets/Public/Glia/Glia.swift
dukhovnyi: fixed - 1b0731d2195a403fc3f09603a2a7a6a9b046ae07

GliaWidgets/Sources/CallVisualizer/CallVisualizer.Coordinator.swift
dukhovnyi: fixed - 19bfbd773c0715dbbd0aa9244fe163c82b6572b9

rasmustautsglia (approved):

igorkravchenko (approved):
Looks good! 👍

dukhovnyi:
!squash

dukhovnyi:
!merge

PR #488, opened 2023-02-14
Fix Cocoapods warning about missed DEPLOYMENT_TARGET value

yurii-glia:


rasmustautsglia (approved):

gersonnoboa (commented):
What was happening? Where was it happening and how does this fix it?

igorkravchenko (commented):
Was it a warning during installation of Cocoapods or during Xcode build phase?

yurii-glia:
During the build, you can see warnings like this:
image
It happens because dependencies have a minimal deployment target by default 9.0, to avoid these warnings we just need to set it to the same deployment target version as we have in our project.
So, this fix should decrease the number of warnings in the project.
@gersonnoboa @igorkravchenko

gersonnoboa:
@yurii-glia it would be good to mention that on the commit message

EgorovEI (approved):

yurii-glia:
@gersonnoboa  done

gersonnoboa (approved):

yurii-glia:
!merge

PR #490, opened 2023-02-15
Add loading of transcript for secure messaging flow

igorkravchenko:
Add minimum functionality to get messages loaded and shown on screen. File download, view and message sending are to be implemented in subsequent tickets.

MOB-1863

### DIFF
--- a/GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
+++ b/GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
@@ -46,51 +46,29 @@ extension SecureChatModel where Chat: CommonEngagementModel, Transcript: CommonE
     }
 }
 
-extension SecureChatModel where Chat == ChatViewModel, Transcript == SecureConversations.TranscriptModel {
-    typealias Action = SecureChatModel<Chat.Action, Transcript.Action>
+extension SecureConversations.ChatWithTranscriptModel {
+    typealias Action = Chat.Action
     typealias ActionCallback = (Action) -> Void
     typealias DelegateEvent = SecureChatModel<Chat.DelegateEvent, Transcript.DelegateEvent>
     typealias DelegateCallback = (DelegateEvent) -> Void
-    typealias Event = SecureChatModel<Chat.Event, Transcript.Event>
+    typealias Event = Chat.Event
 
     var action: ActionCallback? {
         get {
             switch self {
             case let .chat(model):
-                return model.action.map { callback in { action in
-                        switch action {
-                        case let .chat(chatAction):
-                            callback(chatAction)
-                        case .transcript:
-                            break
-                        }
-                    }
-                }
+                return model.action
             case let .transcript(model):
-                return model.action.map { callback in { action in
-                        switch action {
-                        case .chat:
-                            break
-                        case let .transcript(transcriptAction):
-                            callback(transcriptAction)
-                        }
-                    }
-                }
+                return model.action
             }
         }
 
         set {
             switch self {
             case let .chat(model):
-                model.action = newValue.map { callback in { action in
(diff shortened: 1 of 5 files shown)


GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
```diff
@@ -180,43 +156,214 @@ extension SecureConversations {
...
+
         var action: ActionCallback?
         var delegate: DelegateCallback?
 
         var engagementAction: EngagementViewModel.ActionCallback?
         var engagementDelegate: EngagementViewModel.DelegateCallback?
 
+        private let downloader: FileDownloader
+
+        private let isCustomCardSupported: Bool
```
gersonnoboa, line 175: But we don't support custom cards in secure conversations, or are you future proofing?

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
```diff
@@ -180,43 +156,214 @@ extension SecureConversations {
...
+                    return ChatItem(kind: kind)
+                }
+                return item
+            case .visitorMessage(let message, _):
+                message.downloads = downloader.downloads(
+                    for: message.attachment?.files,
+                    autoDownload: .images
+                )
+                return item
+            case .choiceCard(let message, _, _, let isActive):
```
gersonnoboa, line 242: So this model is going to be a unified model or its own? Because if it will be its own, then it doesn't need choice/custom card handling.

GliaWidgets/SecureConversations/Transcript/TranscriptCoordinator.swift
```diff
@@ -15,7 +12,21 @@ extension SecureConversations {
         }
 
         func start() -> ChatViewController {
-            let model = TranscriptModel()
+            let model = TranscriptModel(
```
gersonnoboa, line 15: Ah okay now I see that we are using the `ChatViewController` which probably have some expectations.

gersonnoboa (commented):
Just commenting because I am a bit confused about some things. But otherwise it seems fine, this is essentially what I did when I was investigating this screen. Maybe I should have offered the approach to you guys at that point and we would have saved a lot of time in the discussions we have had.

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
igorkravchenko: As I understand, chat transcript may return response cards. They are not intractable but visitor can see them. That is why this flag is here.

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
igorkravchenko: But choice cards are a part of history, right? In that case they still need to be rendered.

GliaWidgets/SecureConversations/Transcript/TranscriptCoordinator.swift
igorkravchenko: The plan is at some point to swap it to `ChatViewModel`.

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
gersonnoboa: Yep, makes sense.

gersonnoboa (approved):

EgorovEI (approved):

igorkravchenko:
!squash

igorkravchenko:
!merge

PR #491, opened 2023-02-16
Fix issue with alerts not being enabled for interactions

gersonnoboa:
I don't really know why this line slipped through, I guess it was while I was doing some tests. This line made it so that all alerts were disabled, so acceptance tests would fail because you couldn't click on buttons.


gersonnoboa:
This line was incorrectly added [here](https://github.com/salemove/ios-sdk-widgets/pull/486/files#diff-edbe149f91aff9eb6d17d7685a1c0a2f591a09f81bf838b5427fc3a7c2cfeeddR79)

EgorovEI (approved):

igorkravchenko (approved):

PR #493, opened 2023-02-17
Add message sending to Transcript screen

igorkravchenko:
Add ability to send message from secure Transcript screen.

MOB-1740

### DIFF
--- a/GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
+++ b/GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
@@ -173,6 +173,13 @@ extension SecureConversations {
 
         private let downloader: FileDownloader
 
+        private var messageText = "" {
+            didSet {
+                validateMessage()
+                action?(.setMessageText(messageText))
+            }
+        }
+
         private let isCustomCardSupported: Bool
 
         private var isViewLoaded: Bool = false
@@ -287,16 +294,14 @@ extension SecureConversations {
                 start()
                 isViewLoaded = true
             case .messageTextChanged(let text):
-                // TODO: MOB-1740
-                break
+                messageText = text
             case .sendTapped:
-                // TODO: MOB-1740
-                break
+                sendMessage()
             case .removeUploadTapped:
-                // Not supported for transcript.
+                // TODO: MOB-1742
                 break
             case .pickMediaTapped:
-                // Not supported for transcript.
+                // TODO: MOB-1742
                 break
             case .callBubbleTapped:
                 // Not supported for transcript.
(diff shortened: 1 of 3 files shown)


GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
```diff
@@ -358,6 +364,33 @@ extension SecureConversations {
                 return
             }
         }
+
+        @discardableResult
+        private func validateMessage() -> Bool {
```
gersonnoboa, line 369: i don't like this method because of the name, it says an action but returns a boolean. And because it requires the messageText property instead of asking for it as a parameter. But I also understand that this comes from the ChatViewModel, so maybe we should just keep it like that.

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
```diff
@@ -358,6 +364,33 @@ extension SecureConversations {
...
+
+           _ = environment.sendSecureMessage(
+                messageText,
+                nil, // TODO: File uploading MOB-1742
+                environment.queueIds
+            ) { [weak self] result in
+                switch result {
+                case .success:
+                    self?.messageText = ""
+                    self?.delegate?(.chatEngagementRequested)
```
gersonnoboa, line 387: Why do we request a chat engagement after sending a message?

gersonnoboa (commented):
There's a SwiftLint warning

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
igorkravchenko: To allow upgrade to live chat engagement. But maybe I do not understand it correctly, please advise: what should happen when message is successfully sent?

igorkravchenko: Removed this request. Engagement flow will be addressed separately. b6165071268d0f739f15105ba61b2521f786dbf8

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
igorkravchenko: Yes, it is taken from chat implementation. And because it is not data-driven, it does some side effects  along with returning value. Since for now we are tied to ChatView, I think we have to be consistent between models.

igorkravchenko:
> There's a SwiftLint warning

Fixed b6165071268d0f739f15105ba61b2521f786dbf8.

gersonnoboa (approved):

EgorovEI (approved):

igorkravchenko:
!squash

igorkravchenko:
!merge

PR #495, opened 2023-02-20
Add accessibility to welcome screen

gersonnoboa:
Because the accessibility labels and hints need to be localized, all strings from the welcome screen have been also localized. Labels and hints have been added where needed. The commit seems big but the `renderProps()` method in the welcome view had to be moved somewhere else to avoid a SwiftLint issue.

MOB-1710

### DIFF
--- a/GliaWidgets/L10n.swift
+++ b/GliaWidgets/L10n.swift
@@ -685,8 +685,45 @@ public enum L10n {
   }
   public enum MessageCenter {
     public enum Welcome {
+      // Messaging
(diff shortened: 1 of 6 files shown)


yurii-glia (approved):

igorkravchenko (approved):
In order to be sure that accessibility changes are correct for different states and they are not affected but subsequent changes, I would recommend to add accessibility snapshot tests.

EgorovEI (approved):

gersonnoboa:
!merge

PR #498, opened 2023-02-22
Implement unified customization for secure conversations welcome screen

yurii-glia:
MOB-1713


GliaWidgets/SecureConversations/Welcome/SecureConversations.WelcomeStyle.swift
```diff
@@ -920,5 +1295,15 @@ extension SecureConversations.WelcomeStyle {
...
+
+        mutating func apply(
+            configuration: RemoteConfiguration.Color?,
+            assetBuilder: RemoteConfiguration.AssetsBuilder
+        ) {
+            configuration?.value
+                .map { UIColor(hex: $0) }
+                .first
+                .unwrap { color = $0 }
+        }
```
gersonnoboa, line 1307: SwiftLint error because of line length

GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration+SecureConversations.swift
```diff
@@ -0,0 +1,27 @@
+import Foundation
+
+extension RemoteConfiguration {
+    struct SecureConversations: Codable {
```
gersonnoboa, line 4: For others it's also done like this? Because it's quite big, and then adding confirmation and chat transcript later will mean that this struct will have, what, 50 properties at least.

igorkravchenko (approved):
Looks good to me 👍

GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration+SecureConversations.swift
rasmustautsglia: Gerson is making a good point. I think this should be broken into smaller objects, similar to what a CallStyle, or ChatStyle has done.

GliaWidgets/SecureConversations/Welcome/SecureConversations+RemoteConfig.swift
```diff
@@ -0,0 +1,69 @@
+import UIKit
+
+extension SecureConversations.WelcomeStyle {
+
```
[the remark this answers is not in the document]
yurii-glia: fixed - cb21a975fc7adaf4689d775dbcd1a1703f6acd57

GliaWidgets/SecureConversations/Welcome/SecureConversations.WelcomeStyle.swift
yurii-glia: fixed - 83157fb43a4418d4875c221c67224bb3f63cf139

GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration+SecureConversations.swift
yurii-glia: RemoteConfiguration defines JSON structure and now it has 2 level of nesting. If we would like to make a more complex structure and add one more nesting level, we should introduce one more base object and consider how it affects integrators and how difficult it could be to manually create JSON.
However, if we would like to have separate structures, imho, better to create a separate structure like this on the root level, something like "secureConversationWelcomeScreen" and "secureConversationXxxxScreen".

gersonnoboa: I find the suggestion of `secureConversationsWelcomeScreen` a good enough compromise.

EgorovEI (approved):

gersonnoboa (approved):

rasmustautsglia (approved):

PR #499, opened 2023-02-22
Show alert for unavailable message center in chat transcript

gersonnoboa:
Move the availability check to its own struct so that it can be reused on welcome and on chat transcript. A follow up task has been created (MOB-1882) to hide the text field when the message center is unavailable.

MOB-1744

### DIFF
--- a/GliaWidgets/SecureConversations/SecureConversations.Availability.swift
+++ b/GliaWidgets/SecureConversations/SecureConversations.Availability.swift
@@ -1,0 +1,51 @@
+import Foundation
+
+extension SecureConversations {
+    struct Availability {
+        typealias CompletionResult = (Result<Bool, CoreSdkClient.SalemoveError>) -> Void
+        let environment: Environment
+
+        func checkSecureConversationsAvailability(completion: @escaping CompletionResult) {
+            environment.listQueues { queues, error in
+                if let error = error {
+                    completion(.failure(error))
+                    return
+                }
+
+                self.checkQueues(queues, completion: completion)
+            }
+        }
+
+        private func checkQueues(
+            _ queues: [CoreSdkClient.Queue]?,
+            completion: (Result<Bool, CoreSdkClient.SalemoveError>) -> Void
+        ) {
+            guard let queues = queues, isSecureConversationsAvailable(in: queues) else {
+                completion(.success(false))
+                return
+            }
+
+            completion(.success(true))
+        }
+
+        private func isSecureConversationsAvailable(
+            in queues: [CoreSdkClient.Queue]
+        ) -> Bool {
+            let filteredQueues = queues
+                .filter {
+                    self.environment.queueIds.contains($0.id) &&
+                    $0.state.status != .closed &&
+                    $0.state.media.contains(CoreSdkClient.MediaType.messaging)
+                }
+
+            return !filteredQueues.isEmpty
+        }
+    }
+}
+
+extension SecureConversations.Availability {
+    struct Environment {
+        let listQueues: CoreSdkClient.ListQueues
+        let queueIds: [String]
+    }
+}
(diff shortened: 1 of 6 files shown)


GliaWidgets/SecureConversations/SecureConversations.Availability.swift
```diff
@@ -0,0 +1,50 @@
+import Foundation
+
+extension SecureConversations {
+    struct Availability {
+        typealias CompletionResult = (Result<Bool, CoreSdkClient.SalemoveError>) -> Void
+        var environment: Environment
```
yurii-glia, line 6: do we have any specific reason to keep `environment` as `var`?

GliaWidgets/SecureConversations/SecureConversations.Availability.swift
```diff
@@ -0,0 +1,50 @@
...
+            }
+        }
+
+        private func checkQueues(
+            _ queues: [CoreSdkClient.Queue]?,
+            completion: (Result<Bool, CoreSdkClient.SalemoveError>) -> Void
+        ) {
+            guard let queues = queues, isSecureConversationsAvailable(in: queues) else {
+                completion(.success(false))
+
```
yurii-glia, line 25: would you like to keep this newline for some reason or we can remove it?

GliaWidgets/SecureConversations/SecureConversations.Availability.swift
```diff
@@ -0,0 +1,50 @@
...
+        private func isSecureConversationsAvailable(
+            in queues: [CoreSdkClient.Queue]
+        ) -> Bool {
+            let filteredQueues = queues
+                .filter { self.environment.queueIds.contains($0.id) }
+                .filter { $0.state.status != .closed }
+                .filter { $0.state.media.contains(CoreSdkClient.MediaType.messaging) }
+
+            return !filteredQueues.isEmpty
```
yurii-glia, line 40: in sense of optimization, I'd like to propose using one `contains(where:)` or `first(where:)` instead of three filters. What do you think, does it make sense?

GliaWidgets/SecureConversations/SecureConversations.Availability.swift
```diff
@@ -0,0 +1,50 @@
...
+        }
+    }
+}
+
+extension SecureConversations.Availability {
+    struct Environment {
+        var listQueues: CoreSdkClient.ListQueues
+        var queueIds: [String]
```
yurii-glia, line 48: do we need to have the possibility to change these properties in runtime or could we make them as `let`?

GliaWidgets/SecureConversations/SecureConversations.Availability.swift
gersonnoboa: Not really, I kept them as var because of my struggles with the `Queue` topic I talked about earlier in the week. However moving this to let and providing init should also be fine. Applies to the Environment conversation below too.

gersonnoboa: Fixed fbb0d8ef7636a0daeeaf95d166b4084fca2f42f9

GliaWidgets/SecureConversations/SecureConversations.Availability.swift
gersonnoboa: Removed fbb0d8ef7636a0daeeaf95d166b4084fca2f42f9

GliaWidgets/SecureConversations/SecureConversations.Availability.swift
gersonnoboa: Makes complete sense, I would guess that maybe Swift when compiling would optimize it into just making one, but I am not sure, so fbb0d8ef7636a0daeeaf95d166b4084fca2f42f9. In Elixir you can make the first two calls lazy using the `Stream` module and then on the last one you use the `Map` module and the three would be applied at the same time, so that you don't have to do && all the time. But I don't know Swift has something like that.

GliaWidgets/SecureConversations/SecureConversations.Availability.swift
gersonnoboa: Fixed fbb0d8ef7636a0daeeaf95d166b4084fca2f42f9

igorkravchenko (approved):
Looks good to me 👍

yurii-glia (approved):

gersonnoboa:
!merge

PR #501, opened 2023-02-22
Fix file_lenght SwiftLint warning

yurii-glia:
Glia.swift has been separated and functions for starting engagement have been moved to separate file to fix SwiftLint warning about size of the class body.

### DIFF
--- a/GliaWidgets/Public/Glia/Glia.swift
+++ b/GliaWidgets/Public/Glia/Glia.swift
@@ -1,5 +1,5 @@
-import UIKit
 import SalemoveSDK
+import UIKit
 
 /// Engagement media type.
 public enum EngagementKind: Equatable {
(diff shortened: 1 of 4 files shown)


GliaWidgets/Public/Glia/Glia.swift
```diff
@@ -161,116 +151,12 @@ public class Glia {
...
-
     /// Maximizes engagement view if ongoing engagment exists.
     /// Throws error if ongoing engagement not exist.
     /// Use this function for resuming engagement view If bubble is hidden programmatically and you need to
     /// present engagement view.
     public func resume() throws {
-        guard engagement != .none else {
-            throw GliaError.engagementNotExist
-        }
+        guard engagement != .none else { throw GliaError.engagementNotExist }
```
gersonnoboa, line 159: I never saw this when it was created. Unfortunately it's too late, because gramatically it doesn't sound good. `engagementInexistent` or `engagementDoesNotExist` would have been better.

gersonnoboa (approved):

igorkravchenko (approved):

EgorovEI (approved):

GliaWidgets/Public/Glia/Glia.swift
yurii-glia: yep, it has been introduced a long time ago...

PR #502, opened 2023-02-23
Hide back button on welcome and confirmation screens

gersonnoboa:
This requires some changes in the header that affect all other views. Chat transcript wasn’t done because it uses ChatView, which is not data driven. Another PR will be created to deal with that one in isolation.

Let me know if you think of a better approach for it. It feels a little bit cumbersome.

MOB-1866

### DIFF
--- a/GliaWidgets/SecureConversations/Confirmation/SecureConversations.ConfirmationViewModel.swift
+++ b/GliaWidgets/SecureConversations/Confirmation/SecureConversations.ConfirmationViewModel.swift
@@ -53,11 +53,13 @@ extension SecureConversations.ConfirmationViewModel {
         backButtonCmd: Cmd,
         closeButtonCmd: Cmd
     ) -> Header.Props {
-        Header.Props(
+        let backButton = style.header.backButton.map { HeaderButton.Props(tap: backButtonCmd, style: $0) }
+
+        return Header.Props(
             title: style.headerTitle,
             effect: .none,
             endButton: .init(style: style.header.endButton),
-            backButton: .init(tap: backButtonCmd, style: style.header.backButton),
+            backButton: backButton,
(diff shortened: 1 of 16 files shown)


EgorovEI (approved):

GliaWidgets/SecureConversations/Confirmation/SecureConversations.ConfirmationViewModel.swift
```diff
@@ -53,11 +53,16 @@ extension SecureConversations.ConfirmationViewModel {
         backButtonCmd: Cmd,
         closeButtonCmd: Cmd
     ) -> Header.Props {
-        Header.Props(
+        var backButton: HeaderButton.Props?
+        if let backButtonStyle = style.header.backButton {
```
igorkravchenko, line 57: This approach with `if let` works well, but maybe it makes sense to take advantage of `Optional.map` to do unwrapping and assignment in one take?
```
var backButton: HeaderButton.Props? = style.header.backButton.map { backButtonStyle in .init(tap: backButtonCmd, style: backButtonStyle) } 
```

yurii-glia: we introduced `unwrap` function also, perhaps it can be used also.

rasmustautsglia (approved):

GliaWidgets/SecureConversations/Confirmation/SecureConversations.ConfirmationViewModel.swift
gersonnoboa: Fix 8b6c42659f6ac84f0cb4f9f8cf8e4335fed6cc05

igorkravchenko (approved):
Looks good 👍

PR #504, opened 2023-02-23
Fix ScreenSharing screen navigation issue

EgorovEI:
Fix releasing reference to ScreenSharing viewController in delegate closure


EgorovEI:
!squash

igorkravchenko (approved):
Looks good to me. One thing - command invocations, like
`tap.execute()` can be simplified to just `tap()`.

PR #506, opened 2023-02-27
Fix call visualizer double video call issue

rasmustautsglia:
Fix an issue when during the one-way video, visitors can accept two-way video which is displayed on top of the one-way video screen.

### DIFF
--- a/GliaWidgets/Sources/CallVisualizer/Coordinator/CallVisualizer.Coordinator.swift
+++ b/GliaWidgets/Sources/CallVisualizer/Coordinator/CallVisualizer.Coordinator.swift
@@ -96,6 +96,11 @@ extension CallVisualizer {
                 .present(alert, animated: true)
         }
 
+        func handleAcceptedUpgrade() {
+            guard videoCallCoordinator == nil else { return }
+            showVideoCallViewController()
+        }
+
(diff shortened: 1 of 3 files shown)


gersonnoboa (approved):

yurii-glia (approved):

GliaWidgets/Sources/CallVisualizer/Coordinator/CallVisualizer.Coordinator.swift
```diff
@@ -96,6 +96,11 @@ extension CallVisualizer {
                 .present(alert, animated: true)
         }
 
+        func handleAcceptedUpgrade() {
+            guard videoCallCoordinator == nil else { return }
+            showVideoCallViewController()
+        }
+
```
rasmustautsglia, line 103: > I wonder if unit test can be added to enforce the fix.

This is a bit hard to test or perhaps even not necessary and here is my reasoning:
By looking at this method above, you can see that the method checks for the presence of `videoCallCoordinator`.
If there isn't one, it will create one in a subsequent method(`showVideoCallViewController`). So to test for instance if `videoCallCoordinator` is present after calling this method will result in the same result

rasmustautsglia:
!merge

PR #508, opened 2023-02-27
Clear transcript inputs and show delivered message

igorkravchenko:
Handle sending and delivering of visitor message in secure transcript screen.

MOB-1887


GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
```diff
@@ -429,26 +437,116 @@ extension SecureConversations {
...
+           case let .failure(error):
+               // TODO: MOB-1874
+               break
            }
-        }
+       }
+
+        // Clear inputs, thus disabling them
+        // and preventing the message to be sent again.
+        messageText = ""
```
gersonnoboa, line 479: NIT: maybe these cleanup actions can be its own small method. Not super necessary, but it seems to me here like the send has already several parts: the validation, the preparation (of files and stuff), the sending, and the cleanup.

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
```diff
@@ -429,26 +437,116 @@ extension SecureConversations {
...
+        fileUploadListModel.removeSucceededUploads()
+        action?(.scrollToBottom(animated: true))
+    }
+
+    func replace(
```
gersonnoboa, line 484: NIT: Same, since we are stuck with having to bring over things from chat screen, at least in my opinion, they could be made slightly easier to work with if we separate this into a couple of smaller scopes to make it easier to read and modify in the future.

gersonnoboa (commented):
Don't worry about the warning in `TranscriptCoordinator.swift:21`, I have fixed it in another PR.

gersonnoboa (approved):

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
igorkravchenko: Moved it into separate method 6564e2daf2894759c00f133f1b98524324e3c750.

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
igorkravchenko: I think maybe it makes sense to have it in one place, instead of copying overt here. However I would like that to be done apart from this PR to avoid distraction from the feature development. I will create a separate ticket for this.

yurii-glia (approved):

EgorovEI (approved):

igorkravchenko:
!squash

PR #510, opened 2023-02-27
Allow welcome screen send button press regardless of touch coordinates

gersonnoboa:
If you press the button dead center, where the "Send" text is, the button doesn't react. You need to press it on the sides where there is no text for it to work. This is because the stack view is absorbing the touch event and not relaying it to the UIControl. It doesn't happen to the sides because the stack view doesn't extend to the whole width of the button.

MOB-1891


igorkravchenko (approved):

yurii-glia (approved):

EgorovEI (approved):

gersonnoboa:
!merge

PR #512, opened 2023-02-28
Rename fontColor field of NumberSlot element

EgorovEI:
For consistency we agreed to rename numberSlot.fontColor property to foreground.

### DIFF
--- a/GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration+CallVisualizer.swift
+++ b/GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration+CallVisualizer.swift
@@ -9,9 +9,11 @@ extension RemoteConfiguration {
     struct VisitorCode: Codable {
         let title: Text?
         let actionButton: Button?
-        let numberSlot: NumberSlot?
+        let numberSlotText: Text?
+        let numberSlotBackground: Layer?
         let background: Layer?
         let loadingProgressColor: Color?
+        let closeButtonColor: Color?
     }
 
     struct ScreenSharing: Codable {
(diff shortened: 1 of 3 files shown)


gersonnoboa (commented):
So other things have `foreground` as the name for the font color instead of `fontColor`? Cause imo `foreground` is a lot worse name than `fontColor`

EgorovEI:
> So other things have `foreground` as the name for the font color instead of `fontColor`? Cause imo `foreground` is a lot worse name than `fontColor`

Since `Text` element, that is used for every screen, already contains `foreground` field, we decided to keep consistency there also

GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration+CallVisualizer.swift
```diff
@@ -24,6 +24,6 @@ extension RemoteConfiguration {
     struct NumberSlot: Codable {
         let background: Layer?
         let font: Font?
-        let fontColor: Color?
+        let foreground: Color?
```
yurii-glia, line 27: I raised this question in schema repo, however, does it look like `Text`?

EgorovEI: Almost. I thought to use `Text`, but `Text` contains `background` as `Color` and `NumberSlot` contains `background` as `Layer`

gersonnoboa (approved):

igorkravchenko (approved):

GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration+CallVisualizer.swift
EgorovEI: 5ba9dd172544bb117127667abf2b04729e058ca8

rasmustautsglia (approved):

yurii-glia (approved):

yurii-glia:
!squash

PR #522, opened 2023-03-06
Fix restoring Call Visualizer engagement

EgorovEI:
[MOB-1902](https://glia.atlassian.net/browse/MOB-1902)


rasmustautsglia (approved):
Looks good!

GliaWidgets/Sources/CallVisualizer/CallVisualizer.swift
```diff
@@ -81,22 +51,54 @@ public final class CallVisualizer {
...
+    ///  needs to be passed in as an associated value. If, however, embedding is chosen, the view into which you want to
+    ///  embed it has to be passed in as an associated value.
+    public func showVisitorCodeViewController(
+        by presentation: Presentation
+    ) {
+        coordinator.showVisitorCodeViewController(by: presentation)
     }
+}
 
+// MARK: - Internal
```
gersonnoboa, line 89: Maybe then the extension can be private? It doesn't seem like any of these methods should be accessible even internally.

gersonnoboa (commented):
Would be good to add a commit description about what was fixed and how this accomplishes it

yurii-glia (approved):

GliaWidgets/Sources/CallVisualizer/CallVisualizer.swift
EgorovEI: Methods in that extension are accessed internally, from `Glia` class

PR #523, opened 2023-03-06
Handle error when sending message from transcript fails

igorkravchenko:
Show error alert in case of send message request failure.

MOB-1874


gersonnoboa (approved):

yurii-glia (approved):

igorkravchenko:
!merge

PR #524, opened 2023-03-06
Change header title after upgrade from transcript to chat

igorkravchenko:
Change header title after upgrade from secure transcript to chat.

MOB-1915

### DIFF
--- a/GliaWidgets/Sources/View/Chat/ChatView.swift
+++ b/GliaWidgets/Sources/View/Chat/ChatView.swift
@@ -47,7 +47,11 @@ class ChatView: EngagementView {
 
     private var heightCache: [String: CGFloat] = [:]
 
-    var props: Props
+    var props: Props {
+        didSet {
+            renderHeaderProps()
+        }
+    }
 
     init(
         with style: ChatStyle,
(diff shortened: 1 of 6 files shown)


GliaWidgets/Sources/View/Chat/ChatView.swift
```diff
@@ -47,7 +47,11 @@ class ChatView: EngagementView {
 
     private var heightCache: [String: CGFloat] = [:]
 
-    var props: Props
+    var props: Props {
+        didSet {
+            header.props = props.header
```
gersonnoboa, line 52: Do you think that it would be beneficial to make a method called `renderProps` like in the other views?

gersonnoboa (approved):

GliaWidgets/Sources/View/Chat/ChatView.swift
yurii-glia: agree with @gersonnoboa, it makes sense to keep consistency and have a `renderProps` function even if it will have only one line.

yurii-glia (approved):

GliaWidgets/Sources/View/Chat/ChatView.swift
igorkravchenko: I don't mind having dedicated method for rendering props. However ChatView is not fully data driven, so for now I added  `renderHeaderProps()` because that what is really does at the moment. 2d40d7f83be93d2a6188aa8ea2074c38cdbd67f0

igorkravchenko:
!squash

igorkravchenko:
!merge

PR #526, opened 2023-03-07
Change header title and skip enqueueing state for secure flow

igorkravchenko:
Make start of engagement implicit by skipping enqueueing state and changing title only when engagement is created for secure messaging flow.
Also move choice cards extension to separate file to address linter warning about class body length of `ChatViewModel`.

MOB-1920

### DIFF
--- a/GliaWidgets/Sources/ViewModel/Chat/ChatViewModel+ChoiceCards.swift
+++ b/GliaWidgets/Sources/ViewModel/Chat/ChatViewModel+ChoiceCards.swift
@@ -1,0 +1,62 @@
+import Foundation
+import SalemoveSDK
+
+extension ChatViewModel {
+    func sendChoiceCardResponse(_ option: ChatChoiceCardOption, to messageId: String) {
+        guard let option = option.asSingleChoiceOption() else { return }
+        environment.sendSelectedOptionValue(option) { [weak self] result in
+            guard let self = self else { return }
+
+            switch result {
+            case .success(let message):
+                guard
+                    let selection = message.attachment?.selectedOption
+                else { return }
+
+                self.respond(to: messageId, with: selection)
+
+            case .failure:
+                self.showAlert(
+                    with: self.alertConfiguration.unexpectedError,
+                    dismissed: nil
+                )
+            }
+        }
+    }
+
+    private func respond(to choiceCardId: String, with selection: String?) {
+        guard let index = messagesSection.items
+            .enumerated()
+            .first(where: {
+                guard case .choiceCard(let message, _, _, _) = $0.element.kind
+                else { return false }
+                return message.id == choiceCardId
+            })?.offset
+        else { return }
+
+        let choiceCard = messagesSection[index]
+
+        guard case .choiceCard(
+            let message,
+            let showsImage,
+            let imageUrl,
+            _
+        ) = choiceCard.kind else { return }
+
+        message.attachment?.selectedOption = selection
+        message.queueID = interactor.queueID
+        let item = ChatItem(kind: .choiceCard(
+            message,
+            showsImage: showsImage,
+            imageUrl: imageUrl,
+            isActive: false
+        ))
+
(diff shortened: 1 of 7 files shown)


GliaWidgets/Sources/ViewModel/Chat/ChatViewModel+ChoiceCards.swift
```diff
@@ -0,0 +1,63 @@
+import Foundation
+import SalemoveSDK
+
+extension ChatViewModel {
+    func sendChoiceCardResponse(_ option: ChatChoiceCardOption, to messageId: String) {
+
```
yurii-glia, line 6:
```suggestion
```

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel+ChoiceCards.swift
```diff
@@ -0,0 +1,63 @@
+import Foundation
+import SalemoveSDK
+
+extension ChatViewModel {
+    func sendChoiceCardResponse(_ option: ChatChoiceCardOption, to messageId: String) {
```
yurii-glia, line 5: nit, since the param name is not a part of the signature, I think it should not be omitted.

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel+ChoiceCards.swift
```diff
@@ -0,0 +1,63 @@
...
+        environment.sendSelectedOptionValue(option) { [weak self] result in
+            guard let self = self else { return }
+
+            switch result {
+            case .success(let message):
+                guard
+                    let selection = message.attachment?.selectedOption
+                else { return }
+
+                self.respond(to: messageId, with: selection)
```
yurii-glia, line 17: I have an interesting question :) 
you got the option from the response, however, messageId from params. Perhaps we can make decision to get both from the response, or both from params (and don't need to have one more guard :)

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel+ChoiceCards.swift
```diff
@@ -0,0 +1,63 @@
...
+                self.showAlert(
+                    with: self.alertConfiguration.unexpectedError,
+                    dismissed: nil
+                )
+            }
+        }
+    }
+
+    private func respond(to choiceCardId: String, with selection: String?) {
+        guard let index = messagesSection.items
```
yurii-glia, line 29: - since there could be a lot of items, perhaps we can optimize a little bit this enumeration and make it lazy? would it make sense?
```swift
guard let index = messagesSection.items.lazy
```

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel+ChoiceCards.swift
igorkravchenko: These lines moved to separate file from ChatViewModel to address linter warning on class body length. They are not new. I would hold off changing them now since they are not related to the goal of the PR.

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel+ChoiceCards.swift
```diff
@@ -0,0 +1,62 @@
...
+            case .failure:
+                self.showAlert(
+                    with: self.alertConfiguration.unexpectedError,
+                    dismissed: nil
+                )
+            }
+        }
+    }
+
+    private func respond(to choiceCardId: String, with selection: String?) {
```
gersonnoboa, line 27: I know that this code was moved only but wow, `respond` does SO much.

gersonnoboa (approved):

igorkravchenko:
!squash

EgorovEI (approved):

igorkravchenko:
!merge

PR #528, opened 2023-03-08
Support sending secure files with empty message

gersonnoboa:
Allow sending a secure file with an empty message from both the welcome screen and the chat transcript screen. Also added a check for file upload limit, just as the secure conversations chat transcript is doing.

MOB-1904


yurii-glia (approved):

rasmustautsglia (approved):

gersonnoboa:
!merge

PR #530, opened 2023-03-08
Close Visitor Code When Engagement Starts

rasmustautsglia:
When SDK accepts an operator engagement request, the visitor code will automatically close. This is only for the alert presentation mode, as the embedded one will not be affected.

MOB-1932 and MOB-1934


EgorovEI (approved):

rasmustautsglia:
!merge

PR #531, opened 2023-03-08
Close screen sharing view silently when video call is present

rasmustautsglia:
When visitor was on screen sharing view, and they accepted video call view, if they clicked back button from video view they should have landed in main view. However they were back in screen sharing view. This PR fixes it

MOB-1925


gersonnoboa (approved):

EgorovEI (approved):

rasmustautsglia:
!merge

PR #533, opened 2023-03-09
Remove the back button on secure conversations chat transcript

gersonnoboa:
SORRY THAT THIS TOOK SO LONG :(

The current logic correctly identifies when to show the secure conversations title or the chat title. This logic is encapsulated into its own function so that it can be used to determine back button behavior. The secure chat transcript header style is the same as the regular chat style but with the back button set as `nil`.

MOB-1888

### DIFF
--- a/GliaWidgets/Sources/Component/Header/Header.swift
+++ b/GliaWidgets/Sources/Component/Header/Header.swift
@@ -60,7 +60,7 @@ final class Header: BaseView {
     }
 
     func renderProps() {
-        backButton?.isEnabled = props.backButton != nil
+        backButton?.isHidden = props.backButton == nil
         if let backButtonProps = props.backButton {
             backButton?.props = backButtonProps
         }
--- a/GliaWidgets/Sources/ViewController/Chat/ChatViewController.swift
+++ b/GliaWidgets/Sources/ViewController/Chat/ChatViewController.swift
@@ -182,21 +182,25 @@ class ChatViewController: EngagementViewController, MediaUpgradePresenter,
         guard let chatView: ChatView = view as? ChatView else { return }
         let headerProps = chatView.props.header
         let headerTitle: String
-        switch viewModel {
-        case let .chat(chatViewModel):
-            headerTitle = Self.headerTitleForChatModel(
-                chatViewModel,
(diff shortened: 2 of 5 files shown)


GliaWidgets/Sources/Component/Header/Header.swift
```diff
@@ -60,7 +60,7 @@ final class Header: BaseView {
     }
 
     func renderProps() {
-        backButton?.isEnabled = props.backButton != nil
+        backButton?.isHidden = props.backButton == nil
         if let backButtonProps = props.backButton {
             backButton?.props = backButtonProps
         }
```
rasmustautsglia, line 66: I guess here you check if `props.backButton` is there before hiding `backButton` and hide it in the `else` block. Both ways works though :)

rasmustautsglia (approved):
Looks good! I added a little comment that isn't vital if left unaddressed.

igorkravchenko (approved):
Looks good to me.

GliaWidgets/Sources/ViewController/Chat/ChatViewController.swift
```diff
@@ -224,15 +228,56 @@ extension ChatViewController {
...
-        return headerTitle
+    private static func currentChatModelType(
+        _ viewModel: SecureConversations.ChatWithTranscriptModel
+    ) -> CurrentChatModelType {
+        switch viewModel {
+        case .chat(let chatViewModel):
+            // Even though we are using the chat view model already,
+            // if the engagement is not active, we still need to show
+            // secure transcript UI.
+            if chatViewModel.shouldSkipEnqueueingState {
```
EgorovEI, line 239: Does it intersects somehow with regular Live Chat history (transcript)?

GliaWidgets/Sources/ViewController/Chat/ChatViewController.swift
```diff
@@ -224,15 +228,56 @@ extension ChatViewController {
...
+                return chatViewModel.activeEngagement != nil
```
EgorovEI, line 240: Does it intersects somehow with regular Live Chat history (transcript)?

gersonnoboa:
PR will be closed because it breaks back button on live chat.

PR #536, opened 2023-03-10
Add loading of transcript and unread count in parallel

igorkravchenko:
Add ability to combine unread count and transcript request required for message divider implementation.
With this change both requests will start separately but always finish together, so then they can already be displayed with the new message divider if needed.

MOB-1949

### DIFF
--- a/GliaWidgets/SecureConversations/SecureConversations.MessagesWithUnreadCountLoader.swift
+++ b/GliaWidgets/SecureConversations/SecureConversations.MessagesWithUnreadCountLoader.swift
@@ -1,0 +1,93 @@
+import Foundation
+
+extension SecureConversations {
+    struct MessagesWithUnreadCount {
+        var messages: [ChatMessage]
+        var unreadCount: Int
+    }
+
+    /// Due to async nature of requests, we need to make sure
+    /// that `unread message count` and `transcript messages`
+    /// are delivered at once. This entity makes it possible.
+    struct MessagesWithUnreadCountLoader {
+        static let unreadCountFallbackTimeoutSeconds: TimeInterval = 3.0
+
+        struct TimeoutError: Error {}
+
+        var environment: Environment
+
+        func loadMessagesWithUnreadCount(callback: @escaping (Result<MessagesWithUnreadCount, Error>) -> Void) {
+            typealias ReactiveSwift = CoreSdkClient.ReactiveSwift
+            let unreadCountTimeoutProducer = ReactiveSwift.SignalProducer<Void, Never>(value: ())
+            let unreadCountRequestSource = ReactiveSwift.Signal<Result<Int, Error>, Never>.pipe()
+            let unreadCountTimeoutSource = ReactiveSwift.Signal<Result<Int, Error>, Never>.pipe()
+            // In case if `unreadCount` will never be delivered,
+            // (for example if sockets are disconnected, because
+            // `unreadCount` is based on sockets)
+            // we will still receive callback by timeout.
+            let unreadCountRequestWithTimeoutSource = ReactiveSwift.Signal.merge(
+                unreadCountRequestSource.output,
+                unreadCountTimeoutSource.output
+            ).take(first: 1)
+
+            let messagesSource = ReactiveSwift.Signal<Result<[ChatMessage], CoreSdkClient.SalemoveError>, Never>.pipe()
+
+            let combined = ReactiveSwift.Signal.zip(
+                messagesSource.output,
+                unreadCountRequestWithTimeoutSource
+            )
+
+            unreadCountTimeoutProducer
+                .delay(Self.unreadCountFallbackTimeoutSeconds, on: environment.scheduler)
+                .startWithCompleted {
+                    unreadCountTimeoutSource.input.send(value: .failure(TimeoutError()))
+                }
+
+            combined
+                .observe(on: environment.scheduler)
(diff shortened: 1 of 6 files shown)


GliaWidgets/SecureConversations/SecureConversations.UnreadMessageLoader.swift
```diff
@@ -0,0 +1,96 @@
...
+
+extension SecureConversations {
+    struct MessagesWithUnreadCount {
+        var messages: [ChatMessage]
+        var unreadCount: Int
+    }
+
+    /// Due to async nature of requests, we need to make sure
+    /// that `unread message count` and `transcript messages`
+    /// are delivered at once. This entity makes it possible.
```
yurii-glia, line 12: would we like to have it as doc or as comment?
```suggestion
    // Due to async nature of requests, we need to make sure
    // that `unread message count` and `transcript messages`
    // are delivered at once. This entity makes it possible.
```

GliaWidgets/SecureConversations/SecureConversations.UnreadMessageLoader.swift
```diff
@@ -0,0 +1,96 @@
...
+    struct MessagesWithUnreadCountLoader {
+        static let unreadCountFallbackTimeoutSeconds: TimeInterval = 3.0
```
yurii-glia, line 14: should it be 3 sec or has it been doubled to 6sec?

GliaWidgets/SecureConversations/SecureConversations.UnreadMessageLoader.swift
igorkravchenko: Doc for internal usage.

GliaWidgets/SecureConversations/SecureConversations.UnreadMessageLoader.swift
igorkravchenko: That is different delay, like in [Androids version](https://github.com/salemove/android-sdk-widgets/blob/29dd51c48ee3ea803745a05df178f0c9d4cf7e5f/widgetssdk/src/main/java/com/glia/widgets/core/secureconversations/domain/GetUnreadMessagesCountWithTimeoutUseCase.kt#L19). 6 sec will be for marking messages read.

gersonnoboa: So messages get marked as read after three seconds, but the message divider gets marked as read after six? That sounds different to what we discussed [in this thread](https://salemove.slack.com/archives/C047N9EAG8Y/p1677076646000439?thread_ts=1675767504.581449&cid=C047N9EAG8Y).

GliaWidgets/SecureConversations/SecureConversations.UnreadMessageLoader.swift
```diff
@@ -0,0 +1,96 @@
...
+                // of failure for both requests. Same for success - ignore
+                // unreadCount failure in case of successful loading of messages.
+                switch (messageResult, unreadCountResult) {
+                case let (.success(messages), .success(unreadCount)):
+                    return .success(MessagesWithUnreadCount(messages: messages, unreadCount: unreadCount))
+                case let (.success(messages), .failure):
+                    return .success(MessagesWithUnreadCount(messages: messages, unreadCount: .zero))
+                case let (.failure(error), .failure):
+                    return .failure(error)
+                case let (.failure(error), .success):
```
gersonnoboa, line 76: Are these cases split for better readability? Because they both return the same regardless of what the second value is.

GliaWidgets/SecureConversations/SecureConversations.UnreadMessageLoader.swift
igorkravchenko: This PR is not about marking messages as read. It is about merging two requests into one. Three seconds is the timeout for unread count request.

GliaWidgets/SecureConversations/SecureConversations.UnreadMessageLoader.swift
igorkravchenko: Seems like indeed those can be combined for brevity. 935ba0b72b25c657b3d8cce6f03fd05ed8cc5414

GliaWidgets/SecureConversations/SecureConversations.UnreadMessageLoader.swift
igorkravchenko: I've updated PR description to make it more clear.

GliaWidgets/SecureConversations/SecureConversations.MessagesWithUnreadCountLoader.swift
```diff
@@ -0,0 +1,91 @@
+extension SecureConversations {
```
yurii-glia, line 1: do we need to add `import Foundation` for SPM support?

yurii-glia (approved):

gersonnoboa (approved):

igorkravchenko:
!squash

GliaWidgets/SecureConversations/SecureConversations.MessagesWithUnreadCountLoader.swift
igorkravchenko: Good catch. Added it e20b1b5aaad07503d4989e9de5aef9f0d165c4f4.

PR #537, opened 2023-03-12
Fix long messages being displayed out of bounds

gersonnoboa:
These constraints are important, so they should be required. When set to `.defaultHigh`, under some circumstances, they would not be respected and the message would appear out of bounds.

MOB-1941


yurii-glia (approved):

igorkravchenko (approved):

gersonnoboa:
!merge

PR #538, opened 2023-03-13
Add Listener for engagement start in Call Visualizer

rasmustautsglia:
In order to trigger callback for embedded visitor code view, a listener for engagement start needed to be refactored.

MOB-1934

### DIFF
--- a/GliaWidgets/Public/Glia/Glia.swift
+++ b/GliaWidgets/Public/Glia/Glia.swift
@@ -254,6 +254,12 @@ public class Glia {
 private extension Glia {
     func startObservingInteractorEvents() {
         interactor?.addObserver(self) { [weak self] event in
+            // Since engagement is accepted before current engagement is established,
+            // this case needs to be checked before any other so that visitor code alert
+            // can be dismissed properly and visitor code embedded view can trigger callback.
+            if case .engagementRequestAccepted = event {
+                self?.callVisualizer.handleEngagementRequestAccepted()
+            }
             guard let engagement = self?.environment.coreSdk.getCurrentEngagement(), engagement.source == .callVisualizer else {
                 return
             }
--- a/GliaWidgets/Sources/CallVisualizer/Coordinator/CallVisualizer.Coordinator.swift
+++ b/GliaWidgets/Sources/CallVisualizer/Coordinator/CallVisualizer.Coordinator.swift
@@ -34,6 +34,13 @@ extension CallVisualizer {
                 switch event {
                 case .closeTap:
                     self?.visitorCodeCoordinator = nil
+                case .engagementAccepted:
+                    switch coordinator.presentation {
+                    case let .embedded(_, onEngagementAccepted: callback):
+                        callback()
+                    case .alert:
+                        break
+                    }
                 }
             }
 
@@ -117,6 +124,7 @@ extension CallVisualizer {
         }
 
         func handleEngagementRequestAccepted() {
+            visitorCodeCoordinator?.delegate?(.engagementAccepted)
             visitorCodeCoordinator?.codeViewController?.dismiss(animated: true)
             visitorCodeCoordinator = nil
         }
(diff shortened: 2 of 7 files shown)


GliaWidgets/Sources/CallVisualizer/Coordinator/CallVisualizer.Coordinator.swift
```diff
@@ -34,6 +34,13 @@ extension CallVisualizer {
                 switch event {
                 case .closeTap:
                     self?.visitorCodeCoordinator = nil
+                case .engagementAccepted:
+                    switch coordinator.presentation {
+                    case let .embedded(_, onEngagementAccepted: callBack):
```
gersonnoboa, line 39: `callback` is fine

GliaWidgets/Public/Glia/Glia.swift
```diff
@@ -254,6 +254,10 @@ public class Glia {
 private extension Glia {
     func startObservingInteractorEvents() {
         interactor?.addObserver(self) { [weak self] event in
+            /// Since engagement is accepted before current engagement is established, this case needs to be checked before any other so that visitor code alert can be dismissed properly and visitor code embedded view can trigger callback
+            if case .engagementRequestAccepted = event {
```
gersonnoboa, line 258: Don't know what others think but shouldn't it be that rather `self?.callVisualizer` should get the event and decide itself if it can process it, rather than `Glia` determining if call visualizer needs to process the event?

yurii-glia (approved):

igorkravchenko (approved):
Taking into account points from Gerson I approve.

GliaWidgets/Public/Glia/Glia.swift
rasmustautsglia: That event is exclusive for call visualizer, so Glia must determine it. It is possible however to have an event listener specifically for this event in call visualizer, but since all events would be published to both listeners, I combined them

GliaWidgets/Sources/CallVisualizer/Coordinator/CallVisualizer.Coordinator.swift
rasmustautsglia: fixed [3461dd9](https://github.com/salemove/ios-sdk-widgets/pull/538/commits/3461dd9de31bdc991a93f12e25092d53f83d2fc6)

GliaWidgets/Public/Glia/Glia.swift
```diff
@@ -254,6 +254,10 @@ public class Glia {
 private extension Glia {
     func startObservingInteractorEvents() {
         interactor?.addObserver(self) { [weak self] event in
+            /// Since engagement is accepted before current engagement is established, this case needs to be checked before any other so that visitor code alert can be dismissed properly and visitor code embedded view can trigger callback
```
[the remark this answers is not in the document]
rasmustautsglia: fixed [3461dd9](https://github.com/salemove/ios-sdk-widgets/pull/538/commits/3461dd9de31bdc991a93f12e25092d53f83d2fc6)

GliaWidgets/Public/Glia/Glia.swift
gersonnoboa: Yes, Glia listens for the events, that is fine. What I meant was something like

```
self?.callVisualizer.handleEventOrSomeSimilarName(event)
```

and then inside callVisualizer

```
fund handleEvent(event: Event) {
    if case .engagementRequestAccepted = event {
        handleEngagementRequestAccepted()
    }
}
```

that way the only entity that knows about this is callVisualizer, which is the only one that should be interested about the fact that we need to do some handling of this event. No knowledge then is spilled into `Glia.swift`.

gersonnoboa: Discussed privately with Rasmus and doing it like this would unfortunately make it inconsistent with the rest of the code so it can just stay as it is. It creates coupling of this code with call visualizer but it is what it is.

gersonnoboa (approved):

GliaWidgets/Public/Glia/Glia.swift
EgorovEI: It's reasonable, but can be a bit confusing in future, because `CallVisualizer` has its own subscription on interactor events and handling these events.

EgorovEI (approved):

rasmustautsglia:
!squash

rasmustautsglia:
!merge

PR #539, opened 2023-03-13
Remove the back button on secure conversations chat transcript

gersonnoboa:
The current logic correctly identifies when to show the secure conversations title or the chat title. This logic is encapsulated into its own function so that it can be used to determine back button behavior. The whole props for the header are regenerated every time the new renderProps method (previously renderTitle) is called, so that it correctly identifies the different chat models and its actions.

MOB-1888

### DIFF
--- a/GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
+++ b/GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
@@ -153,36 +153,6 @@ class ChatViewModel: EngagementViewModel, ViewModel {
     }
     // swiftlint:enable function_body_length
 
-    func event(_ event: Event) {
-        switch event {
-        case .viewDidLoad:
-            start()
-            isViewLoaded = true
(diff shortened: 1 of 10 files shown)


GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -296,6 +266,38 @@ class ChatViewModel: EngagementViewModel, ViewModel {
     }
 }
 
+extension ChatViewModel {
```
gersonnoboa, line 269: This change is only because SwiftLint was complaining

igorkravchenko (approved):
Looks good to me.

EgorovEI (approved):

yurii-glia (approved):

gersonnoboa:
!merge

gersonnoboa:
!squash

gersonnoboa:
!merge

PR #543, opened 2023-03-15
CI fixes

yurii-glia:
- Update accessibility identifiers for Header component
- Disable push notifications by default. BrowserStack doesn't work well with push notifications and makes unnecessary delay during acceptance tests run
- Improve Bitrise workflow for sending notification if workflow is failed


bitrise.yml
```diff
@@ -181,10 +181,15 @@ workflows:
         inputs:
         - notify_email_list: $NOTIFY_EMAILS
     - slack@3:
-        is_always_run: false
         inputs:
         - channel: '#tm-mobile-builds'
         - text: iOS Build Succeeded!
+        - webhook_url_on_error: $SLACK_IOS_WEBHOOK
+        - channel_on_error: '#tm-mobile'
+        - text_on_error: iOS Build Failed!
```
gersonnoboa, line 189: I would suggest to add `@mobile-caretaker`

GliaWidgets/SecureConversations/Confirmation/SecureConversations.ConfirmationViewModel.swift
```diff
@@ -58,7 +58,7 @@ extension SecureConversations.ConfirmationViewModel {
         return Header.Props(
             title: style.headerTitle,
             effect: .none,
-            endButton: .init(style: style.header.endButton),
+            endButton: .init(style: style.header.endButton, accessibilityIdentifier: "header_end_button"),
```
gersonnoboa, line 61: Do we have a file with constants? This is used so much that it would be cool to consolidate it in one place.

bitrise.yml
yurii-glia: fixed - 437446f765dc0d2068b01c7b3429ebe9c1f5fa98

GliaWidgets/SecureConversations/Confirmation/SecureConversations.ConfirmationViewModel.swift
yurii-glia: we don't, however, we can think about it to combine all accessibility identifiers together and use them. Would you like me to do it in this PR?

gersonnoboa: Maybe a ticket can be created for cooldown

gersonnoboa (approved):

EgorovEI (approved):

bitrise.yml
```diff
@@ -181,10 +181,15 @@ workflows:
         inputs:
         - notify_email_list: $NOTIFY_EMAILS
     - slack@3:
-        is_always_run: false
         inputs:
         - channel: '#tm-mobile-builds'
         - text: iOS Build Succeeded!
+        - webhook_url_on_error: $SLACK_IOS_WEBHOOK
+        - channel_on_error: '#tm-mobile'
```
EgorovEI, line 188: Maybe we will send notifications about fails to #tm-mobile-builds as we do with successful runs?

gersonnoboa: I think the channel is unimportant because of the fact that the caretaker is going to be tagged. However, personally I would prefer if the errors are indeed sent to tm-mobile so that everyone is aware. tm-mobile-builds has so much noise that the error might get lost if you are not the caretaker.

yurii-glia:
!fixup

yurii-glia:
!squash

PR #544, opened 2023-03-16
Temporarily disable Browserstack uploads

gersonnoboa:
Acceptance tests need to be fixed first before uploading new builds to Browserstack, so they are temporarily disabled. These need to be enabled again once acceptance tests are fixed.


yurii-glia (commented):
do we still need this changes?

gersonnoboa:
Maybe not, if the acceptance tests succeed with https://github.com/salemove/ios-sdk-widgets/pull/543

gersonnoboa:
Said PR succeeded, closing this as it's unnecessary.

PR #546, opened 2023-03-16
Hide input for transcript when message center is unavailable

igorkravchenko:
In case of message center is not available, visitor should not be able to interact with text input. This is achieved by hiding chat entry area for unavailable message center.

MOB-1882

### DIFF
--- a/GliaWidgets/Sources/ViewController/Chat/ChatViewController.swift
+++ b/GliaWidgets/Sources/ViewController/Chat/ChatViewController.swift
@@ -227,8 +227,9 @@ class ChatViewController: EngagementViewController, MediaUpgradePresenter,
         switch type {
         case .chat:
             chatView.props = .init(header: props.chat)
-        case .secureTranscript:
+        case let .secureTranscript(needsTextInput):
             chatView.props = .init(header: props.secureTranscript)
+            chatView.messageEntryView.isHidden = !needsTextInput
         }
     }
 }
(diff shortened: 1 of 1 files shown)


EgorovEI (approved):

GliaWidgets/Sources/ViewController/Chat/ChatViewController.swift
```diff
@@ -227,8 +227,9 @@ class ChatViewController: EngagementViewController, MediaUpgradePresenter,
         switch type {
         case .chat:
             chatView.props = .init(header: props.chat)
-        case .secureTranscript:
+        case let .secureTranscript(allowsInput):
             chatView.props = .init(header: props.secureTranscript)
+            chatView.messageEntryView.isHidden = !allowsInput
```
gersonnoboa, line 232: What happens with the messages when the message entry view is hidden? Do texts get hidden behind the unavailable alert?

GliaWidgets/Sources/ViewController/Chat/ChatViewController.swift
```diff
@@ -262,12 +263,12 @@ extension ChatViewController {
...
                 ? .chat
-                : .secureTranscript
+                : .secureTranscript(allowsInput: true)
             } else {
                 return .chat
             }
-        case .transcript:
-            return .secureTranscript
+        case .transcript(let transcriptModel):
+            return .secureTranscript(allowsInput: transcriptModel.isSecureConversationsAvailable)
```
gersonnoboa, line 271: Wouldnt it be slighty more "Swifty" if this is called `needsInput` rather?

igorkravchenko: needsTextInput d1fc0441d3c41af49d93b72b4227be56ba84fc03.

GliaWidgets/Sources/ViewController/Chat/ChatViewController.swift
igorkravchenko: Yes, like so:

gersonnoboa (approved):

igorkravchenko:
!squash

igorkravchenko:
!merge

PR #550, opened 2023-03-20
Keep display on during audio and video calls

rasmustautsglia:
Currently during audio or video call the screen would turn off after some time if there is no interaction with the screen. This PR fixes this issue and keeps the screen on until the end of the call. In addition, if user decides to hold the phone to their ear, the proximity sensor is activated, which will dim the screen and disable interaction until phone has been removed from close contact.

MOB-1661


gersonnoboa (commented):
I am wondering if this should be done on the widgets or on the core SDK? Just like for example the green line in screen sharing is handled by the core SDK. What do others think?

yurii-glia (commented):
I'd like to propose to make separate `ProximityManaging` structure and move implementation there, so we will have possibility to reuse this code where we would like easily

igorkravchenko (changes requested):
UIDevice, UIApplication (also NotificationCenter with UIScreen) are dependencies and have to be managed, otherwise we will not be able to test entities that use them.

igorkravchenko:
> I'd like to propose to make separate `ProximityManaging` structure and move implementation there, so we will have possibility to reuse this code where we would like easily

@yurii-glia @rasmustautsglia  I would suggest to have UIKitBased.UIDevice instead of  `ProximityManaging`.

GliaWidgets/Sources/CallVisualizer/Coordinator/CallVisualizer.Coordinator.swift
gersonnoboa, line 1: This fine now has a SwiftLint warning about body length

GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
```diff
@@ -9,7 +10,7 @@ class CallViewModel: EngagementViewModel, ViewModel {
     private let call: Call
     private let startWith: StartAction
     private let durationCounter: CallDurationCounter
-    private let unreadMessages: ObservableValue<Int>
+    private let unreadMessages: ObservableValue<Int>
```
gersonnoboa, line 13: SwiftLint issue

GliaWidgets/Sources/CallVisualizer/VideoCall/ViewController/VideoCallViewController.swift
```diff
@@ -32,6 +35,25 @@ public extension CallVisualizer {
...
+                    uiScreen: environment.uiScreen,
+                    notificationCenter: environment.notificationCenter
+                )
+            )
+            proximityManager?.start()
+        }
+
+        public override func viewWillDisappear(_ animated: Bool) {
+            super.viewWillDisappear(animated)
+            proximityManager?.stop()
```
gersonnoboa, line 55: Do you think that there could be a situation where the view disappears and it appears again? Because `viewWillLoad` and `viewWillDisappear` are not equivalent. I don't know if it could go on `deinit` instead, or where could be a good place to put it. Someone help me :)

GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
rasmustautsglia: fixed [48c489d](https://github.com/salemove/ios-sdk-widgets/pull/550/commits/48c489d3f16381a55eb7b2085d91fc877f609265)

GliaWidgets/Sources/CallVisualizer/Coordinator/CallVisualizer.Coordinator.swift
rasmustautsglia: fixed [a740841](https://github.com/salemove/ios-sdk-widgets/pull/550/commits/a740841ed2fdefdc7820eee1615c226f67fbdb13)

GliaWidgets/Sources/CallVisualizer/VideoCall/ViewController/VideoCallViewController.swift
rasmustautsglia: Yes good point in the context of bubble appearing. The call is still on when bubble is present so it makes sense to move it to deinit

rasmustautsglia: fixed [648be6f](https://github.com/salemove/ios-sdk-widgets/pull/550/commits/648be6f90a46a7c5ccadf00f0514b099a48f6c73)

igorkravchenko (commented):
Looks good. I think you need to address compilation issues of unit test target by adding missing managed dependencies you have added, probably adding `Failing` counterparts.

yurii-glia (approved):

gersonnoboa (approved):
I know that yesterday you worked to address the issues with failing tests with Igor, so I will approve.

rasmustautsglia:
!squash

rasmustautsglia:
!merge

igorkravchenko (approved):
Looks good to me 👍

rasmustautsglia:
!merge

PR #551, opened 2023-03-20
Move parts of reusable code into shared ChatWithTranscriptModel

igorkravchenko:
Place parts of reusable code into shared `ChatWithTranscriptModel` as static methods with more parameters. Note that while we reuse code more, we also may be in situation where it would make harder to understand these shared parts code and unintentionally break it.

MOB-1892


yurii-glia (approved):

gersonnoboa (approved):

igorkravchenko:
!merge

PR #553, opened 2023-03-20
Add snapshot tests for secure connversation welcome screen

yurii-glia:


igorkravchenko (commented):
Left some comments. The main [point](https://github.com/salemove/ios-sdk-widgets/pull/553#discussion_r1144229954) that clean up of things can introduce subtle issues, not being seen at first sight. I think we should first focus on adding snapshots and address consistency improvements separately in other PR. The only exception is when it is not possible to add snapshots without some of these changes.
Again by all means the code needs those improvements but it should be done apart from the main goal of this PR.

GliaWidgets/SecureConversations/FileUploadListView/SecureConversations.FilePreviewView.swift
```diff
@@ -5,7 +5,7 @@
 import UIKit
 
 extension SecureConversations {
-    final class FilePreviewView: UIView {
+    final class FilePreviewView: BaseView {
```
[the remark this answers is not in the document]
yurii-glia: I did it because previously adding subviews and establishing contains happened in the initializer is not the fully correct way, as a result of it SnapshotTest generated a white screen instead of all layout. I thought about moving establishing constraints into updateConstratint methods, however, decided to add consistency also and:
- changed base class to BaseView
- moved addSubview to setup
- moved constraints stuff to defineLayout
- `makeView` sets translatesAutoresizingMaskIntoConstraints to false, however, it doesn't required if it placed in stack view.

If you think these changes are not necessary for now and better to fix issue only without additional changes - let me know and I will revert changes back.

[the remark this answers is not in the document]
igorkravchenko: Got it, so partially it was necessary for making snapshots work. Thanks for explanation. I don't mind these changes. We have FileUpload related stuff covered in chat screen so probably it is safe to proceed as is.

gersonnoboa (commented):
I am sorry but there's too many changes in this PR for it not to have a commit description. Why were the `addSubviews` moved, how come the setup is now done in `defineLayout`, why there is no way to do the text view thing without `mainQueue.async` (this is a big one imo).

gersonnoboa:
Okay I see this comment https://github.com/salemove/ios-sdk-widgets/pull/553#discussion_r1145170318 but it's hidden under resolved changes. Those explanations need to be in the commits for the future, and still I don't know if all changes made here were done because the snapshot appeared empty. For example the text view use of `mainQueue.async` seems like something different. Maybe this PR could have benefited from 2-3 commits, the couple of first ones fixing issues with the snapshots, and the last one with the code for the snapshots themselves.

igorkravchenko (approved):

yurii-glia:
> Okay I see this comment [#553 (comment)](https://github.com/salemove/ios-sdk-widgets/pull/553#discussion_r1145170318) but it's hidden under resolved changes. Those explanations need to be in the commits for the future, and still I don't know if all changes made here were done because the snapshot appeared empty. For example the text view use of `mainQueue.async` seems like something different. Maybe this PR could have benefited from 2-3 commits, the couple of first ones fixing issues with the snapshots, and the last one with the code for the snapshots themselves.

@gersonnoboa updated and separated for 3 commits.

EgorovEI (approved):

SnapshotTests/SecureConversationsWelcomeScreenTests.swift
```diff
@@ -0,0 +1,125 @@
...
+        )
+    }
+
+    static func makeWelcomeProps(
+        theme: SecureConversations.WelcomeStyle,
+        headerProps: Header.Props = headerProps(),
+        uploads: [SecureConversations.FileUploadView.Props] = [],
+        warningMessage: String = ""
+    ) -> SecureConversations.WelcomeView.Props {
+
```
gersonnoboa, line 101: Unneeded space

SnapshotTests/SecureConversationsWelcomeScreenTests.swift
```diff
@@ -0,0 +1,125 @@
+import AccessibilitySnapshot
+@testable import GliaWidgets
+import SnapshotTesting
+import XCTest
+
+class SecureConversationsWelcomeScreenTests: SnapshotTestCase {
+
```
gersonnoboa, line 7: Unneeded space

SnapshotTests/SecureConversationsConfirmationScreenTests.swift
```diff
@@ -0,0 +1,48 @@
+import AccessibilitySnapshot
+@testable import GliaWidgets
+import SnapshotTesting
+import XCTest
+
+class SecureConversationsConfirmationScreenTests: SnapshotTestCase {
+
```
gersonnoboa, line 7: Unneeded space

gersonnoboa (approved):

PR #554, opened 2023-03-21
Add unit tests for IdCollection

igorkravchenko:
Cover IdCollection with unit tests. This custom collection is used for list based UI like file upload list for Props diffing.

MOB-1841


yurii-glia (commented):
should we handle use case when class user will try to add two items with the same id?

EgorovEI (approved):

igorkravchenko:
> should we handle use case when class user will try to add two items with the same id?

Good catch! Adding test for _appending_ item with same identifier caused crash. Fixed 4bfe13d8ea40ad1cc1f2e356104d8a1f0b96454b

rasmustautsglia (approved):

yurii-glia (approved):

igorkravchenko:
!merge

PR #555, opened 2023-03-21
Fix dismissing screen sharing screen

EgorovEI:
There was an issue when VideoCall screen was presented over ScreenSharing screen and user press "back" button, ScreenSharing screen was not dismissed along with VideoCall. 
This PR fixes it. 

MOB-1925


yurii-glia (approved):
I think it is ok as a hotfix, however, we should think about improve handling for presented viewcontrollers.

igorkravchenko (approved):

EgorovEI:
!squash

PR #556, opened 2023-03-22
Add accessibility identifier for header

gersonnoboa:
This will be used in acceptance tests to check if the header has changed from messaging to chat in secure conversations. As an added benefit, it can be used whenever checking the header is needed, and it will have the identifier `header_` and the title in lowercase appended.

MOB-1752

### DIFF
--- a/GliaWidgets/Sources/Component/Header/Header.swift
+++ b/GliaWidgets/Sources/Component/Header/Header.swift
@@ -70,6 +70,7 @@ final class Header: BaseView {
 
         titleLabel.text = props.title
         titleLabel.accessibilityLabel = props.title
+        titleLabel.accessibilityIdentifier = "header_view_title_label"
 
         effectView.isHidden = props.effect == .none
         titleLabel.font = props.style.titleFont


igorkravchenko (commented):
Does it make sense to rely on header text or accessibility label instead? Identifier seems to me like a permanent thing that should not be changed.

gersonnoboa:
Is there a way to get an element by accessibility label and not by accessibility identifier? I did it like that because from [here](https://appium.io/docs/en/commands/element/find-elements/) it doesn't look like I can use something else. If I recall correctly, the "Configure SDK" button is working like this, where the identifier is changed to have a "_loading" suffix when it is loading, and then it returns to the version without suffix once it is completed. However if there is a better way to do it, I can definitely change it.

igorkravchenko:
> Is there a way to get an element by accessibility label and not by accessibility identifier? I did it like that because from [here](https://appium.io/docs/en/commands/element/find-elements/) it doesn't look like I can use something else. If I recall correctly, the "Configure SDK" button is working like this, where the identifier is changed to have a "_loading" suffix when it is loading, and then it returns to the version without suffix once it is completed. However if there is a better way to do it, I can definitely change it.

Maybe something like [this](https://github.com/salemove/acceptance_tests/blob/32fd980343ae7b6b6f7fccdb7eddcb8a5e1dfe56/lib/features/engagement/session/mobile_visitor_session/ios_widget_visitor_session.rb#L137)?
But in case it is not straightforward, I do not insist on finding different solution, so having it as you implemented is also fine.

rasmustautsglia (approved):

gersonnoboa:
@igorkravchenko I have followed your suggestion and have just created a generic identifier for the label. Then on acceptance tests, I have created a [fixup commit](https://github.com/salemove/acceptance_tests/pull/3103/commits/1071f4cafe97e31c5a62d21ce6e0102a07ea4d62) that uses this identifier and checks the label's text.

yurii-glia (approved):
some time ago, I read in docs that using accessibilityIdentifier is the fastest way to get access to components with Appium.

igorkravchenko (approved):

gersonnoboa:
!merge

PR #561, opened 2023-03-24
Update core-sdk dependency version

EgorovEI:


gersonnoboa (approved):
Sorry that you had to do this manually, the release process got broken when we stopped using glia-ios-podspecs, as that repo was the trigger for updating core-sdk in the widgets. Will fix it once the secure conversation project ends.

igorkravchenko (approved):

PR #562, opened 2023-03-24
Reworking ScreenShareHandler

EgorovEI:
`ScreenShareHandler` has been reworked to provide ability to use it in testing purposes
Added tests for `ScreenShareHandler`


GliaWidgets/Sources/Screensharing/ScreenShareHandler.Interface.swift
```diff
@@ -0,0 +1,11 @@
+enum ScreenSharingStatus {
+    case started
+    case stopped
+}
+
+struct ScreenShareHandler {
+    var status: () -> ObservableValue<ScreenSharingStatus>
+    var updateState: (CoreSdkClient.VisitorScreenSharingState?) -> Void
+    var stop: ((() -> Void)?) -> Void
```
igorkravchenko, line 9: I think readability can be improved by introducing type alias. Something like `Callback`.  So that it could be:
```
var stop: (Callback?) -> Void
```

GliaWidgets/Sources/Screensharing/ScreenShareHandler.Live.swift
```diff
@@ -0,0 +1,35 @@
...
+                case .sharing:
+                    status.value = .started
+                case .notSharing:
+                    status.value = .stopped
+                @unknown default:
+                    break
+                }
+            },
+            stop: { completion in
+                visitorState?.localScreen?.stopSharing()
```
igorkravchenko, line 26: This part contains some important business logic. like  calling some methods and `nil`-ing out state etc. Taking into account that we do not use `live` implementation it unit tests we will miss out opportunity to write tests for this logic. For me it seems like this part is totally testable as is, so maybe it should not be called live. We can discuss it separately.

GliaWidgetsTests/CallVisualizer/VideoCall/VideoCallTests.swift
```diff
@@ -12,21 +12,27 @@ final class VideoCallTests: XCTestCase {
     }
 
     func test_end_screen_sharing() {
-        let viewModel: CallVisualizer.VideoCallViewModel = .mock()
-        viewModel.environment.screenShareHandler.status.value = .started
+        let screenShareHandlerMock = ScreenShareHandler.live
```
igorkravchenko, line 15: As you can see you have to use `live` is tests, and this uses singleton in tests, even though `ScreenShareHandler` is a struct but Observables there are classes. Meaning that they will be affected by other tests. This is what is mentioning earlier.

igorkravchenko (commented):
Left come comments about using it as live singleton.

GliaWidgets/Sources/Screensharing/ScreenShareHandler.Interface.swift
EgorovEI: Fixed c17663b6a91b022e075a6c811e0fefa5a9ae7420

GliaWidgets/Sources/Screensharing/ScreenShareHandler.Live.swift
EgorovEI: - added tests for `ScreenShareHandler`
- renamed Live with Implementation
- added `create()` static method
- replaced `.live` instance in tests with `.mock`
c17663b6a91b022e075a6c811e0fefa5a9ae7420

GliaWidgetsTests/CallVisualizer/VideoCall/VideoCallTests.swift
EgorovEI: c17663b6a91b022e075a6c811e0fefa5a9ae7420

GliaWidgets/Sources/Screensharing/ScreenShareHandler.Interface.swift
```diff
@@ -0,0 +1,13 @@
+enum ScreenSharingStatus {
+    case started
+    case stopped
+}
+
+struct ScreenShareHandler {
+    typealias CallBack = () -> Void
```
yurii-glia, line 7:
```suggestion
    typealias Callback = () -> Void
```

yurii-glia (approved):
Looks good, only one minor suggestion about terminology.

GliaWidgets/Sources/Screensharing/ScreenShareHandler.Interface.swift
EgorovEI: bfc1ee04edcb8d119f1e4cf03e6d05c487d06577

igorkravchenko (approved):
Looks good 👍

EgorovEI:
!squash

GliaWidgetsTests/ScreenShareHandler/ScreenShareHandlerTests.swift
```diff
@@ -0,0 +1,51 @@
+import XCTest
+@testable import GliaWidgets
+import SalemoveSDK
+
+final class ScreenShareHandlerTests: XCTestCase {
+
```
gersonnoboa, line 6: Unneeded space

GliaWidgets/Sources/Screensharing/ScreenShareHandler.Implementation.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+
+extension ScreenShareHandler {
+    static func create() -> ScreenShareHandler {
```
gersonnoboa, line 4: I think normally these are being called `live` rather than `create`. Especially as the mock is called `mock`. Also is there a way this is a variable rather than a function? Like the mock https://github.com/salemove/ios-sdk-widgets/pull/562/files#diff-a06be315abeb7eff822e44608e9f208a6d718246da45d7b116beb7a563621d2dR5.

GliaWidgets/Sources/CallVisualizer/VideoCall/ViewModel/VideoCallViewModel.swift
```diff
@@ -174,7 +174,7 @@ extension CallVisualizer.VideoCallViewModel {
                     closeButton: .init(style: style.header.closeButton),
                     endScreenshareButton: .init(
                         tap: .init { [weak self] in
-                            self?.environment.screenShareHandler.stop()
+                            self?.environment.screenShareHandler.stop(nil)
```
gersonnoboa, line 177: I believe the stop parameter could be a little bit more descriptive. When you are using it by sending a callback, you are using the `stop { }` notation anyways so it doesn't matter what the name is. But in here, you just read `stop(nil)` and don't know what the `nil` is referring to.

GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
```diff
@@ -222,11 +222,11 @@ class EngagementViewModel: CommonEngagementModel {
     }
 
     func updateScreenSharingState(to state: CoreSdkClient.VisitorScreenSharingState) {
-        screenShareHandler.updateState(to: state)
+        screenShareHandler.updateState(state)
```
gersonnoboa, line 225: Why was the `to` removed? That way it was clear that the state was updated to `state`. Without it, you just read `updateState(state)` and it's not 100% clear what happens after updating state. Does `state` itself gets updated? Or the internal screenShareHandler state?

gersonnoboa:
Damn I commented after it was merged...

GliaWidgets/Sources/Screensharing/ScreenShareHandler.Implementation.swift
EgorovEI: We had discussion with @igorkravchenko about that. 
@igorkravchenko could you share our agreement about that?

GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
EgorovEI: I removed `to` due to compiler error, because now `updateState` is `var`, not method

PR #569, opened 2023-03-28
Fix set of issues related to upgrade from secure transcript to chat

igorkravchenko:
These changes address the following issues:
   * Sending a message from welcome screen and navigating to chat transcript doesn't show operator messages (MOB-2002)
   * Queue ticket gets cancelled when sending a message from secure conversations chat transcript (MOB-1990)
   * Incorrect message positioning when sending multiple messages before starting engagement (MOB-2018)
   * Secure messaging flow does not work in case of restored engagement

The key point is to evaluate `InteractorState` and based on that do enqueueing, upgrading etc.

MOB-2002

### DIFF
--- a/GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
+++ b/GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
@@ -501,10 +501,13 @@ extension ChatViewModel {
 
             pendingMessages.append(pendingMessage)
 
-            let queueItem = ChatItem(kind: .queueOperator)
+            // Avoid showing enqueud operator according to parameter.
+            if !shouldSkipEnqueueingState {
+                let queueItem = ChatItem(kind: .queueOperator)
+                queueOperatorSection.set([queueItem])
+                action?(.refreshSection(queueOperatorSection.index))
+            }
 
-            queueOperatorSection.set([queueItem])
-            action?(.refreshSection(2))
             action?(.scrollToBottom(animated: true))
         }
     }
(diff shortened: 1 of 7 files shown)


gersonnoboa:
There is a case that doesn't work still, which is one of the things I also struggled and I believe enters under MOB-1990.

Steps to reproduce:
1. Authenticate.
2. Go to secure conversations welcome screen, but press "Check messages". Don't send any messages from welcome screen.
3. Send two messages.

Both MOB-1990 and MOB-2018 happen. However, MOB-2002 is fixed indeed.

yurii-glia (approved):

igorkravchenko:
> There is a case that doesn't work still, which is one of the things I also struggled and I believe enters under MOB-1990.
> 
> Steps to reproduce:
> 
> 1. Authenticate.
> 2. Go to secure conversations welcome screen, but press "Check messages". Don't send any messages from welcome screen.
> 3. Send two messages.
> 
> Both MOB-1990 and MOB-2018 happen. However, MOB-2002 is fixed indeed.

I've changed how we do the upgrade: instead of enqueueing after sending message, just subscribe to `engaged` event. This seem to resolve mentioned issues 7dc931ecf1b8b317cce11aec4ff0cd370a413119.

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -501,10 +501,13 @@ extension ChatViewModel {
 
             pendingMessages.append(pendingMessage)
 
-            let queueItem = ChatItem(kind: .queueOperator)
+            // Avoid showing enqueud operator according to parameter.
+            if !shouldSkipEnqueueingState {
+                let queueItem = ChatItem(kind: .queueOperator)
+                queueOperatorSection.set([queueItem])
+                action?(.refreshSection(2))
```
gersonnoboa, line 508: Weird to see the number 2 just like that

gersonnoboa (approved):

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
igorkravchenko: That how it was initially before I wrapped it in `if !shouldSkipEnqueueingState` condition. Updated ced3a40255b8defb13022647300adf1e3cdae6a2.

EgorovEI (approved):

igorkravchenko:
!merge

PR #570, opened 2023-03-31
Fix incorrect message when opening SC while unauthenticated

gersonnoboa:
Everything was correct, but the configuration sent to the alert was the same as the one that is shown when secure conversations are unavailable because the queue doesn't support it.

MOB-2032


igorkravchenko (approved):

EgorovEI (approved):

gersonnoboa:
!merge

PR #574, opened 2023-04-04
Add accessibility identifier to unread count divider

igorkravchenko:
Add accessibility identifier to unread count divider and make it accessible to VoiceOver.

MOB-2000

### DIFF
--- a/GliaWidgets/Sources/View/Chat/UnreadMessageDividerView.swift
+++ b/GliaWidgets/Sources/View/Chat/UnreadMessageDividerView.swift
@@ -2,6 +2,7 @@ import Foundation
 import UIKit
 
 final class UnreadMessageDividerView: BaseView {
+    static let accessibilityIdentifier = "chatTranscript_unreadMessageDivider_view"
     private let style: UnreadMessageDividerStyle
     private let textLabel = UILabel().makeView()
     private lazy var leadingLine = LineView(color: style.lineColor).makeView()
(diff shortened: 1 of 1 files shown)


GliaWidgets/Sources/View/Chat/UnreadMessageDividerView.swift
```diff
@@ -2,6 +2,7 @@ import Foundation
 import UIKit
 
 final class UnreadMessageDividerView: BaseView {
+    static let accessibilityIdentifier = "unread_message_divider_accessibility_identifier"
```
gersonnoboa, line 5: For the previous accessibility identifiers, we would use something like functionality_description_typeofview. So for example here it could have been chatTranscript_unreadMessageDivider_view. Other parts have it like chat_transcript_unread_message_divider_view. Kind of sucks that we don't have a standard, but I believe that we should at least have the format I mentioned.

igorkravchenko: Good point d9809248b6e12bb627f0ed19f522d238a362033d.

gersonnoboa (approved):

EgorovEI (approved):

rasmustautsglia (approved):

igorkravchenko:
!fixup

igorkravchenko:
!merge

PR #577, opened 2023-04-11
Cover secure transcript with unit test

igorkravchenko:
Add partial coverage of transcript model with unit tests. More tests will be added later in other PR(s)

MOB-2062

### DIFF
--- a/GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
+++ b/GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
@@ -301,7 +301,7 @@ extension SecureConversations {
         private let downloader: FileDownloader
         let fileUploadListModel: SecureConversations.FileUploadListViewModel
 
-        private var messageText = "" {
+        private (set) var messageText = "" {
             didSet {
                 validateMessage()
                 action?(.setMessageText(messageText))
@@ -314,12 +314,7 @@ extension SecureConversations {
 
         var environment: Environment
         var availability: Availability
-        var isSecureConversationsAvailable: Bool = true {
-            didSet {
-                // Hide text field in MOB-1882
-                print("availability has changed")
-            }
-        }
+        private (set) var isSecureConversationsAvailable: Bool = true
 
         var siteConfiguration: CoreSdkClient.Site?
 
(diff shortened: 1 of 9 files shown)


GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
```diff
@@ -303,7 +303,7 @@ extension SecureConversations {
         private let downloader: FileDownloader
         let fileUploadListModel: SecureConversations.FileUploadListViewModel
 
-        private var messageText = "" {
+        private (set) var messageText = "" {
```
gersonnoboa, line 306: I wonder why the whole messageText was private in the first place when in the welcome view model it is public. I am using the "publicness" of the messageText to test that whenever you change it, the view model correctly triggers the render props.

GliaWidgets/Sources/CoreSDKClient/CoreSDKClient.Mock.swift
```diff
@@ -367,4 +367,96 @@ extension CoreSdkClient {
     }
 }
 
+extension CoreSdkClient.QueueState {
```
gersonnoboa, line 370: I think this will not be needed anymore because now the Queue init will be public.

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
igorkravchenko: I do not allow changing `messageText` externally here, but only reading from it for unit tests, so it instead of being fully inaccessible for test case it is now read-only.

gersonnoboa (approved):

GliaWidgetsTests/SecureConversations/TranscriptModelTests.swift
```diff
@@ -0,0 +1,151 @@
...
+            environment: modelEnv,
+            availability: .init(
+                environment: availabilityEnv
+            ),
+            deliveredStatusText: ""
+        )
+        viewModel.start()
+        XCTAssertEqual(
+            calls.map(\.rawValue),
+            [Call.loadSiteConfiguration, .getSecureUnreadMessageCount, .fetchChatHistory].map(\.rawValue)
```
EgorovEI, line 119: Why do you need to compare rawValue's rather then just cases?

EgorovEI: Oh got it. Useful enough

igorkravchenko:
!squash

igorkravchenko:
!merge

PR #578, opened 2023-04-11
Add unit tests for some methods of WelcomeViewModel

gersonnoboa:
The methods for sending messages, delegate calls, attachment availability, and reporting changes for prop rendering have been tested. More tests will be added in subsequent PRs. There's still some coverage to go.

MOB-2056


GliaWidgetsTests/SecureConversations/Welcome/SecureConversations.WelcomeViewModelSpec.swift
```diff
@@ -0,0 +1,214 @@
...
+    }
+
+}
+
+// Send message
+extension SecureConversationsWelcomeViewModelTests {
+    func testSuccessfulMessageSend() {
+        let expectation = expectation(description: "")
+        var isCalled = false
+        self.viewModel.environment.sendSecureMessage = { _, _, _, completion in
```
EgorovEI, line 118: In each test above you initialize viewModel, but here you don't, so changing viewModel in previous test might affect  it in following tests. Look like would be better to initialize viewModel in each test or in setup/tearDown method. 
What do you think?

gersonnoboa: You are correct. This already bit me in one of the tests I am writing now. Will add a `setUp` override.

EgorovEI (approved):

GliaWidgetsTests/SecureConversations/Welcome/SecureConversations.WelcomeViewModelSpec.swift
```diff
@@ -0,0 +1,218 @@
...
+            case .backTapped:
+                isCalled = true
+                expectation.fulfill()
+            default: XCTFail()
+            }
+        }
+
+        viewModel.event(.backTapped)
+
+        waitForExpectations(timeout: 1)
```
igorkravchenko, line 27: I strongly suggest to omit waiting, to avoid flaky tests. In the SDK we have tools that allow accomplishing that, because all of the dependencies are controlled. So instead of waiting, for testing callback based results we use this approach:
```
var receivedValue: Value?

entity.methodWithCallback { value in
   receivedValue = value 
}

try XCTAssertEqual(XCTUnwrap(receivedValue), expectedValue)
```

igorkravchenko (commented):
With a bit of restructuring I think waiting is not necessary for these tests.

GliaWidgetsTests/SecureConversations/Welcome/SecureConversations.WelcomeViewModelSpec.swift
gersonnoboa: Removed all expectations e2b286d5e22d5ad787f913ab27d524c56417aa9e

GliaWidgetsTests/SecureConversations/Welcome/SecureConversations.WelcomeViewModelSpec.swift
```diff
@@ -0,0 +1,196 @@
...
+
+            completion(.success(site))
+        }
+
+        viewModel = .init(environment: environment, availability: .mock)
+
+        XCTAssertTrue(viewModel.isAttachmentsAvailable)
+    }
+
+    func testIsAttachmentNotAvailable() {
```
igorkravchenko, line 62: You can mark you tests to be throwing, and thus avoiding usage of `try!`. This is better because in case of throwing error your whole test suite will not crash, letting other tests to be evaluated, while throwing test will just fail.

```suggestion
    func testIsAttachmentAvailable() throws {
         var environment: SecureConversations.WelcomeViewModel.Environment = .mock
         environment.fetchSiteConfigurations = { completion in
             let site: CoreSdkClient.Site = try .mock(
                 allowedFileSenders: .init(operator: true, visitor: true)
             )

             completion(.success(site))
         }

         viewModel = .init(environment: environment, availability: .mock)

         XCTAssertTrue(viewModel.isAttachmentsAvailable)
     }
```

GliaWidgetsTests/SecureConversations/Welcome/SecureConversations.WelcomeViewModelSpec.swift
```diff
@@ -0,0 +1,196 @@
...
+        var environment: SecureConversations.WelcomeViewModel.Environment = .mock
+        environment.fetchSiteConfigurations = { completion in
+            let site: CoreSdkClient.Site = try! .mock(
```
igorkravchenko, line 65:
```suggestion
            let site: CoreSdkClient.Site = try .mock(
```

igorkravchenko (approved):
One more improvement: you can mark you tests throwing to avoid crashing test suite.

gersonnoboa:
As agreed with Igor, I will remove the `try!` instances on the next PR regarding unit tests, so I can merge this one.

gersonnoboa:
!squash

gersonnoboa:
!merge

PR #581, opened 2023-04-12
Choose download endpoint based on supplied URL

igorkravchenko:
Add ability for file download decide at runtime which endpoint to use based on provided URL. This is needed because secure download endpoint fails for files that were uploaded during engagement.

MOB-2051

### DIFF
--- a/GliaWidgets/Sources/Download/FileDownload.Environment.Interface.swift
+++ b/GliaWidgets/Sources/Download/FileDownload.Environment.Interface.swift
@@ -1,6 +1,7 @@
 extension FileDownload {
     struct Environment {
-        var fetchFile: FetchFile
+        var fetchFile: CoreSdkClient.FetchFile
+        var downloadSecureFile: CoreSdkClient.DownloadSecureFile
         var fileManager: FoundationBased.FileManager
         var gcd: GCD
         var localFileThumbnailQueue: FoundationBased.OperationQueue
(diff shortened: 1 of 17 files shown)


EgorovEI (approved):

GliaWidgets/Sources/Download/FileDownload.Environment.Interface.swift
```diff
@@ -38,4 +39,22 @@ extension FileDownload.Environment.FetchFile {
...
         }
+
+    static func fetchForEngagementFile(
+        _ file: CoreSdkClient.EngagementFile,
+        environment: Environment
+    ) -> FileDownload.Environment.FetchFile {
+        guard let components = file.url
+            .flatMap({ url in URLComponents(url: url, resolvingAgainstBaseURL: false) }),
+            components.path.starts(with: "/messaging/files/") else {
+            return .fromEngagement(environment.fetchFile)
```
gersonnoboa, line 50: Do you think that it is advisable to have a third type, something like unknown, in case the file doesn't have `/messaging/` or `/engagement/`? Or it's not necessary, as the SDK is already kind of working like that anyways at the moment? Just wondering about any slight possibility of us downloading malicious files.

igorkravchenko: I think it is already does quite a lot as of Widgets side. Maybe that case needs to be done on CoreSDK or server side. I'm not sure if any protection for open source code like this can do much.

gersonnoboa (approved):

rasmustautsglia (approved):

igorkravchenko:
!merge

PR #583, opened 2023-04-13
Add tests for remaining methods in WelcomeViewModel

gersonnoboa:
Tried to get to 100% coverage as much as it made sense. Also `try!` has been removed from all tests.

Sorry for the big PR :( I always complain about big PRs and here we are. I was working on this while fixing some stuff on the other open PR about unit tests, and at some point I even thought I had lost changes because of moving around between two different branches on the same file. So I decided to keep all together for this.

MOB-2056


GliaWidgetsTests/SecureConversations/Welcome/SecureConversations.WelcomeViewModelSpec.swift
```diff
@@ -190,7 +215,489 @@ extension SecureConversationsWelcomeViewModelTests {
...
+        XCTAssertEqual(
+            alertConfiguration?.message,
+            viewModel.environment.alertConfiguration.unavailableMessageCenter.message
+        )
+    }
+
+    func testAvailabilityUnavailableUnauthenticated() {
+        var alertConfiguration: MessageAlertConfiguration?
+
+        let uuid = UUID().uuidString
```
igorkravchenko, line 667: N. B. I'd suggest to use UUID.mock to encourage same results in tests. For example if we are to use text-based snapshot test it will be noticeable right away.

igorkravchenko (approved):
Looks good. One thing: instead of `UUID()` it is better to use UUID.mock to have predictable UUID generation.

rasmustautsglia (approved):

GliaWidgetsTests/SecureConversations/Welcome/SecureConversations.WelcomeViewModelSpec.swift
gersonnoboa: Fixed b3af2943f357e79a5759a7110a5d31c334eafaf9

gersonnoboa:
!squash

gersonnoboa:
!merge

PR #588, opened 2023-04-17
Prevent starting engagement during ongoing one

EgorovEI:
Previously we didn't check CallVisualizer engagement existence on `startEngagement` call. This led to unexpected behaviour of SDK. This PR fixes it.

Also added tests for all exceptions that can be thrown on `startEngagement` call.

MOB-2087

### DIFF
--- a/GliaWidgetsTests/Sources/GliaTests.swift
+++ b/GliaWidgetsTests/Sources/GliaTests.swift
@@ -132,6 +132,37 @@ class GliaTests: XCTestCase {
         XCTAssertTrue(try XCTUnwrap(receivedSceneProvider as? MockedSceneProvider) === expectedSceneProvider)
     }
 
+    func testStartEngagementThrowsErrorWhenEngagementAlreadyExists() throws {
+        let sdk = Glia(environment: .failing)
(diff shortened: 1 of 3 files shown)


gersonnoboa (approved):

GliaWidgetsTests/Sources/GliaTests.swift
```diff
@@ -132,6 +132,78 @@ class GliaTests: XCTestCase {
...
+
+        do {
+            try sdk.startEngagement(engagementKind: .chat)
+        } catch GliaError.engagementExists {
+            isErrorThrown = true
+        } catch {
+            XCTFail()
+        }
+
+        XCTAssertTrue(isErrorThrown)
```
igorkravchenko, line 159: N.B. You can also use `XCTAssertThrowsError`.

igorkravchenko (approved):

GliaWidgetsTests/Sources/GliaTests.swift
EgorovEI: Fixed 69dceb129ca16184df4581dec6e62fd7900c1810. Now it looks significantly better. Thank you!

PR #591, opened 2023-04-19
Update bitrise.yml file

EgorovEI:
Removed unused workflows
Updated outdated steps
Added variables

MOB-1685


bitrise.yml
```diff
@@ -40,24 +40,24 @@ workflows:
     steps:
     - activate-ssh-key@4:
         run_if: '{{getenv "SSH_RSA_PRIVATE_KEY" | ne ""}}'
-    - git-clone@6: {}
+    - git-clone@8: {}
```
gersonnoboa, line 43: So it was a conscious decision to use the currently latest version, rather than removing the version and always using the latest?

EgorovEI: Not at all. I made this conclusion based on our discussion https://salemove.slack.com/archives/C01F9K88SQZ/p1681820837845999. 

But I could remove versioning to always use the latest one.
@gersonnoboa @yurii-glia @igorkravchenko @rasmustautsglia wdyt?

gersonnoboa: I don't have a strong preference towards either. Keeping track of those step updates might be too much and then if something breaks, we just would know because the build fails.

igorkravchenko: @EgorovEI could you please provide more info how this change affects workflows?

igorkravchenko (commented):
@EgorovEI could you please broaden the description of this PR. I'm not familiar with Bitrise configuration yet, unfortunately, so it would be nice to have wider description.

bitrise.yml
EgorovEI: Updating major step version does not affect our workflows now, I run them with these changes. 
Removing versioning at all leads to automatic updating to a new major version, which might potentially break the build in future. But since major versions of steps are not released so frequently, and have not led to breaking build before, we might try to remove versioning at all.

igorkravchenko (approved):

gersonnoboa (approved):

PR #593, opened 2023-04-20
Update secure message response to use socket based message

igorkravchenko:
Attachment is missing in secure message response for REST API call when message is sent. We fallback on response from socket because of this.

MOB-2042

### DIFF
--- a/GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
+++ b/GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
@@ -270,7 +270,7 @@ extension SecureConversations {
         typealias DelegateCallback = (DelegateEvent) -> Void
         typealias Action = ChatViewModel.Action
 
-        static let unavailableMessageCenterAlertAccIdentidier = "unavailable_message_center_alert_identifier"
+        static let unavailableMessageCenterAlertAccIdentifier = "unavailable_message_center_alert_identifier"
         static let markUnreadMessagesDelaySeconds = 6
 
         enum DelegateEvent {
@@ -333,6 +333,9 @@ extension SecureConversations {
 
         var historySection: Section<ChatItem> { sections[0] }
         var pendingSection: Section<ChatItem> { sections[1] }
+
+        private (set) var receivedMessages = [String: [MessageSource]]()
+
         let deliveredStatusText: String
 
         var numberOfSections: Int {
@@ -433,7 +436,7 @@ extension SecureConversations {
             self.delegate?(
                 .showAlertAsView(
                     configuration,
-                    accessibilityIdentifier: Self.unavailableMessageCenterAlertAccIdentidier,
+                    accessibilityIdentifier: Self.unavailableMessageCenterAlertAccIdentifier,
                     dismissed: nil
                 )
             )
@@ -505,6 +508,7 @@ extension SecureConversations {
         }
 
         func start() {
+            environment.startSocketObservation()
             fetchSiteConfigurations()
             loadHistory { _ in }
         }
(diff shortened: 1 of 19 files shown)


GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
```diff
@@ -910,6 +913,58 @@ extension SecureConversations.TranscriptModel {
...
+            switch self {
+            case let .api(_, outgoingMessage):
+                return outgoingMessage
+            case  .socket:
+                return nil
+            }
+        }
+    }
+
+    func receiveMessage(_ message: MessageSource) {
```
gersonnoboa, line 945: Maybe `receiveMessage(fromSource source: MessageSource)` as the previous name indicates that we are receiving a message.

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
```diff
@@ -505,6 +508,7 @@ extension SecureConversations {
         }
 
         func start() {
+            environment.startSocketObservation()
```
gersonnoboa, line 511: Where is the observation stopped? [In my PR](https://github.com/salemove/ios-sdk-widgets/pull/592/files#diff-8f58d92d9bfc067e9074401288bdb06350dd64a36e63bd024d89a2e38b116b8bR480) about system messages, I have created an event that gets called on the `deinit` of ChatViewController so that the observation is stopped at that point. If you don't want to deal with that, you can leave it and it would be added by my system messages PR, but we have to decide about it.

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
igorkravchenko: Renamed 3118054ac0dd26ec64a9ea495d2a0b338a2fb2e7.

GliaWidgets/SecureConversations/SecureConversations.ChatWithTranscriptModel.swift
igorkravchenko: I think the observation has to be stopped on upgrade to engagement, because ChatViewModel will replace TranscriptModel.

gersonnoboa (approved):
There are two SwiftLint warnings because the file is too big and `navigateToTranscript` is too big. But I will approve.

igorkravchenko:
> There are two SwiftLint warnings because the file is too big and `navigateToTranscript` is too big. But I will approve.

Addressed linter warnings 3af5e778b943250b00d36b0c1a7b2da00847a739.

GliaWidgets/SecureConversations/TranscriptModel.DividedChatItemsForUnreadCount.swift
```diff
@@ -0,0 +1,47 @@
+import Foundation
+
+extension SecureConversations.TranscriptModel {
+    static func dividedChatItemsForUnreadCount(
```
yurii-glia, line 4: would it make sense to simplify the function name to something "addDividerToChatItems"?

yurii-glia (approved):

GliaWidgets/SecureConversations/TranscriptModel.DividedChatItemsForUnreadCount.swift
igorkravchenko: I moved this part to address linter warning. I would prefer to leave it as it is for now, to avoid adding more changes to this PR including renaming in unit tests (the whole test case).

igorkravchenko:
!squash

igorkravchenko:
!merge

PR #598, opened 2023-04-24
Fix configuring SDK during active engagement

EgorovEI:
Previously SDK allowed to call `configure` method during active engagement, which broke socket connection, so visitor could see operator messages. This commit fixes it by throwing an error about active engagement.

MOB-2113

### DIFF
--- a/GliaWidgets/Public/Glia/Glia+StartEngagement.swift
+++ b/GliaWidgets/Public/Glia/Glia+StartEngagement.swift
@@ -62,7 +62,6 @@ extension Glia {
     ///   - engagementKind: Engagement media type.
     ///   - configuration: Engagement configuration.
     ///   - queueID: Queue identifier.
-    ///   - visitorContext: Visitor context.
     ///   - theme: A custom theme to use with the engagement.
     ///   - sceneProvider: Used to provide `UIWindowScene` to the framework. Defaults to the first active foreground scene.
     ///
@@ -78,27 +77,28 @@ extension Glia {
         _ engagementKind: EngagementKind,
         configuration: Configuration,
         queueID: String,
-        visitorContext: VisitorContext?, // GliaCoreSDK.VisitorContext
         theme: Theme = Theme(),
         features: Features = .all,
         sceneProvider: SceneProvider? = nil
     ) throws {
-        try configure(
-            with: configuration,
-            queueId: queueID,
-            visitorContext: visitorContext
-        ) { [weak self] in
+        let completion = { [weak self] in
             self?.setChatMessageRenderer(messageRenderer: .webRenderer)
-            do {
-                try self?.startEngagement(
-                    engagementKind: engagementKind,
-                    theme: theme,
-                    features: features,
-                    sceneProvider: sceneProvider
-                )
-            } catch {
-                print("Engagement has not been started. Error='\(error)'.")
+            try self?.startEngagement(
+                engagementKind: engagementKind,
+                theme: theme,
+                features: features,
+                sceneProvider: sceneProvider
+            )
+        }
+        do {
+            try configure(
+                with: configuration,
+                queueId: queueID
+            ) {
+                try? completion()
             }
+        } catch GliaError.configuringDuringEngagementIsNotAllowed {
+            try completion()
         }
     }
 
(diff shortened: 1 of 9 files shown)


TestingApp/ViewController/ViewController.swift
```diff
@@ -182,14 +183,22 @@ extension ViewController {
         let originalIdentifier = configureButton.accessibilityIdentifier
         configureButton.setTitle("Configuring ...", for: .normal)
         configureButton.accessibilityIdentifier = "main_configure_sdk_button_loading"
-        try? Glia.sharedInstance.configure(
-            with: configuration,
-            queueId: queueId,
-            uiConfig: uiConfig
-        ) { [weak self] in
+
+        let completed = { [weak self] printable in
```
gersonnoboa, line 187: Maybe `completion` rather than `completed`

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
```diff
@@ -83,22 +83,24 @@ extension Glia {
...
+                features: features,
+                sceneProvider: sceneProvider
+            )
+        }
+        do {
+            try configure(
+                with: configuration,
+                queueId: queueID
+            ) {
+                try? completion()
```
gersonnoboa, line 100: If this completion fails, what happens? I think it would fail silently, no? Maybe it would be good to also notify the integrator about the error in starting an engagement.

rasmustautsglia: I agree with Gerson that we should notify the integrator about an error

EgorovEI: In this case:
- it's not needed, because `startEngagement` throws an error when SDK is not configured (in this case SDK is already configured) or active engagement exists (in this case `configure` will throw `GliaError.configuringDuringEngagementIsNotAllowed` and `try completion()` on 103th line will be executed.
- I have not found any way to throw an error from `configure` completion scope, because this completion closure is not throwable (i.e. not `throws`)

GliaWidgetsTests/Sources/Glia/GliaTests.swift
```diff
@@ -184,6 +151,10 @@ final class GliaTests: XCTestCase {
             queueId: "queueId"
         )
 
+        sdk.environment.coreSdk.getCurrentEngagement = {
+            .init(id: "", engagedOperator: nil, source: .callVisualizer, fetchSurvey: { _, _ in })
```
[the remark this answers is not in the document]
EgorovEI: cbe51712900d18252b30e6cadaa57dddefb12633

TestingApp/ViewController/ViewController.swift
EgorovEI: cbe51712900d18252b30e6cadaa57dddefb12633

gersonnoboa (approved):

igorkravchenko (approved):

rasmustautsglia (approved):
Looks good to me!

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
```diff
@@ -83,22 +83,24 @@ extension Glia {
         features: Features = .all,
         sceneProvider: SceneProvider? = nil
     ) throws {
-        try configure(
-            with: configuration,
-            queueId: queueID,
-            visitorContext: visitorContext
```
yurii-glia, line 89: you have dropped `visitorContext` from the call which is totally correct, however, in the interface the visitorContext still exists. In my opinion, it can make some confusion for an integrator, perhaps we should deprecate this function and throw away `visitorContext` + we can add print, 
```swift
if visitorContext != nil { print('visitorContext has been deprecated, please pass correct visitor context during creation SDk configuration.') }
```
what do you think, does it make sense?
@gersonnoboa  @igorkravchenko @rasmustautsglia

EgorovEI: Makes sense to me

rasmustautsglia: Yes, makes sense to me to deprecate this method and add a new one

EgorovEI: Deprecated current `start` method and add new one without `visitorContext` parameter 80a88f831ce57f7cfddba7cf3733161c14af1f59

yurii-glia (approved):

PR #606, opened 2023-04-28
Fix attachment picker distortion when placed at top of the screen

igorkravchenko:
Address attachment picker popover being incorrectly displayed when presented on top part of the screen. This is done by dynamically determining arrow direction for popover, based on source view for it in global coordinate system. Theses changes address only vertical direction.

MOB-2127


gersonnoboa (approved):
Currently the attachment button is always displayed on the right side of the screen, so if at some point we need to display it on the left, then we can come back and figure out that.

EgorovEI (approved):

yurii-glia (approved):

rasmustautsglia (approved):

igorkravchenko:
!merge

PR #607, opened 2023-05-01
Fix confirmation screen text disappearing in landscape orientation

gersonnoboa:
The confirmation image was too big for landscape orientation, so the text below it would disappear. To fix it, the image size is made smaller, as well as the factor that determines where the top of the image view is.

MOB-2129


yurii-glia (approved):

rasmustautsglia (approved):

gersonnoboa:
!merge

PR #608, opened 2023-05-02
Add unit tests for confirmation view model

gersonnoboa:
Add tests for all actions and deleted unneeded code, such as the code for back button functionality. This code is unreachable because we don't have a back button in confirmation screen. So not only it is unneeded but also I couldn't achieve full coverage of this file otherwise.

### DIFF
--- a/GliaWidgetsTests/SecureConversations/Confirmation/SecureConversations.ConfirmationViewModelTests.swift
+++ b/GliaWidgetsTests/SecureConversations/Confirmation/SecureConversations.ConfirmationViewModelTests.swift
@@ -1,0 +1,105 @@
+import Foundation
+import XCTest
+@testable import GliaWidgets
+
+final class SecureConversationsConfirmationViewModelTests: XCTestCase {
+    typealias ConfirmationViewModel = SecureConversations.ConfirmationViewModel
+    var viewModel: ConfirmationViewModel = .mock
+
+    override func setUp() {
+        viewModel = .mock
+    }
+}
+
+// Props
+extension SecureConversationsConfirmationViewModelTests {
+    func testPropsDoNotGenerateABackButton() {
+        let props = viewModel.props().confirmationViewProps.header
+
+        XCTAssertNil(props.backButton)
+    }
+
+    func testPropsGenerateCorrectTitle() {
+        let title = "Test"
+        var style = Theme().secureConversationsConfirmation
+        style.headerTitle = title
+
+        viewModel.environment = .init(confirmationStyle: style)
+
+        let props = viewModel.props()
+        XCTAssertEqual(props.confirmationViewProps.header.title, title)
+    }
+
+    func testPropsGenerateEndButtonWithAccessibilityIdentifier() {
+        let props = viewModel.props().confirmationViewProps.header.endButton
+
+        XCTAssertEqual(props.accessibilityIdentifier, "header_end_button")
+    }
+}
+
+// Delegate
+extension SecureConversationsConfirmationViewModelTests {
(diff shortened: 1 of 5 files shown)


rasmustautsglia (approved):
Looks good! :)

GliaWidgetsTests/SecureConversations/Confirmation/SecureConversations.ConfirmationViewModelTests.swift
```diff
@@ -0,0 +1,89 @@
...
+    func testSendingCloseButtonEventCallsDelegate() {
+        viewModel.delegate = { event in
+            switch event {
+            case .closeTapped:
+                XCTAssertTrue(true)
+            default: XCTFail()
+            }
+        }
+
+        viewModel.event(.closeTapped)
```
yurii-glia, line 51: should be better to use `XCTestExpectation` here?
I am just worried, if sometime delegate will be async, this test will be succeeded because after line 51 everything in the test will be released, and since after this line no XCTFail Xcode will think this test is succeeded, or do I miss something?

igorkravchenko: XCTestExpectation is actually not so good idea. All async processes in unit tests must to be taken under control.

gersonnoboa: Yes for my welcome screen tests I was using expectation and then I was asked not to use it :) but how to figure out if the delegate stops working if it becomes async, I have no idea. Probably should be done by the person who makes that task at that point in time.

igorkravchenko: @yurii-glia @gersonnoboa 
I would suggest to create placeholder variable, store result there, and then check against expected result:
```
var receivedEvent: SomeEventType? = ...
viewModel.delegate = { event in
   receivedEvent = event
}

switch XCTUnwrap(receivedEvent) {
      case .closeTapped:
             XCTAssertTrue(true)
      default: XCTFail()
 }

```

Note that `testSendingCloseButtonEventCallsDelegate ` has to be marked as _throwing_.

GliaWidgetsTests/SecureConversations/Confirmation/SecureConversations.ConfirmationViewModelTests.swift
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    func testPressingCloseButtonCallsDelegate() {
+        viewModel.delegate = { event in
```
igorkravchenko, line 55: I thing @yurii-glia has good point that if for some reason delegate is not called it, then these tests will not evaluate.
I would suggest to change it as [suggested above](https://github.com/salemove/ios-sdk-widgets/pull/608/files#r1183284084).

igorkravchenko (commented):
After Yurii's comment, I think some slight changes have to be done to ensure that delegate closures are being called. I provided [example](https://github.com/salemove/ios-sdk-widgets/pull/608/files#r1183284084)

GliaWidgetsTests/SecureConversations/Confirmation/SecureConversations.ConfirmationViewModelTests.swift
yurii-glia: agree with @igorkravchenko suggestion.

igorkravchenko (approved):

gersonnoboa:
!squash

gersonnoboa:
!merge

PR #612, opened 2023-05-03
Take system message into account for unread message divider

igorkravchenko:
Address wrong placement of new message divider by taking system message into account.

MOB-2151

### DIFF
--- a/GliaWidgetsTests/ViewModel/Chat/ChatItemTests.swift
+++ b/GliaWidgetsTests/ViewModel/Chat/ChatItemTests.swift
@@ -1,0 +1,43 @@
+@testable import GliaWidgets
+import XCTest
+
+final class ChatItemTests: XCTestCase {
+    func testIsOperatorReturnsTrueForExpectedKindCases() {
+        let callUpgrade = ChatItem.Kind.callUpgrade(.init(with: .audio), duration: .init(with: .zero))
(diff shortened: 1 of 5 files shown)


rasmustautsglia (approved):

GliaWidgetsTests/ViewModel/Chat/ChatItemTests.swift
```diff
@@ -0,0 +1,60 @@
...
+            operatorMessage,
+            systemMessage,
+            unreadMessageDivider,
+            visitorMessage,
+            transferring
+        ]
+        for kind in kinds {
+            let chatItem = ChatItem(kind: kind)
+            switch kind {
+            case .callUpgrade:
```
gersonnoboa, line 33: I think the ones that need to be true can be grouped into one case, no? and then the same for the false ones

gersonnoboa (approved):

GliaWidgetsTests/ViewModel/Chat/ChatItemTests.swift
igorkravchenko: Indeed a84c9d6236615aba80dddccce0a80b7dd616247f.

igorkravchenko:
!squash

igorkravchenko:
!merge

PR #614, opened 2023-05-08
Fix attachment not being sent from welcome screen while on engagement

gersonnoboa:
When sending an image from the welcome screen, we need to know if an engagement is ongoing or not. If an attachment is sent to the secure messaging endpoint while on an engagement, it doesn't arrive. Thus, we check for the current engagement every time a file is uploaded. This cannot be done only on the init of the environment, because the status of the ongoing engagement can change between the moment that the screen is shown and files are uploaded.

MOB-2164

### DIFF
--- a/GliaWidgets/SecureConversations/SecureConversations.Coordinator.swift
+++ b/GliaWidgets/SecureConversations/SecureConversations.Coordinator.swift
@@ -61,7 +61,10 @@ extension SecureConversations {
                     createFileUploadListModel: environment.createFileUploadListModel,
                     fetchSiteConfigurations: environment.fetchSiteConfigurations,
                     startSocketObservation: environment.startSocketObservation,
-                    stopSocketObservation: environment.stopSocketObservation
+                    stopSocketObservation: environment.stopSocketObservation,
+                    getCurrentEngagement: environment.getCurrentEngagement,
+                    uploadSecureFile: environment.uploadSecureFile,
+                    uploadFileToEngagement: environment.uploadFileToEngagement
                 ),
                 availability: .init(
                     environment: .init(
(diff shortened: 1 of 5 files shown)


yurii-glia (approved):

GliaWidgets/SecureConversations/SecureConversations.Coordinator.swift
```diff
@@ -37,7 +37,13 @@ extension SecureConversations {
         }
 
         private func makeWelcomeViewModel() -> SecureConversations.WelcomeViewModel {
-            SecureConversations.WelcomeViewModel(
+            let uploadFile: () -> FileUpload.Environment.UploadFile = { [environment] in
+                return environment.getCurrentEngagement() == nil ?
```
igorkravchenko, line 41: Here the decision to determine which endpoint to use it placed within coordinator. I would recommend to add unit test for this logic. Aslo I wonder if this logic can be moved to view model? View models seem to be common for testing.

igorkravchenko (commented):
I think regarding testability it would be better to have decision making code within view model, instead of coordinator.

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -83,7 +83,7 @@ class ChatViewModel: EngagementViewModel, ViewModel {
         let uploader = FileUploader(
             maximumUploads: Self.maximumUploads,
             environment: .init(
-                uploadFile: .toEngagement(environment.uploadFileToEngagement),
+                uploadFile: { .toEngagement(environment.uploadFileToEngagement)} ,
```
EgorovEI, line 86:
```suggestion
                uploadFile: { .toEngagement(environment.uploadFileToEngagement) },
```

EgorovEI (approved):

GliaWidgets/SecureConversations/SecureConversations.Coordinator.swift
gersonnoboa: I have reworked the whole PR so the decision is made on the view model. Also tests were added.

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
gersonnoboa: This line was deleted so it's not necessary to fix anymore.

igorkravchenko (approved):

gersonnoboa:
!merge

PR #619, opened 2023-05-11
Add support for push notifications timing

gersonnoboa:
Receiving a push notification while in the app would make the push handler be executed because there was no way to differentiate between pushes that came while in the app vs pushes in the background. Thanks to the `timing` property added to `Push`, we can now only execute the required stuff when the push notification came while the app was on the background and the visitor accepts it.

Also, support for the new type `queuedMessage` has been added.

MOB-2177
MOB-2161

### DIFF
--- a/TestingApp/ViewController/ViewController.swift
+++ b/TestingApp/ViewController/ViewController.swift
@@ -48,15 +48,27 @@ class ViewController: UIViewController {
     }
 
     @IBAction private func chatTapped() {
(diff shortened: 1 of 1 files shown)


EgorovEI (approved):

rasmustautsglia (approved):

yurii-glia (approved):

gersonnoboa:
!merge

PR #622, opened 2023-05-15
Simulate welcome messages arrive after engagement start event

gersonnoboa:
The backend sends us messages in the incorrect order. However, product still wants us to show them correctly. Thus, what is done is the following:

1. Have socket only listen for visitor and system messages.
2. When the engagement starts, have the chat transcript downloaded.
3. Determine which operator messages are new, and add those to the messages section, not the history section.
4. Restart the socket observation.

This code is incredibly hacky, but as long as we are required to do this from our side, all solutions will be hacky. Please help me make it as clean as an abomination like this allows.

Note that this fix only works 100% well with Igor's PR in ios-sdk about message duplication. Otherwise, welcome messages are duplicated. However you can still see that all duplicated welcome messages are correctly added after the operator header.

MOB-2183

### DIFF
--- a/GliaWidgets/Sources/Coordinators/Call/CallCoordinator.swift
+++ b/GliaWidgets/Sources/Coordinators/Call/CallCoordinator.swift
@@ -81,7 +81,9 @@ private extension CallCoordinator {
                 uiApplication: environment.uiApplication,
                 fetchChatHistory: environment.fetchChatHistory,
                 fileUploadListStyle: viewFactory.theme.chatStyle.messageEntry.uploadList,
-                createFileUploadListModel: environment.createFileUploadListModel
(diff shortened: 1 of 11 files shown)


igorkravchenko (commented):
I don't know how to achieve the expected result otherwise, so it seems like working a solution. Do you plan to add tests for it?

gersonnoboa:
@igorkravchenko let's see what happens in the channel. There's a possibility that this behavior will be kept as it is before this PR, so I won't add unit tests at this time. However, if it is decided that we should move forward with this solution, i will add tests and then convert this draft into a normal PR.

gersonnoboa:
Closed because it was decided to fix this from backend.

PR #623, opened 2023-05-15
Fix media upgrade not working on transcript converted to live chat

gersonnoboa:
When the secure transcript is upgraded to live chat, the root coordinator doesn't know know about this and thinks it is still in secure conversations. With this commit, at the time where the view models are swapped, a message is sent through the delegate chain to correctly assign the type of engagement to the coordinator. This in turn means that the coordinator can handle upgrades correctly.

MOB-2192


EgorovEI (approved):

igorkravchenko (approved):

rasmustautsglia (approved):

gersonnoboa:
!merge

PR #624, opened 2023-05-15
Avoid last message being cut off when clearing unread message divider

gersonnoboa:
When the unread message divider was disappearing, the scroll view wasn't going to the bottom again (if already at the bottom). With this commit, if the visitor is at the bottom, then the scroll view is scrolled to the bottom. If not, then the default behavior is that it more or less keeps its place, taking into consideration that the unread message divider has disappeared.

MOB-2193


EgorovEI (approved):

igorkravchenko (approved):

gersonnoboa:
!merge

PR #626, opened 2023-05-16
Make old deprecated methods unavailable

rasmustautsglia:
Glia's policy is to keep deprecated methods in use for additional 6 months. This PR makes older methods unavailable. This PR also covers 1319 and 1348 tickets

MOB-1319 and MOB-1348


EgorovEI (approved):

igorkravchenko (approved):
@rasmustautsglia I think next logical step would be to schedule unavailable APIs removal, by creating corresponding tickets.

PR #621, opened 2023-05-12
Devs want to deprecate iOS 12 support

yurii-glia:
MOB-2185

PR #627, opened 2023-05-17
Add QuickLookBased.ThumbnailGenerator for cropping images

EgorovEI:
Previously for retrieving images from file system UIImage.imageWithContentsOfFileAtPath was used and heavyweight operation for cropping. Now it was replaced with QuickLookBased.ThumbnailGenerator which retrieves already compressed image and use new lightweight operation for cropping images to necessary size.

MOB-2149


EgorovEI:
!squash

GliaWidgets/Sources/QuickLookBased/QuickLookBased.Implementation.swift
```diff
@@ -0,0 +1,35 @@
...
+                        size: newSize,
+                        scale: scale,
+                        representationTypes: .thumbnail
+                    )
+                    generator.generateBestRepresentation(for: request) { thumbnail, error in
+                        if let image = thumbnail?.uiImage {
+                            let cropped = image.resized(to: size)
+                            completion(cropped)
+                        } else if error != nil {
+                            // TODO: need to handle error state
```
gersonnoboa, line 24: Is there a task for this?

gersonnoboa (approved):

GliaWidgets/Sources/QuickLookBased/QuickLookBased.Implementation.swift
```diff
@@ -0,0 +1,35 @@
+import Foundation
+import UIKit
+import QuickLook
+
+extension QuickLookBased.ThumbnailGenerator {
+    static func create(uiScreen: UIKitBased.UIScreen) -> Self {
+        if #available(iOS 13.0, *) {
```
igorkravchenko, line 7: Will we need this check if this change will be applied once deployment target is iOS 13?

GliaWidgets/Sources/QuickLookBased/QuickLookBased.Implementation.swift
```diff
@@ -0,0 +1,35 @@
...
+                            let cropped = image.resized(to: size)
+                            completion(cropped)
+                        } else if error != nil {
+                            // TODO: need to handle error state
+                        }
+                    }
+                }
+            )
+        } else {
+            return .init { _, _, completion in
```
igorkravchenko, line 30: In case we if need this `else` case, can we think of fallback, like provide laggy but non-crashing solution?

rasmustautsglia (approved):
Looks good to me

GliaWidgets/Sources/Upload/FileUploader.Environment.Interface.swift
```diff
@@ -7,8 +7,7 @@ extension FileUploader {
         var data: FoundationBased.Data
         var date: () -> Date
         var gcd: GCD
-        var localFileThumbnailQueue: FoundationBased.OperationQueue
-        var uiImage: UIKitBased.UIImage
+        var createThumbnailGenerator: () -> QuickLookBased.ThumbnailGenerator
```
igorkravchenko, line 10: I wonder if we are to benefit from multiple instances of `ThumbnailGenerator`. Maybe one will be enough to be shared between cells? Or it does impact performance and causes issue when shared?

igorkravchenko (commented):
Overall looks good. Left some questions.

GliaWidgets/Sources/QuickLookBased/QuickLookBased.Implementation.swift
EgorovEI: I would raise discussion about handling this error state. Because it might require some design work. Will do this right now

EgorovEI: https://salemove.slack.com/archives/C01F9K88SQZ/p1684403981765899

GliaWidgets/Sources/QuickLookBased/QuickLookBased.Implementation.swift
EgorovEI: No, it will be removed once deployment target is changed.

GliaWidgets/Sources/QuickLookBased/QuickLookBased.Implementation.swift
EgorovEI: It's not necessary because it will be removed once deployment target is changed.

GliaWidgets/Sources/Upload/FileUploader.Environment.Interface.swift
EgorovEI: I've tested using one `QLThumbnailGenerator.shared` instance and it worked significantly worse rather then separate ones

EgorovEI:
Waits for https://github.com/salemove/ios-sdk-widgets/pull/621 being merged

GliaWidgets/Sources/File/LocalFile.swift
```diff
@@ -53,9 +53,14 @@ extension LocalFile {
...
             return
         } else {
-            environment.thumbnailGenerator.generateBestRepresentation(size, url.standardizedFileURL) { [weak self] image in
+            environment.thumbnailGenerator.createImageThumbnail(
+                fileURL: url,
+                size: size,
+                scale: environment.uiScreen.scale()
+            ) { [weak self] image in
+                let cropped = image?.resized(to: size)
                 self?.environment.gcd.mainQueue.async {
```
igorkravchenko, line 62: Is it still necessary to use thread hop like `mainQueue.async`?

GliaWidgets/Sources/QuickLookBased/QuickLookBased.Live.swift
```diff
@@ -0,0 +1,43 @@
...
+            fileAt: fileURL.standardizedFileURL,
+            size: newSize,
+            scale: scale,
+            representationTypes: .thumbnail
+        )
+        generateBestRepresentation(request) { thumbnail, error in
+            if let image = thumbnail?.uiImage {
+                completion(image)
+            } else if error != nil {
+                // TODO: need to handle error state
```
igorkravchenko, line 33: Maybe it makes sense to at least log error to console?

igorkravchenko (commented):
Left a few minot comments.

GliaWidgets/Sources/File/LocalFile.swift
EgorovEI: Yes, because `QLThumbnailGenerator` return representation on background thread

GliaWidgets/Sources/QuickLookBased/QuickLookBased.Live.swift
EgorovEI: abaeca442de15a713f0fe344a234883623185251

PR #628, opened 2023-05-18
Make old configure method unavailable

rasmustautsglia:
This PR makes the old configure method unavailable and since an even older one called the first one, it was removed entirely. In addition, app token was removed from authorization methods, because it has been deprecated for more than 12 months.

MOB-1912

### DIFF
--- a/GliaWidgets/Public/Glia/Glia.Deprecated.swift
+++ b/GliaWidgets/Public/Glia/Glia.Deprecated.swift
@@ -2,7 +2,7 @@ import SalemoveSDK
 
 extension Glia {
     /// Deprecated.
-    @available(*, deprecated, message: "Use configure(with:queueId:uiConfig:assetsBuilder:completion:) instead.")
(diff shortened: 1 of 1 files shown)


EgorovEI:
Are you sure that can be marked as unavailable? 
@yurii-glia said appToken is still in use https://salemove.slack.com/archives/CHN6UB7UH/p1684403642031079?thread_ts=1684400655.954519&cid=CHN6UB7UH

gersonnoboa (approved):

EgorovEI (approved):

gersonnoboa:
@EgorovEI I guess that they use it but with older versions of the SDK. But that is a valid concern, could be asked in the channel.

EgorovEI:
> @EgorovEI I guess that they use it but with older versions of the SDK. But that is a valid concern, could be asked in the channel.

@rasmustautsglia could raise this question in channel one more time?

rasmustautsglia:
!squash

rasmustautsglia:
!merge

PR #629, opened 2023-05-18
Introduce new `clearVisitorSession` method

EgorovEI:
New`clearVisitorSession(_:)` method was added to provide the ability to throw the error to integrator in case when they try to clear visitor session during active engagement.

MOB-2136

### DIFF
--- a/GliaWidgets/Public/Glia/Glia.Deprecated.swift
+++ b/GliaWidgets/Public/Glia/Glia.Deprecated.swift
@@ -70,4 +70,23 @@ extension Glia {
             sceneProvider: sceneProvider
         )
     }
+
+    /// Deprecated, use `clearVisitorSession(_:)` instead.
+    @available(
+        *,
+         deprecated,
+         message: "Use clearVisitorSession(_:) instead."
+    )
+    public func clearVisitorSession() {
+        if environment.coreSdk.getCurrentEngagement() != nil {
+            print("⚠️ Don't call `clearVisitorSession` during active engagement. Otherwise, it might break the engagement interaction.")
+        }
+        environment.coreSdk.clearSession()
+    }
+
+    /// Deprecated, use ``callVisualizer.showVisitorCodeViewController`` instead.
+    @available(*, deprecated, message: "Deprecated, use ``CallVisualizer.showVisitorCodeViewController`` instead.")
+    public func requestVisitorCode(completion: @escaping (Result<VisitorCode, Swift.Error>) -> Void) {
+        _ = environment.coreSdk.requestVisitorCode(completion)
+    }
 }
--- a/GliaWidgets/Public/Glia/Glia.swift
+++ b/GliaWidgets/Public/Glia/Glia.swift
@@ -157,8 +157,19 @@ public class Glia {
     }
 
     /// Clear visitor session
-    public func clearVisitorSession() {
+    ///
+    /// - Parameter completion: Completion handler.
+    ///
+    /// - Important: Note, that in case of ongoing engagement, `clearVisitorSession` must be called after ending engagement,
+    /// because `GliaError.clearingVisitorSessionDuringEngagementIsNotAllowed` will occur otherwise.
+    ///
+    public func clearVisitorSession(_ completion: @escaping (Result<Void, Error>) -> Void) {
+        guard environment.coreSdk.getCurrentEngagement() == nil else {
+            completion(.failure(GliaError.clearingVisitorSessionDuringEngagementIsNotAllowed))
+            return
+        }
         environment.coreSdk.clearSession()
+        completion(.success(()))
     }
 
     /// Fetch current Visitor's information.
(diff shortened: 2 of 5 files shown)


rasmustautsglia (approved):
Looks good!

igorkravchenko (approved):

GliaWidgets/Public/Glia/Glia.Deprecated.swift
```diff
@@ -84,4 +84,24 @@ extension Glia {
...
+        do {
+            try clearVisitorSessionIfPossible()
+        } catch {
+            print("⚠️ `clearVisitorSession` failed with error: `\(error)`")
+        }
+    }
+
+    /// Deprecated, use ``callVisualizer.showVisitorCodeViewController`` instead.
+    @available(*, deprecated, message: "Deprecated, use ``CallVisualizer.showVisitorCodeViewController`` instead.")
+    public func requestVisitorCode(completion: @escaping (Result<VisitorCode, Swift.Error>) -> Void) {
```
gersonnoboa, line 104: I don't remember, was this method part of the SDK before the call visualizer project? If it was, then it's fine that it's deprecated. If it was done during the CV project, then it doesn't need to be deprecated because it was never officially released.

GliaWidgets/Public/Glia/Glia.Deprecated.swift
```diff
@@ -84,4 +84,24 @@ extension Glia {
...
         )
     }
+
+    /// Deprecated, use `clearVisitorSessionIfPossible() throws` instead.
+    @available(
+        *,
+         deprecated,
+         message: "Use clearVisitorSessionIfPossible() throws instead."
+    )
+    public func clearVisitorSession() {
```
gersonnoboa, line 94: But isn't this changing the previous version? With the previous one, the clearing was executed always, with whatever side effects it would had. With this new version, it only is executed if it's possible, otherwise it fails silently. While this is clearly better, it is still modifying the old one which I think we are not allowed to do.

GliaWidgets/Public/Glia/Glia.Deprecated.swift
EgorovEI: Good point. I'll double check it

GliaWidgets/Public/Glia/Glia.Deprecated.swift
EgorovEI: I will raise discussion.

GliaWidgets/Public/Glia/Glia.Deprecated.swift
EgorovEI: We added this interface in January. Then we had 3 releases. Form one side, that were not releases for CV, but potentially some integrators could use it. And this deprecation was made in February and I just moved it to appropriate file. 
Based on that I would leave it as deprecated. 
@igorkravchenko @rasmustautsglia @yurii-glia wdyt?

GliaWidgets/Public/Glia/Glia.Deprecated.swift
EgorovEI: Conclusion of the [discussion](https://salemove.slack.com/archives/C01F9K88SQZ/p1684504908521729):
`add to existing old method the error logging (into debug console) about active engagement. Existing old method will work as previously, but with logging error to debug console.`

c4dbce994198b4b072929de2e4133ec8f0df2e60

rasmustautsglia (approved):

GliaWidgets/Public/Glia/Glia.Deprecated.swift
igorkravchenko: If it was not a part of new call visualizer, but officially (mentioned in documentation) released prior to that, it makes sense to deprecate it.

GliaWidgets/Public/Glia/Glia.swift
```diff
@@ -157,8 +157,19 @@ public class Glia {
...
 
     /// Clear visitor session
-    public func clearVisitorSession() {
+    ///
+    /// - Parameter completion: Completion handler.
+    ///
+    /// - Important: Note, that in case of ongoing engagement, `clearVisitorSession` must be called after ending engagement,
+    /// because `GliaError.clearingVisitorSessionDuringEngagementIsNotAllowed` will occur otherwise.
+    ///
+    public func clearVisitorSession(_ completion: (Result<Void, Error>) -> Void) {
```
igorkravchenko, line 166: Since `clearVisitorSession` has changed its type from `() -> Void` to `((Result<Void, Error>) -> Void ) -> Void` that will become a breaking change for integrators. Also completion is not `@escaping` at the moment, meaning if we are to do some async work there in future, we will be forced be compiler to deprecate it and introduce `@escaping` completion. Also what benefit do we get right now from it if it is not escaping? So to address both potential problems I suggest to add `@escaping` and make compilation optional.

```suggestion
    public func clearVisitorSession(_ completion: @escaping ((Result<Void, Error>) -> Void)? = nil) {
```

Does it make sense @EgorovEI @gersonnoboa @yurii-glia @rasmustautsglia?

igorkravchenko (commented):
Left comment about completion handler.

GliaWidgets/Public/Glia/Glia.swift
EgorovEI: You are totally right.

EgorovEI: 85e0de910eba66310c6f8ab45a11b11e991df6ca

GliaWidgets/Public/Glia/Glia.Deprecated.swift
EgorovEI: But this interface deprecation was already published also.

gersonnoboa: Ahhh okay, then leave it as it is.

gersonnoboa (approved):

EgorovEI:
!squash

yurii-glia (approved):

PR #636, opened 2023-05-30
Make accessible file download and upload view

EgorovEI:
MOB-1414


gersonnoboa (commented):
Normally we put in the accessibility identifier the feature to which it is part. So for example `secureConversations_fileUploading_view` or button or something similar. Although this is not enforced throughout the code, but when I have the opportunity to code review, I think it's better to have some kind of structure for the identifier. Do you think that it's worth it to add here?

EgorovEI:
!squash

igorkravchenko (approved):

yurii-glia (approved):

PR #637, opened 2023-05-31
Fix backend response error if external_access_token is empty string

yurii-glia:


TestingApp/ViewController/ViewController.swift
```diff
@@ -412,7 +412,7 @@ extension ViewController {
             title: "Create Authentication",
             style: .default
         ) { _ in
-            authentication.authenticate(with: enteredJwt, accessToken: enteredAccessToken) { [weak self] result in
+            authentication.authenticate(with: enteredJwt, accessToken: enteredAccessToken.isEmpty ? nil : enteredAccessToken) { [weak self] result in
```
gersonnoboa, line 415: I know that we are a little bit in a hurry but this line is enormous :) can the access token be its own variable and then used here?

EgorovEI (approved):

igorkravchenko (approved):

rasmustautsglia (approved):

TestingApp/ViewController/ViewController.swift
yurii-glia: fixed - cd1dcd461e9552867685c3d2f28e35dfb0e986fd

yurii-glia:
!squash

PR #641, opened 2023-06-01
PoC for theme remote configuration tests

igorkravchenko:
Provide proof of concept of how theme remote configuration can be tested using snapshot images. The goal of this PR is serve as example how we can ensure that changes in unified customisation can be backed up by tests. If something like this can be achieved for Android, we can use this approach to cover both platforms.

MOB-2010

### DIFF
--- a/SnapshotTests/ThemeRemoteConfigurationTests.swift
+++ b/SnapshotTests/ThemeRemoteConfigurationTests.swift
@@ -1,0 +1,34 @@
+@testable import GliaWidgets
+import XCTest
+import SnapshotTesting
+
+final class ThemeRemoteConfigurationTests: SnapshotTestCase {
+    func test_themeForMessagesFromHistory() throws {
+        let url = try XCTUnwrap(
+            Bundle(for: Self.self)
+                .url(
+                    forResource: "unified_sample.json",
+                    withExtension: nil
+                )
+        )
+
+        let jsonData = try Data(contentsOf: url)
+        let config = try JSONDecoder().decode(
+            RemoteConfiguration.self,
+            from: .init(jsonData)
+        )
+
+        let theme = Theme(
+            uiConfig: config,
+            assetsBuilder: .standard
+        )
+
+        let viewController = ChatViewController.mockHistoryMessagesScreen(theme: theme)
+        viewController.view.frame = UIScreen.main.bounds
+        assertSnapshot(
+            matching: viewController,
+            as: .image,
+            named: nameForDevice()
+        )
+    }
+}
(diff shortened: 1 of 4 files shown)


SnapshotTests/ThemeRemoteConfigurationTests.swift
```diff
@@ -0,0 +1,34 @@
+@testable import GliaWidgets
+import XCTest
+import SnapshotTesting
+
+final class ThemeRemoteConfigurationTests: SnapshotTestCase {
+    func test_themeForMessagesFromHistory() throws {
```
gersonnoboa, line 6: So the idea is to make snapshots with the default theme and the same snapshots but with remote configuration?

gersonnoboa (approved):

EgorovEI (approved):

SnapshotTests/ThemeRemoteConfigurationTests.swift
igorkravchenko: Yes, If needed we can capture default theme styling, then from file configuration. That will allow us keep changes in both themes.

andrews-moc (approved):
Looks good to me.
However, I can't say how it can look like on Android because we don't have snapshot tests on Android yet, unfortunately.

yurii-glia (approved):

SnapshotTests/ThemeRemoteConfigurationTests.swift
```diff
@@ -0,0 +1,34 @@
...
+            assetsBuilder: .standard
+        )
+
+        let viewController = ChatViewController.mockHistoryMessagesScreen(theme: theme)
+        viewController.view.frame = UIScreen.main.bounds
+        assertSnapshot(
+            matching: viewController,
+            as: .image,
+            named: nameForDevice()
+        )
```
gugalo, line 32: Need confirmation that I understood that correctly..
Tests generate an image of the mocked view that is generated based on remote theme configurations`unified_sample` and compare it to the view that is currently on the device screen? Is that correct?

gugalo: But basically, the test is using the same view as the Widget SDK for the test but just with mocked properties? Wouldn't that mean that if there is a bug in setting the color of the icon then whatever the theme (default/remote) the result would be the same and the test would anyway pass?

gugalo (commented):
My iOS experience is lacking. Did not really understand how this works

DavDo (approved):

SnapshotTests/ThemeRemoteConfigurationTests.swift
igorkravchenko: @gugalo Using same view with mocked properties will not pass if theme related configuration logic will change, because snapshots will be different. 

Test scenario works like this:
1. Configure theme using `unified_sample.json`.
2. Apply configured theme to UI.
3. Bring UI into desired state by mocking business logic.
4. Create snapshot image of a view under test.
5. Compare current snapshot with previously generated one.
6. If snapshots ([current](
https://user-images.githubusercontent.com/3148347/242498686-ef225be3-54b2-4228-bc41-9f59a2ac2d0f.png) and [previous](https://user-images.githubusercontent.com/3148347/242496791-79b274e0-fa18-42fd-bb54-f0f779eb11a8.png)) have difference, then raise test failure, resulting in [diff image](
https://user-images.githubusercontent.com/3148347/242500384-f93afdea-c0b2-4622-b816-6d652fd07d8e.png)
7. In case this test is run for the first time ever, it will generate initial image, but it will also fail, forcing test creator to check if generated image looks correctly. That initial generated image will serve as reference for the subsequent test cases.

gugalo: Thank you! That was very nice explanation 👍

gugalo (approved):

PR #646, opened 2023-06-02
Add correct commit author to Bitrise commits

gersonnoboa:
Bitrise does commits sometimes for us, like for example updating dependencies. Bitrise's documentation recommends to introduce an envvar with the values. These are the same values used in other projects.

MOB-1696

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -1,5 +1,4 @@
----
-format_version: '8'
+format_version: "8"
 default_step_lib_source: https://github.com/bitrise-io/bitrise-steplib.git
 project_type: ios
 workflows:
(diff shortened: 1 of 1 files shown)


EgorovEI (approved):

yurii-glia (approved):
we worked with Rasmus on upload to AppStoreConnect task and I found that Bitrise stores configuration in another format, so decided to align with what we have in bitrise.yml just to avoid big diff in future.
however, if your editor think this is more correct format, we can leave. what editor do you use?

igorkravchenko (approved):

gersonnoboa:
@yurii-glia I think the discrepancy comes when you save the file from Bitrise, vs when you copy the contents that Bitrise gives you and paste them yourself in the bitrise.yml. Or something liek that. They have had this issue for ages https://github.com/bitrise-io/bitrise-workflow-editor/issues/572

gersonnoboa:
!merge

PR #647, opened 2023-06-08
Address crash in chat view during video call.

igorkravchenko:
If chat contains some messages (like default welcome message for queue), and engagement is started as video engagement, a crash could occur during 'tableView.performBatchUpdates'. This is addressed by avoiding calling 'tableView.performBatchUpdates' for case when table view size is zero.

MOB-2260


yurii-glia (approved):

EgorovEI (approved):

igorkravchenko:
!merge

PR #653, opened 2023-06-09
Increment project version to 2.0.2

sm-deployer:


yurii-glia (changes requested):
podspec has not been updated

PR #659, opened 2023-06-22
Fixed subsequent media upgrade alerts

EgorovEI:
Before:
- if operator requests more than one media upgrade in a row, we show only the first one and skip further. As a result, even if visitor presses “Accept” button, the SDK sends response for outdated request, so nothing happens on operator side, but it causes the issue on SDK side. 
- If PopoverViewController is presented, no alerts could be presented, even alert about ending engagement (only if no survey).
- If any AlertViewController is presented, end engagement `singleAction` alert could not be presented, as result, SDK doesn't close an engagement (only if no survey)

Now: 
- `PresentationPriority` is introduced, which we can rely on when decide whether already presented alert should be replaced with incoming one.
- Media upgrade alerts along with screen sharing request can be replaced with another one if operator re-requests it.
- PopoverViewController will be dismissed when any media request comes.
- When visitor presses "End" engagement button, or presses stop screen sharing button, so appropriate confirmation alert is shown and operator requests any media upgrade, this media upgrade alert is skipped, because "user-initiated" alert has `high` priority.
- (only if no survey for the queue) When operator end engagement and PopoverViewController or any media upgrade alert or "user-initiated" alert is presented, it will be dismissed and end engagement `singleAction` alert will be presented.

MOB-2264

### DIFF
--- a/GliaWidgets/Sources/ViewController/Common/Alert/AlertViewController.swift
+++ b/GliaWidgets/Sources/ViewController/Common/Alert/AlertViewController.swift
@@ -1,6 +1,6 @@
 import UIKit
 
-class AlertViewController: UIViewController {
+class AlertViewController: UIViewController, Replaceable {
     enum Kind {
         case message(
             MessageAlertConfiguration,
@@ -27,9 +27,28 @@ class AlertViewController: UIViewController {
             accepted: () -> Void,
             declined: () -> Void
         )
+
+        /// Indicating presentation priority of an alert.
+        /// Based on comparing values we can decide whether an alert can be replaced with another alert.
+        fileprivate var presentationPriority: PresentationPriority {
+            switch self {
+            case .singleAction:
+                return .highest
+            case .confirmation:
+                return .high
+            case .message, .singleMediaUpgrade, .screenShareOffer:
+                return .regular
+            }
+        }
     }
 
     let viewFactory: ViewFactory
+
+    /// Indicating presentation priority of an alert.
+    /// Based on comparing values we can decide whether an alert can be replaced with another alert.
+    var presentationPriority: PresentationPriority {
+        kind.presentationPriority
+    }
 
     private let kind: Kind
     private var alertView: AlertView?
--- a/GliaWidgets/Sources/ViewController/Common/MediaUpgrade/MediaUpgradePresenter.swift
+++ b/GliaWidgets/Sources/ViewController/Common/MediaUpgrade/MediaUpgradePresenter.swift
@@ -1,6 +1,6 @@
 import UIKit
 
-protocol MediaUpgradePresenter where Self: UIViewController {
+protocol MediaUpgradePresenter: DismissalAndPresentationController where Self: UIViewController {
     var viewFactory: ViewFactory { get }
 
     func offerMediaUpgrade(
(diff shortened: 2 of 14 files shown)


GliaWidgets/Sources/ViewController/Common/Alert/AlertViewController.swift
```diff
@@ -27,10 +27,28 @@ class AlertViewController: UIViewController {
             accepted: () -> Void,
             declined: () -> Void
         )
+
+        /// Indicating whether an alert is as a result on user action,
+        /// like pressing "End" engagement button, or pressing stop screen sharing button.
+        /// Based on that user-initiated can be only the `confirmation`case
+        fileprivate var isUserInitiated: Bool {
```
igorkravchenko, line 34: Maybe this property can be `private (set)`? This may allow unit testing of it as well.

GliaWidgets/Sources/ViewController/Common/Alert/AlertViewController.swift
```diff
@@ -27,10 +27,28 @@ class AlertViewController: UIViewController {
...
+                return false
+            }
+        }
     }
 
     let viewFactory: ViewFactory
 
+    /// Indicating whether alert can be replaced with another alert.
+    /// When `kind.isUserInitiated` is true, the alert should not be dismissed
+    var canBeReplaced: Bool {
```
igorkravchenko, line 48: I wonder if there's a better wording for this. `isReplaceable`, `allowsSubstitution`?

GliaWidgets/Sources/ViewController/Common/MediaUpgrade/MediaUpgradePresenter.swift
```diff
@@ -16,12 +16,25 @@ extension AlertPresenter {
...
+                kind: .singleMediaUpgrade(
+                    conf,
+                    accepted: accepted,
+                    declined: declined
+                ),
+                viewFactory: self.viewFactory
+            )
+            self.present(alert, animated: true)
+        }
+        guard let presented = self.presentedViewController else {
```
igorkravchenko, line 30: Just a thought: is it possible to automate the logic of previous view dismissal by moving it into alert? Because currently such logic will be scattered all over the codebase.

igorkravchenko (commented):
Left a few comments.

GliaWidgets/Sources/ViewController/Common/Alert/AlertViewController.swift
EgorovEI: This is enum property, so it can be stored

GliaWidgets/Sources/ViewController/Common/Alert/AlertViewController.swift
EgorovEI: 958dc160f6820f72e126ecfb07bd21edb6fe3f49

GliaWidgets/Sources/ViewController/Common/Alert/AlertViewController.swift
```diff
@@ -27,10 +27,29 @@ class AlertViewController: UIViewController {
             accepted: () -> Void,
             declined: () -> Void
         )
+
+        /// Indicating presentation priority of an alert.
+        /// Based on comparing values we can decide whether an alert can be replaced with another alert.
+        fileprivate var presentationPriority: PresentationPriority {
```
yurii-glia, line 33: `fileprivate` and `private` work the same way in one file, so we can simplify it and use just `private` instead. would it make sense?

yurii-glia (commented):
we will need to re-work alert presentation and support presenting them in a queue (one by one), would this solution be scalable for achieving this mechanism?

GliaWidgets/Sources/ViewController/Common/Alert/AlertViewController.swift
EgorovEI: When `presentationPriority` is `private`, it can't be accessible out of `Kind` scope. So that's why I make it `fileprivate`

EgorovEI:
> we will need to re-work alert presentation and support presenting them in a queue (one by one), would this solution be scalable for achieving this mechanism?

- I thought about "chaining alerts in a queue" and the only reason to implement it is chaining media upgrade with screen sharing request. It requires defining another strategy for replacing, adding, removing and forcing alerts in this "chaining queue"
Current approach might be partially used.

GliaWidgets/Sources/ViewController/Common/Alert/AlertViewController.swift
yurii-glia: oh, that is in nested class. ok.

yurii-glia (approved):

EgorovEI:
!squash

rasmustautsglia (approved):

PR #661, opened 2023-06-29
Expand textview in the presence of a long text

rasmustautsglia:
This PR makes the text field expandable if the text is longer than there is room on a line. In total, the textfield can expand to up to 4 lines. If the text is longer than 4 lines, then the textfield becomes scrollable. 

In addition, the max character limit had been previously set to 200. I have increased it to 10000 characters to align it with web, which is also the API's limit.

MOB-2334

### DIFF
--- a/GliaWidgets/Sources/View/Chat/Entry/ChatMessageEntryView.swift
+++ b/GliaWidgets/Sources/View/Chat/Entry/ChatMessageEntryView.swift
@@ -3,7 +3,6 @@ import UIKit
 class ChatMessageEntryView: BaseView {
     let pickMediaButton: MessageButton
     let uploadListView: SecureConversations.FileUploadListView
-    var maxCharacters: Int = 200
     var textChanged: ((String) -> Void)?
     var sendTapped: (() -> Void)?
     var pickMediaTapped: (() -> Void)?
@@ -55,8 +54,9 @@ class ChatMessageEntryView: BaseView {
     private let sendButton: MessageButton
     private let buttonsStackView = UIStackView()
     private var textViewHeightConstraint: NSLayoutConstraint?
-    private let kMinTextViewHeight: CGFloat = 24
-    private let kMaxTextViewHeight: CGFloat = 200
+    private let minTextViewHeight: CGFloat = 24
+    private let maxTextLines = 4
+    private let maxCharacters = 10000
 
     private let environment: Environment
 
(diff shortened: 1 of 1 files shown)


GliaWidgets/Sources/View/Chat/Entry/ChatMessageEntryView.swift
```diff
@@ -219,17 +218,23 @@ class ChatMessageEntryView: BaseView {
...
-        if newHeight > kMaxTextViewHeight {
-            newHeight = kMaxTextViewHeight
-        } else if newHeight < kMinTextViewHeight {
-            newHeight = kMinTextViewHeight
+        textView.isScrollEnabled = Int(numLines) > maxTextLines
+
+        if Int(numLines) > maxTextLines {
+            if let lineHeight = textView.font?.lineHeight {
+                newHeight = lineHeight * CGFloat(maxTextLines)
+            }
```
yurii-glia, line 236:
```suggestion
        if Int(numLines) > maxTextLines {
            newHeight = font.lineHeight * CGFloat(maxTextLines)
```

GliaWidgets/Sources/View/Chat/Entry/ChatMessageEntryView.swift
```diff
@@ -219,17 +217,23 @@ class ChatMessageEntryView: BaseView {
             height: CGFloat.greatestFiniteMagnitude
         )
 
-        // We need to take into account larger height
-        // of text view or placeholder label for
-        // updated text entry height.
-        var newHeight = max(textView.sizeThatFits(size).height, placeholderLabel.sizeThatFits(size).height)
+        guard let font = textView.font else {
```
yurii-glia, line 220: do we need to handle here not valid lineHeight that can lead to division by zero error?
```suggestion
        guard let font = textView.font, font.lineHeight > 0 else {
```

yurii-glia (approved):

EgorovEI (approved):

GliaWidgets/Sources/View/Chat/Entry/ChatMessageEntryView.swift
rasmustautsglia: Fixed [04c7b69](https://github.com/salemove/ios-sdk-widgets/pull/661/commits/04c7b6911627931a298245bae76d014f53953098)

rasmustautsglia:
!squash

PR #664, opened 2023-07-05
Unit tests for text entry field expansion

rasmustautsglia:
Textfield is now expandable and after being over 4 lines, need to be scrollable. This PR covers this functionality with tests to ensure the quality.

MOB-2336

### DIFF
--- a/GliaWidgetsTests/ViewModel/Chat/ChatMessagaEntryViewTests.swift
+++ b/GliaWidgetsTests/ViewModel/Chat/ChatMessagaEntryViewTests.swift
@@ -1,0 +1,40 @@
+@testable import GliaWidgets
+import XCTest
+
+final class ChatMessageEntryViewTests: XCTestCase {
+    var messagageEntryView: ChatMessageEntryView!
+
+    override func setUp() {
+        super.setUp()
+        messagageEntryView = ChatMessageEntryView.mock()
+    }
+
+    override func tearDown() {
+        messagageEntryView = nil
+        super.tearDown()
+    }
+
+    func test_text_view_height_with_more_than_max_lines() {
+        let exampleText = "This\nis a\nfive\nlines\ntext"
+        let maxTextLines = CGFloat(messagageEntryView.maxTextLines)
+        let oneLineHeight = Theme().chatStyle.messageEntry.messageFont.lineHeight
+        let expectedTextViewHeight = oneLineHeight * maxTextLines
+
+        messagageEntryView.messageText = exampleText
+        messagageEntryView.defineLayout()
+        let textActualHeight = messagageEntryView.textViewHeightConstraint?.constant
+
+        XCTAssertEqual(textActualHeight ?? 0.0, expectedTextViewHeight)
+    }
+
+    func test_text_view_is_scrollable() {
+        messagageEntryView.defineLayout()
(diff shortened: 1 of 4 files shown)


GliaWidgetsTests/ViewModel/Chat/ChatMessagaEntryViewTests.swift
```diff
@@ -0,0 +1,30 @@
+@testable import GliaWidgets
+import XCTest
+
+class ChatMessageEntryViewTests: XCTestCase {
+    let messagageEntryView = ChatMessageEntryView.mock()
```
EgorovEI, line 5: Since `messagageEntryView` is shared across tests, you need to erase its state in `setUp` method

GliaWidgetsTests/ViewModel/Chat/ChatMessagaEntryViewTests.swift
```diff
@@ -0,0 +1,30 @@
+@testable import GliaWidgets
+import XCTest
+
+class ChatMessageEntryViewTests: XCTestCase {
```
EgorovEI, line 4: `final`

GliaWidgetsTests/ViewModel/Chat/ChatMessagaEntryViewTests.swift
```diff
@@ -0,0 +1,30 @@
...
+    let messagageEntryView = ChatMessageEntryView.mock()
+
+    func test_text_view_height_with_more_than_max_lines() {
+        let exampleText = "This\nis a\nfive\nlines\ntext"
+        let maxTextLines = CGFloat(messagageEntryView.maxTextLines)
+        let oneLineHeight = Theme().chatStyle.messageEntry.messageFont.lineHeight
```
gersonnoboa, line 10: I am not familiar enough with this logic, but does this getting the line height and then multiplying to get the text view height works well with other fonts that are not San Francisco?

rasmustautsglia: Yes, the lineHeight property of UIFont is designed to work with any font, not just San Francisco. When you use a different font, the lineHeight property will return a different value that is appropriate for that font. Different fonts can have different line heights even for the same point size, because the line height is part of the design of the font and is set by the font designer. This is why it's important to use the lineHeight property rather than hard-coding a specific line height, because it ensures that your layout respects the design of the font.

GliaWidgetsTests/ViewModel/Chat/ChatMessagaEntryViewTests.swift
rasmustautsglia: fixed [cc188a5](https://github.com/salemove/ios-sdk-widgets/pull/664/commits/cc188a5d227e4a0a199d66682f8699a55ac65fd7)

EgorovEI (approved):

GliaWidgetsTests/ViewModel/Chat/ChatMessagaEntryViewTests.swift
gersonnoboa: Great information here :) thank you

gersonnoboa (approved):

rasmustautsglia:
!squash

PR #669, opened 2023-07-06
Run SwiftLint only on changed files

gersonnoboa:
SwiftLint was deactivated because it would run on every build. This commit makes it so that SwiftLint is only executed on files changed according to `git status`. To make this work, current SwiftLint issues have been fixed. Also, Xcode will report all SwiftLint warnings as errors.

MOB-1973


rasmustautsglia (approved):

EgorovEI (approved):

gersonnoboa:
!merge

PR #670, opened 2023-07-07
Text entry field remains active when interacting with chat

rasmustautsglia:
Previously in the presence of customCard (or now GVA) the text entry field was disabled and couldn't be used until user had made a choice. Now this has been removed and text entry field stays active during custom cards / GVA and even when user has pressed 'send'

MOB-2332

### DIFF
--- a/GliaWidgets/Sources/View/Chat/ChatView.swift
+++ b/GliaWidgets/Sources/View/Chat/ChatView.swift
@@ -312,8 +312,6 @@ extension ChatView {
         )
 
         tapGesture.cancelsTouchesInView = false
-
-        tableView.addGestureRecognizer(tapGesture)
     }
 
     @objc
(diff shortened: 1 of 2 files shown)


EgorovEI:
Are the changes agreed?

rasmustautsglia:
Yes, they are part of the project plan

EgorovEI:
But it look as it affects choice and custom cards. Or it's ok?

rasmustautsglia:
Yes, it will affect custom cards in a way that user is no longer forced to make choice but can continue writing if they wanted to. This is now expected behavior.

EgorovEI (approved):

gersonnoboa (commented):
Code makes sense but there are two SwiftLint issues. Weird that they didn't come up for you in Xcode, could you confirm?

rasmustautsglia:
fixed [53032b6](https://github.com/salemove/ios-sdk-widgets/pull/670/commits/53032b69e47d570711d5d8910801be88d65a8710)

gersonnoboa (approved):

rasmustautsglia:
!squash

PR #672, opened 2023-07-10
Fix hidden end screen sharing button

EgorovEI:
Previously, if one-way video is accepted by visitor during screen sharing session, end screen sharing button is hidden on Call screen. This commit fixes it.

MOB-2452

### DIFF
--- a/GliaWidgetsTests/CallVisualizer/VideoCall/VideoCallTests.swift
+++ b/GliaWidgetsTests/CallVisualizer/VideoCall/VideoCallTests.swift
@@ -66,4 +66,31 @@ final class VideoCallTests: XCTestCase {
         headerProps.backButton?.tap()
         XCTAssertEqual(wasExecuted, true)
     }
+
+    func test_endScreenSharingIsVisible() {
+        let viewModel = mockViewModel(with: .started)
+        let props = viewModel.makeProps()
+
+        XCTAssertFalse(props.videoCallViewProps.endScreenShareButtonHidden)
+    }
+
+    func test_endScreenSharingIsHidden() {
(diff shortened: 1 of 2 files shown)


GliaWidgetsTests/CallVisualizer/VideoCall/VideoCallTests.swift
```diff
@@ -66,4 +66,31 @@ final class VideoCallTests: XCTestCase {
         headerProps.backButton?.tap()
         XCTAssertEqual(wasExecuted, true)
     }
+
+    func test_end_screen_sharing_visible() {
```
igorkravchenko, line 70: We do not have written agreement, but previously, at leat during review of my PRs, there was agreement team mates to stick to conventional naming Swift style, with exception to `test_` prefix:
```suggestion
    func test_endScreenSharingIsVisible() {
```

GliaWidgetsTests/CallVisualizer/VideoCall/VideoCallTests.swift
```diff
@@ -66,4 +66,31 @@ final class VideoCallTests: XCTestCase {
...
+        let viewModel = mockViewModel(with: .started)
+        let props = viewModel.makeProps()
+
+        XCTAssertFalse(props.videoCallViewProps.endScreenShareButtonHidden)
+    }
+
+    func test_end_screen_sharing_hidden() {
```
igorkravchenko, line 77:
```suggestion
    func test_endScreenSharingIsHidden() {
```

igorkravchenko (approved):

rasmustautsglia (approved):

PR #676, opened 2023-07-11
Address video not being shown after operator transfer

igorkravchenko:
Address video being hidden when transferred operator makes video request. The fix is in moving 'mediaStream.value' being assigned and 'updateStarted' called in 'defer' statement.

MOB-2438

### DIFF
--- a/GliaWidgets/Sources/ViewModel/Call/Data/Call.swift
+++ b/GliaWidgets/Sources/ViewModel/Call/Data/Call.swift
@@ -234,27 +234,30 @@ class Call {
     private func updateMediaStream<Streamable>(_ mediaStream: ObservableValue<MediaStream<Streamable>>,
                                                with stream: Streamable,
                                                isRemote: Bool) {
+        let mediaStreamValue: MediaStream<Streamable>
+        defer {
+            mediaStream.value = mediaStreamValue
+            updateStarted()
+        }
         if isRemote {
             switch mediaStream.value {
             case .none, .remote:
-                mediaStream.value = .remote(stream)
+                mediaStreamValue = .remote(stream)
             case .local(let local):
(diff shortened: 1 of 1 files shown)


GliaWidgets/Sources/ViewModel/Call/Data/Call.swift
```diff
@@ -234,27 +234,30 @@ class Call {
     private func updateMediaStream<Streamable>(_ mediaStream: ObservableValue<MediaStream<Streamable>>,
                                                with stream: Streamable,
                                                isRemote: Bool) {
+        let mediaStreamValue: MediaStream<Streamable>
+        defer {
```
gersonnoboa, line 238: Interesting, do you know why this fixes it?

gersonnoboa (approved):

GliaWidgets/Sources/ViewModel/Call/Data/Call.swift
igorkravchenko: My only assumption, that this is timing issue for `ObservableValue`. First I discovered "fix" by putting break point that does not suspend execution, but continues evaluation and plays sound - that was enough to get video to show up, then I tried famous `DispatchQueue.main.async` - that also worked, then at last - `defer`.

EgorovEI: Looks not very reliable. Perhaps would be good to make a build by `mater-build` Bitrise workflow and ask other to verify this fix.
Wdyt?

EgorovEI (approved):

GliaWidgets/Sources/ViewModel/Call/Data/Call.swift
igorkravchenko: @EgorovEI could you please assist me with that?

PR #677, opened 2023-07-12
Handle HTML text in GVA messages

rasmustautsglia:
This PR allows the HTML to be rendered in GVA messages. The conversion of string to attributed string happens in decoding for performance and proper layout reasons.

### DIFF
--- a/GliaWidgets/Sources/View/Chat/Message/Content/Text/ChatTextContentView.swift
+++ b/GliaWidgets/Sources/View/Chat/Message/Content/Text/ChatTextContentView.swift
@@ -6,6 +6,11 @@ class ChatTextContentView: BaseView {
         set { setText(newValue) }
     }
 
+    var attributedText: NSAttributedString? {
+        get { return textView.attributedText }
+        set { return setAttributedText(newValue) }
+    }
+
     var accessibilityProperties: AccessibilityProperties {
         get {
             .init(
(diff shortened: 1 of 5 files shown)


GliaWidgets/Sources/Extensions/Data+Extension.swift
```diff
@@ -0,0 +1,17 @@
...
+extension Data {
+    var html2AttributedString: NSAttributedString? {
+        do {
+            return try NSAttributedString(
+                data: self,
+                options: [.documentType: NSAttributedString.DocumentType.html, .characterEncoding: String.Encoding.utf8.rawValue],
+                documentAttributes: nil
+            )
+        } catch {
+            print("error:", error)
```
EgorovEI, line 12: Would be good to have more descriptive prefix, indicating that html-string decoding failed

EgorovEI: And if we don't log it to Kibana, would be better to prevent logging to console in prod env. So `debugPrint` would be preferable

EgorovEI (approved):

GliaWidgets/Sources/Extensions/Data+Extension.swift
rasmustautsglia: fixed [9c3c9a7](https://github.com/salemove/ios-sdk-widgets/pull/677/commits/9c3c9a71f8cead51b40cadc03f7ecf99a111eb4b)

GliaWidgets/Sources/Extensions/Data+Extension.swift
```diff
@@ -0,0 +1,17 @@
+import Foundation
+
+extension Data {
+    var html2AttributedString: NSAttributedString? {
```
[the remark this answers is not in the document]
EgorovEI: I will move it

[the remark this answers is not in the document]
gersonnoboa: It can also be named `htmlToAttributedString`, same with all of the other ones that have `2` on the name

GliaWidgets/Sources/Extensions/Data+Extension.swift
```diff
@@ -0,0 +1,17 @@
...
+        do {
+            return try NSAttributedString(
+                data: self,
+                options: [.documentType: NSAttributedString.DocumentType.html, .characterEncoding: String.Encoding.utf8.rawValue],
```
gersonnoboa: NIT: This line could be made shorter by either adding a new line on each param, or if it becomes too nested, then the options could be generated as a variable before the return and then just use the variable

```
var options: [whatever type] = [
    .documentType:...,
    ...
]

return try NSAttributedString(
    data: self,
    options: options,
    ....
)
````

File: GliaWidgets/Sources/View/Chat/Message/Content/Text/ChatTextContentView.swift [hunk 4]
```diff
@@ -89,19 +94,52 @@ class ChatTextContentView: BaseView {
...
+            constraints += textView.layoutInSuperview(insets: kTextInsets)
         }
+        textView.text = text
+
         textView.accessibilityIdentifier = text
     }
+
+    private func setAttributedText(_ text: NSAttributedString?) {
+        guard let text, !text.string.isEmpty else {
+            textView.removeFromSuperview()
```
gersonnoboa, line 115: This seems like a quite big side effect of sending an empty/nil string though. Like imagine if setting `label.text = ""` would mea that the label gets removed from the superview. I guess it's more expected that this gets checked before, so 

```
if text.string.isEmpty {
    text.removeFromSuperview()
} else {
    text.setAttributedText(...)
}
```

GliaWidgets/Sources/View/Chat/Message/Content/Text/ChatTextContentView.swift
```diff
@@ -89,19 +94,52 @@ class ChatTextContentView: BaseView {
...
+            return
+        }
+
+        if textView.superview == nil {
```
gersonnoboa, line 119: And this also seems like a lot for this method. Don't know, seems a bit strange that setting the attributed text deals with adding/removing stuff from the superview, setting translatesAutoresizing..., etc.

GliaWidgets/Sources/Extensions/Data+Extension.swift
EgorovEI: 4b3ba3b61831c87a2f3124f2c5ea240f45e7d75a

GliaWidgets/Sources/View/Chat/Message/Content/Text/ChatTextContentView.swift
gersonnoboa: I analyzed this file more and it turns out that the textView is a private thing, it's an internal implementation of the content view that the caller is not aware of. Then I guess there's not much to do

gersonnoboa (approved):

igorkravchenko (approved):

EgorovEI:
!squash

PR #679, opened 2023-07-14
Added actions for GVA buttons

EgorovEI:
Added unit tests for GVA button actions

MOB-2378


EgorovEI:
http(s)-based url opening should discussed in universal link perspective, so it might be changed.

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel+GVA.swift
```diff
@@ -0,0 +1,83 @@
...
+private extension String {
+    static let gvaOptionUrlTargetModal = "modal"
+    static let gvaOptionUrlTargetSelf = "self"
+}
+
+extension ChatViewModel {
+    func gvaOptionAction(for option: GvaOption) -> Cmd {
+        // If option contains `url`, then it's URL Button,
+        // otherwise it's Postback Button and option should be sent
+        // to the server as `SingleChoiceOption`
```
gersonnoboa, line 13: This is good information

gersonnoboa (commented):
Is it fine that both actions just return if there is a failure, instead of giving some kind of error?

EgorovEI:
> Is it fine that both actions just return if there is a failure, instead of giving some kind of error?

Sending SingleChoiseOption failure is handled by presenting an alert.

There is only one case with opening urls is not handled, it's deeplinks without proper urlTarget value. In this case nothing happens. Did you mean this one?

gersonnoboa:
@EgorovEI yeah the return in the default and also for the guards. Maybe the guards are not important, but if the return in the default is correct, then all is well.

gersonnoboa (approved):

EgorovEI:
> yeah the return in the default and also for the guards. Maybe the guards are not important, but if the return in the default is correct, then all is well.

Thank you for raising this question. For now we covered all cases we need to handle. Some things that might be improved we will discuss later.

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel+GVA.swift
```diff
@@ -0,0 +1,83 @@
...
+        if let urlString = option.url,
+           let url = URL(string: urlString) {
+            return urlButtonAction(url: url, urlTarget: option.urlTarget)
+        } else {
```
rasmustautsglia, line 17: I would add another case for a broadcast event that logs an error. So if a broadcast event is present regardless of other values, then error should be logged but not presented

igorkravchenko (approved):

rasmustautsglia (approved):

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel+GVA.swift
EgorovEI: Good point! 
I will add presenting an alert indicating that broadcast events are not supported as mentioned in project plan

EgorovEI:
!squash

PR #680, opened 2023-07-17
Adjust layout to match snapshots

igorkravchenko:
Address layout inconsistencies after PureLayout removal.

MOB-2469

### DIFF
--- a/GliaWidgets/Sources/View/Chat/ChatView.swift
+++ b/GliaWidgets/Sources/View/Chat/ChatView.swift
@@ -162,7 +162,17 @@ class ChatView: EngagementView {
         ])
 
         addSubview(messageEntryView)
-        messageEntryViewBottomConstraint = messageEntryView.layoutIn(safeAreaLayoutGuide, edges: .bottom).first
+        let messageEntryInsets = UIEdgeInsets(
+            top: 0,
+            left: 0,
+            bottom: 4.5,
+            right: 0
+        )
+        messageEntryViewBottomConstraint = messageEntryView.layoutIn(
+            safeAreaLayoutGuide,
+            edges: .bottom,
+            insets: messageEntryInsets
+        ).first
         constraints += messageEntryViewBottomConstraint
 
         constraints += messageEntryView.layoutIn(safeAreaLayoutGuide, edges: .horizontal)
--- a/SnapshotTests/BubbleViewLayoutTests.swift
+++ b/SnapshotTests/BubbleViewLayoutTests.swift
@@ -1,13 +1,12 @@
-import AccessibilitySnapshot
 @testable import GliaWidgets
 import SnapshotTesting
 import XCTest
 
-class BubbleViewVoiceOverTests: SnapshotTestCase {
+class BubbleViewLayoutTests: SnapshotTestCase {
     func test_bubble() {
         let bubble = ViewFactory.mock().makeBubbleView()
         bubble.frame = .init(origin: .zero, size: .init(width: 50, height: 50))
         // If shadow will cause failing test locally or on CI, we should disable it.
-        assertSnapshot(matching: bubble, as: .accessibilityImage(precision: SnapshotTestCase.possiblePrecision))
+        assertSnapshot(matching: bubble, as: .image)
     }
 }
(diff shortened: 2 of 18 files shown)


SnapshotTests/BubbleViewLayoutTests.swift
```diff
@@ -0,0 +1,12 @@
+@testable import GliaWidgets
+import SnapshotTesting
+import XCTest
+
+class BubbleViewLayoutTests: SnapshotTestCase {
+    func test_bubble() {
+        let bubble = ViewFactory.mock().makeBubbleView()
+        bubble.frame = .init(origin: .zero, size: .init(width: 50, height: 50))
+        // If shadow will cause failing test locally or on CI, we should disable it.
```
gersonnoboa, line 9: I guess that now it won't happen but still good comment

SnapshotTests/SecureConversationsWelcomeScreenLayoutTests.swift
```diff
@@ -0,0 +1,123 @@
...
+            id: "id-b",
+            style: .messageCenter(theme.secureConversationsWelcome.attachmentListStyle.item),
+            state: .error(.network),
+            removeTapped: .nop
+        )
+    }
+
+    static func headerProps() -> Header.Props {
+        .mock(
+            title: "Secure Conversations",
```
gersonnoboa, line 87: I guess it is irrelevant which title it is, but the title in the widgets for secure conversations is Messaging

gersonnoboa (approved):

SnapshotTests/SnapshotTestCase.swift
```diff
@@ -38,7 +38,7 @@ extension SnapshotTestCase {
         )
     ]
 
-    static let possiblePrecision: Float = 0.99
+    static let possiblePrecision: Float = 1.0
```
rasmustautsglia, line 41: Nice!

rasmustautsglia (approved):

GliaWidgets/Sources/View/Chat/ChatView.swift
```diff
@@ -162,7 +162,17 @@ class ChatView: EngagementView {
         ])
 
         addSubview(messageEntryView)
-        messageEntryViewBottomConstraint = messageEntryView.layoutIn(safeAreaLayoutGuide, edges: .bottom).first
+        let messageEntryInsets = UIEdgeInsets(
+            top: 0,
+            left: 0,
+            bottom: 4.5,
```
EgorovEI, line 168: Will not it cause some tiny diff on further runs? Could we round this value to be sure that there will not be this disturbing tiny diff? Wdyt?

SnapshotTests/SecureConversationsWelcomeScreenLayoutTests.swift
igorkravchenko: Same title is used for VoiceOver tests, so I'd leave it as it is for now, to avoid regenerating new references.

SnapshotTests/ScreenShareViewControllerLayoutTests.swift
```diff
@@ -0,0 +1,26 @@
+@testable import GliaWidgets
+import SnapshotTesting
+import XCTest
+
+class ScreenShareViewControllerLayoutTests: SnapshotTestCase {
+    func testScreenShareViewController() {
```
EgorovEI, line 6: Just to align with other tests
```suggestion
    func test_screenShareViewController() {
```

EgorovEI (approved):

SnapshotTests/ScreenShareViewControllerLayoutTests.swift
igorkravchenko: It makes sense, for sure, but I think we will not benefit from changing it now, since it will require regeneration of references. It was like that in VoiceOver tests, that is why it is kept like that. The purpose of this PR is to keep differences at minimum in layout. I will create separate ticket for renaming VoiceOver test files, because some of them are missing `VoiceOverTests` suffix and address this suggestion as well.

GliaWidgets/Sources/View/Chat/ChatView.swift
igorkravchenko: Let's see how CI will handle it. I will make adjustment if this will cause issues.

SnapshotTests/ScreenShareViewControllerLayoutTests.swift
EgorovEI: Oh, I see. You are right

PR #682, opened 2023-07-17
Unsupported broadcast events alert action

EgorovEI:
Since mobile does not supported broadcast events, we need to show alert by pressing corresponding GVA message

snake-cased "gva_not_supported" localization key was defined in the ticket

MOB-2388

### DIFF
--- a/GliaWidgets/Resources/en.lproj/Localizable.strings
+++ b/GliaWidgets/Resources/en.lproj/Localizable.strings
@@ -263,3 +263,5 @@
 "callVisualizer.screenSharing.accessibility.messageHint" = "Message label";
 "callVisualizer.screenSharing.accessibility.buttonLabel" = "End screen sharing";
 "callVisualizer.screenSharing.accessibility.buttonHint" = "Ends screen sharing";
+
+"gva_not_supported" = "This action is not currently supported on mobile";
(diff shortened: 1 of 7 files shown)


rasmustautsglia (approved):

EgorovEI:
!squash

GliaWidgets/Resources/en.lproj/Localizable.strings
```diff
@@ -263,3 +263,5 @@
 "callVisualizer.screenSharing.accessibility.messageHint" = "Message label";
 "callVisualizer.screenSharing.accessibility.buttonLabel" = "End screen sharing";
 "callVisualizer.screenSharing.accessibility.buttonHint" = "Ends screen sharing";
+
+"gva_not_supported" = "This action is not currently supported on mobile";
```
gersonnoboa, line 267: If you want to use dots, you can. Don't worry to much at the moment. What is important is that the words in the key are the same as in Android for now.

gersonnoboa (approved):

GliaWidgets/Resources/en.lproj/Localizable.strings
EgorovEI: Let's leave as is for now

PR #686, opened 2023-07-19
Fix increment project version in Bitrise

gersonnoboa:
With git flow, we need to ensure that we are on the development branch, so we must do git checkout development. However, if the workflows are executed in master, Bitrise will not know about the development branch, so a git fetch before is needed.

MOB-2474


igorkravchenko (approved):

EgorovEI (approved):

gersonnoboa:
!merge

PR #689, opened 2023-07-19
Update CoreSDK 1.0.5 dependency

yurii-glia:


rasmustautsglia (approved):

EgorovEI:
Do we need to update also the CoreSDK in Package.swift?

gersonnoboa:
When the CI does this, there's more files there. Should those be here too? For example as Egor said, the Package.swift, also the static version that is set in a Swift file, etc.

EgorovEI:
> 

Static version in swift file changes when CI increments WidgetsSDK version, not CoreSDK

yurii-glia:
Core SDK version incrementation happens in this pr - https://github.com/salemove/ios-sdk-widgets/pull/692

PR #695, opened 2023-07-24
Add Quick Reply button to RemoteConfiguration

EgorovEI:
Also fixed retrieving json configuration files from app bundle

MOB-2397


yurii-glia (approved):

GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration+Chat.swift
```diff
@@ -23,6 +23,7 @@ extension RemoteConfiguration {
 
     struct Gva: Codable {
         let persistentButton: GvaPersistentButton?
+        let quickReplyButton: Button?
```
gersonnoboa, line 26: I appreciate that you have called it quick reply and not something like qrButton :D

gersonnoboa (approved):

PR #696, opened 2023-07-24
Introduce snapshot tests for dynamic type font

igorkravchenko:
Add initial snapshot tests for dynamic type fonts. Note that these tests reveal that some UI does not take into account dynamic font type, so appropriate tickets are to be created.

MOB-2476

### DIFF
--- a/SnapshotTests/VisitorCodeViewControllerDynamicTypeFontTests.swift
+++ b/SnapshotTests/VisitorCodeViewControllerDynamicTypeFontTests.swift
@@ -1,0 +1,108 @@
+@testable import GliaWidgets
+import SnapshotTesting
+import XCTest
+
+// swiftlint:disable type_name
+final class VisitorCodeViewControllerDynamicTypeFontTests: SnapshotTestCase {
+    func testVisitorCodeAlertWhenLoading() {
+        let props: CallVisualizer.VisitorCodeViewController.Props = .init(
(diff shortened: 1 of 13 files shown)


SnapshotTests/VisitorCodeViewControllerDynamicTypeFontTests.swift
```diff
@@ -0,0 +1,114 @@
+@testable import GliaWidgets
+import SnapshotTesting
+import XCTest
+
+// swiftlint:disable type_name
+class VisitorCodeViewControllerDynamicTypeFontTests: SnapshotTestCase {
+    static let extra3LargeFontStrategy = Snapshotting<UIViewController, UIImage>.image(
```
EgorovEI, line 7: Could it be moved to `SnapshotTestCase`, because it's duplicated in each new test class?

EgorovEI (approved):

yurii-glia (approved):

SnapshotTests/VisitorCodeViewControllerDynamicTypeFontTests.swift
igorkravchenko: Good idea. Introduced dedicated strategies 40a9070ea6ec0454458589febfe02a2e1697477c.

igorkravchenko:
!squash

PR #697, opened 2023-07-25
Fix custom cards handling

yurii-glia:
SDK differentiate custom card and response card based on metadata property. 
If the type is single choice card and metadata is nil, SDK should draw a message as a response card.
If the type is single choice card and metadata is not nil, SDK should pass metadata to renderer and render custom card.

MOB-2342


EgorovEI (approved):

igorkravchenko (approved):
Looks good to me 👍

PR #702, opened 2023-07-31
Add missing acc. identifiers and custom segmented control

igorkravchenko:
Add more accessibility identifiers and replace regular UISegmentedControl with custom implementation to overcome Voice Over (screen reader) issue, where views order is broken, if such control is used in section header.

MOB-1100


yurii-glia (approved):

EgorovEI (approved):

igorkravchenko:
!squash

PR #704, opened 2023-08-03
Add translation layer from new strings to old string

gersonnoboa:
In the new list of strings unified between iOS and Android, there are some strings that have changed their name. However, the string itself is the same. Thus, there's a layer that converts between these new strings and the old strings. These aren't used anywhere yet. However, subsequent pull requests will put them to use. The idea is to have those old strings deprecated and only use the new ones, and then delete this layer in a year, just keeping the new ones in use.

Also some strings were changed already if the change was really small and new strings that didn't have an equivalent before were added.

MOB-2492


EgorovEI (approved):
What is the `Updated.strings` file used for?

gersonnoboa:
@EgorovEI that is a mistake, removed ef07ffc66a928b68b6e108bd8968e0472ab5645e

igorkravchenko (approved):

rasmustautsglia (approved):

gersonnoboa:
Bitrise is failing because with the changed strings, the snapshot tests fail. Created a PR for that https://github.com/salemove/ios-widgets-snapshots/pull/28

gersonnoboa:
!squash

gersonnoboa:
!merge

PR #708, opened 2023-08-09
Add Snapshot tests for GVA response text

rasmustautsglia:
This PR adds snapshot tests for GVA response text.

MOB-2509

### DIFF
--- a/GliaWidgets/Sources/View/Chat/Message/ChatMessageView.swift
+++ b/GliaWidgets/Sources/View/Chat/Message/ChatMessageView.swift
@@ -63,10 +63,6 @@ class ChatMessageView: BaseView {
             )
             contentView.attributedText = text
             contentView.linkTapped = { [weak self] in self?.linkTapped?($0) }
-            contentView.accessibilityProperties = .init(
-                label: accProperties.label,
-                value: accProperties.value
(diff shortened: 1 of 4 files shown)


GliaWidgets/Sources/View/Chat/Message/ChatMessageView.swift
```diff
@@ -63,10 +63,6 @@ class ChatMessageView: BaseView {
             )
             contentView.attributedText = text
             contentView.linkTapped = { [weak self] in self?.linkTapped?($0) }
-            contentView.accessibilityProperties = .init(
```
gersonnoboa, line 66: Is there a reason this was removed?

gersonnoboa (approved):

GliaWidgets/Sources/View/Chat/Message/ChatMessageView.swift
rasmustautsglia: This implementation is applicable only to a normal message, and not to an attributed message. Removing that was essential for the accessibility to be displayed properly

rasmustautsglia:
!squash

yurii-glia (approved):

PR #718, opened 2023-08-11
Deprecate old strings resources

gersonnoboa:
Instead of trying to reuse some old strings, deprecate some other strings, and have them coexist in the same file, I have decided to deprecate the `L10n.swift` class altogether, and create a new one called `Localization.swift`. This allows me to remove that horrible abbreviated name, and also make a clear distinction between old and new. With this, all new strings will just go to the `Localizable.strings` as always, and the `Deprecated.strings` ones shouldn't be used anymore.

This was done by adding two `.stencil` files: one that handles adding the `@deprecated` line to the old file, and a new one that handles ignoring all strings that start with `android`. This is because on the new way to do strings, the Android-specific ones will start with `android.`. We could remove these from the file, but this would be manual, some could be missed, and at the moment of this PR, we haven't yet determined how we are going to sync these strings with the strings in the Glia Hub.

Finally, the new `Localization.swift` file is `internal`, as opposed to `L10n.swift` which is public. This is because these strings have changed from being the main way to show strings in the widgets, to being a fallback. Now, the Glia Hub will be the main way to pass strings to the widgets, so there's no need to shoot ourselves in the foot, making this public, and then have to worry about deprecations, as we have to do with `L10n.swift`.

With this, there's no need to have `L10n+BackwardsCompatibility.swift`, as the new strings and the old ones are separated entities. However, I still need to do a layer that goes between the strings and the rest of the code, for when the strings have functionality that is not present anymore (example, placeholders for values that now just have the direct value in the string).

In a year, we can easily remove `L10n.swift` and the associated `.stencil` file.

MOB-2529


GliaWidgets/Localization.swift
```diff
@@ -0,0 +1,573 @@
...
+
+// swiftlint:disable superfluous_disable_command file_length implicit_return prefer_self_in_static_references
+
+// MARK: - Strings
+
+// swiftlint:disable explicit_type_interface function_parameter_count identifier_name line_length
+// swiftlint:disable nesting type_body_length type_name vertical_whitespace_opening_braces
+internal enum Localization {
+  internal enum Alert {
+    internal enum Action {
```
rasmustautsglia, line 14: Is there a way we could potentially make the indentation bigger to conform to swift standards? As this file can be generated from the command line, it's not a big issue but it would just look better

swiftgen.yml
```diff
@@ -1,11 +1,18 @@
...
-        publicAccess: true
-        enumName: L10n
+  - inputs: GliaWidgets/Resources/en.lproj/Localizable.strings
+    outputs:
+      - templatePath: swiftgen-strings.stencil
+        output: GliaWidgets/Localization.swift
+        params:
+          publicAccess: false
+          enumName: Localization
+  - inputs: GliaWidgets/Resources/en.lproj/Deprecated.strings
```
yurii-glia, line 9: should deprecated file still be in the `swiftgen.yml`? i supposed if it is deprecated, so no change will come to this file and we don't need to touch it.

EgorovEI (approved):

GliaWidgets/Localization.swift
gersonnoboa: I could take a look, probably there's a parameter.

gersonnoboa: I investigated, apparently [there's no way of doing this](https://github.com/SwiftGen/SwiftGen/issues/390). In theory, I could modify the stencil so that it has four spaces. This would deviate us greatly from the official one. Wouldn't be the worst but we would have to decide if it's worth it to deviate so much from the original one. A benefit of being close to the original is that if they update it, we can just add our few changes easily. What do you think would be best?

yurii-glia (approved):

igorkravchenko (approved):

gersonnoboa:
!merge

PR #720, opened 2023-08-16
Replace old strings with new strings in some views

gersonnoboa:
Each commit has its own description. This task is quite big so I am doing it in parts. This task only deals with strings that have a direct replacement in the new Localization.swift file. Strings that have functionality, such as `{parameters}` inside, will not be dealt with in this task. There's going to be several PRs associated to this task, and all will be merged to the `custom-locales` branch. When the work there replacing strings has been completed, it can be merged to the development branch to avoid the situation where the feature branch has months of changes and merging it is a mess.


rasmustautsglia (approved):
Looks good to me!

EgorovEI (approved):

gersonnoboa:
!merge

PR #724, opened 2023-08-17
Discard duplicate messages delivered by sockets

igorkravchenko:
When messages from history are loaded, same messages delivered from sockets are to be discarded to avoid duplicates.

MOB-2542

### DIFF
--- a/GliaWidgets/Sources/ViewModel/Chat/Data/ChatMessage.swift
+++ b/GliaWidgets/Sources/ViewModel/Chat/Data/ChatMessage.swift
@@ -25,7 +25,8 @@ enum ChatMessageSender: Int, Codable {
 }
 
 class ChatMessage: Codable {
-    let id: String
+    typealias MessageId = String
+    let id: MessageId
     var queueID: String?
     let `operator`: ChatOperator?
     let sender: ChatMessageSender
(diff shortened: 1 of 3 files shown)


GliaWidgets/Sources/ViewModel/Chat/Data/ChatMessage.swift
```diff
@@ -25,7 +25,8 @@ enum ChatMessageSender: Int, Codable {
 }
 
 class ChatMessage: Codable {
-    let id: String
+    typealias MessageId = String
```
rasmustautsglia, line 28: what is the purpose of this typealias? Can we remove it?

yurii-glia (commented):
@igorkravchenko  to avoid having business logic on Widgets side perhaps it makes sense to do this logic on Core SDK side, would it make sense and possible to do?

GliaWidgets/Sources/ViewModel/Chat/Data/ChatMessage.swift
EgorovEI: This typealias is like a hint for devs that `historyMessageIds` set should store only message ids, because compiler still allows you to store any string there.

igorkravchenko: @rasmustautsglia Egor is making the exact point. It is a hint that it's not just a regular string, but message id. Why do you think it has to be removed?

igorkravchenko:
> @igorkravchenko to avoid having business logic on Widgets side perhaps it makes sense to do this logic on Core SDK side, would it make sense and possible to do?

@yurii-glia On Core SDK side we do not have currently interface that would combine getting messages from history and from sockets at once. This is attempted in Widgets for specific cases.

yurii-glia (approved):

rasmustautsglia (approved):

PR #725, opened 2023-08-21
Replace old strings with new strings in some views

gersonnoboa:
This time, strings in screen sharing, alert, call, and survey mocks were changed. Also a string that was previously missing was added.

MOB-2538

### DIFF
--- a/GliaWidgets/Sources/ViewController/Survey/SurveyViewController.View.swift
+++ b/GliaWidgets/Sources/ViewController/Survey/SurveyViewController.View.swift
@@ -28,7 +28,7 @@ extension Survey {
             $0.layer.shadowRadius = 8
         }
         let cancelButton = UIButton(type: .custom).make {
-            $0.setTitle(L10n.Survey.Action.cancel, for: .normal)
+            $0.setTitle(Localization.General.cancel, for: .normal)
             $0.titleLabel?.numberOfLines = 0
             // Using `byWordWrapping` prevents button text
             // from getting truncated, for example for large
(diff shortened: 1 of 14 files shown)


GliaWidgets/Sources/ViewController/Survey/SurveyViewController.View.swift
```diff
@@ -189,7 +189,7 @@ extension NSAttributedString {
         if isRequired {
             mutableString.append(
                 .init(
-                    string: L10n.Survey.Question.Title.asterisk,
+                    string: "*",
```
yurii-glia, line 192: would we like to prevent customization for required fields marker? I mean if integrator would like to change asterisk symbol to "required" label

gersonnoboa: What could be done is to expose it through the theme. But we decided with Karl to remove it because an asterisk on its own is not localizable, and also this one had some displaying logic because it wasn't only the asterisk, but an asterisk with a space. Would exposing it through the theme be acceptable?

igorkravchenko (approved):

GliaWidgets/Sources/ViewController/Survey/SurveyViewController.View.swift
yurii-glia: I'm not a fun of moving this kind of customization to theme (i suppose to have only ui in the theme structure), but yes, perhaps in future we can create some separate structure for this kind of customization.

yurii-glia (approved):

rasmustautsglia (approved):

EgorovEI (approved):

PR #728, opened 2023-08-22
Gallery Card not rendered correctly when image url is invalid

rasmustautsglia:
It could happen that a GVA gallery card image URL is not a valid URL. This PR allows the gallery card to be rendered regardless, compared to the current implementation where gallery card decoding fails.

MOB-2560

### DIFF
--- a/GliaWidgets/Sources/View/Chat/ChatView.GvaGallery.swift
+++ b/GliaWidgets/Sources/View/Chat/ChatView.GvaGallery.swift
@@ -49,7 +49,7 @@ extension ChatView {
                 )
             } ?? []
             var image: GvaGalleryCardView.Props.Image?
-            if let url = card.imageUrl {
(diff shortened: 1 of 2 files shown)


EgorovEI:
Previous implementation did the same, no? If `imageUrl` string field in the json is invalid url, then `imageUrl` property will be nil and UIImage will not be added to Gallery card. Isn't it?

rasmustautsglia:
No, with the previous implementation it didn't recognize it as gallery card, but normal text messages since it couldn't decode it to GalleryCard

EgorovEI:
> No, with the previous implementation it didn't recognize it as gallery card, but as normal text messages since it couldn't decode it to GalleryCard

Interesting. But ok.

EgorovEI (approved):

yurii-glia (approved):

PR #737, opened 2023-08-25
Fix ChatView TableView crash when scrolling to bottom

rasmustautsglia:
This PR resolves a race condition that caused a crash when scrolling to the bottom. The issue stemmed from outdated section row count data, leading the system to believe a section had rows when, in actuality, it didn't.

This PR also removes the method `scrollToTop` as it wasn't used anywhere

MOB-2578


rasmustautsglia:
!squash

gersonnoboa (approved):

EgorovEI (approved):

PR #742, opened 2023-08-29
Show Quick Replies for Secure Conversations

EgorovEI:
This PR allows Quick Reply buttons to be displayed from Secure Conversations history if quick reply button is the last message.

MOB-2588


EgorovEI:
!squash

yurii-glia (approved):

rasmustautsglia (approved):

PR #747, opened 2023-09-01
Improve URLs handling in text messages

EgorovEI:
Before SDK opened `http(s)` links in `SafaryViewController`, `tel:` and `mailto:` links using `UIApplication.open` method. Deep links were skipped at all. Now any type of URL is opened using `UIApplication.open method` to align behaviour with GVA feature and Android SDK.

MOB-2605

### DIFF
--- a/GliaWidgets/SecureConversations/ChatTranscript/SecureConversations.TranscriptModel.swift
+++ b/GliaWidgets/SecureConversations/ChatTranscript/SecureConversations.TranscriptModel.swift
@@ -11,7 +11,6 @@ extension SecureConversations {
 
         enum DelegateEvent {
             case showFile(LocalFile)
-            case openLink(URL)
             case showAlertAsView(
                 MessageAlertConfiguration,
                 accessibilityIdentifier: String?,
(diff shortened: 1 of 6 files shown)


rasmustautsglia (approved):

GliaWidgets/SecureConversations/ChatTranscript/SecureConversations.TranscriptModel.swift
```diff
@@ -400,22 +399,8 @@ extension SecureConversations.TranscriptModel {
...
-            environment.uiApplication.open(url)
-
-        case "http",
-            "https":
-            delegate?(.openLink(url))
-
-        default:
-            return
-        }
+        guard environment.uiApplication.canOpenURL(url) else { return }
```
gersonnoboa, line 402: What if it cannot open? Nothing will happen? Perhaps it's not that bad of an experience, kind of like just pressing regular text.

igorkravchenko (approved):
Looks good to me.

GliaWidgets/SecureConversations/ChatTranscript/SecureConversations.TranscriptModel.swift
EgorovEI: > What if it cannot open? Nothing will happen?

You are correct. Nothing will happen.
Do you think we need to show an error or/and send it to logger, in case when iOS can not open custom URL scheme, because of there is no app associated with this scheme?
Previously we just skipped handling of URLs with custom schemes at all.

gersonnoboa (approved):

PR #753, opened 2023-09-04
Fix failed unit test visitorAppDefaultLocale error

yurii-glia:


gersonnoboa (approved):
However the title is too long so it's overflowing into the description

igorkravchenko (approved):

PR #754, opened 2023-09-05
Remove unnecessary accessibility hint in Call Visualizer

gersonnoboa:
This string is not needed as a hint because we are working with a label, so the text of the label itself is the hint.

MOB-2614


gersonnoboa:
Equivalent snapshot tests: https://github.com/salemove/ios-widgets-snapshots/pull/40

EgorovEI (approved):
It's really convenient to use linked snapshot test PR. Thanks!

rasmustautsglia (approved):

PR #755, opened 2023-09-06
Remove unnecessary accessibility hint in Call Visualizer

gersonnoboa:
This string is not needed as a hint because we are working with a label, so the text of the label itself is the hint. Same PR has been done for custom-locales branch, but without this PR, all ios-sdk-widgets snapshot tests will fail.

MOB-2614


gersonnoboa:
Related to this https://github.com/salemove/ios-sdk-widgets/pull/754, just for `development` branch this time.

igorkravchenko (approved):

EgorovEI (approved):

PR #758, opened 2023-09-06
Make interface for enqueueing with multiple queue

yurii-glia:
These changes will allow integrators fetching available queues without the needs to configure SDK with specific queue id.

Deprecated:
- Interface for configuration SDK with queue ID

Introduced:
- Configuration method without queue id
- Starting engagement with queue ids

MOB-2594


GliaWidgets/Public/Glia/Glia+StartEngagement.swift
```diff
@@ -22,17 +23,29 @@ extension Glia {
...
+        guard !queueIds.isEmpty || interactor?.queueIds.isEmpty == false else { throw GliaError.startEngagementWithNoQueueIds }
         guard engagement == .none else { throw GliaError.engagementExists }
         guard let interactor = self.interactor else { throw GliaError.sdkIsNotConfigured }
         if let engagement = environment.coreSdk.getCurrentEngagement(),
             engagement.source == .callVisualizer {
             throw GliaError.callVisualizerEngagementExists
         }
 
+        // This check is needed for integrators who uses old interface
+        // and pass queue identifier throught `configuration` function,
```
gersonnoboa, line 42: Through

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
```diff
@@ -55,53 +69,7 @@ extension Glia {
...
-            }
-        } catch GliaError.configuringDuringEngagementIsNotAllowed {
-            try completion()
-        }
-    }
-
-    // MARK: - Private
-
-    private func startRootCoordinator(
+    func startRootCoordinator(
```
gersonnoboa, line 72: Why is the root coordinator now internal?

GliaWidgets/Public/Glia/Glia.swift
```diff
@@ -266,8 +264,8 @@ public class Glia {
 
 // MARK: - Private
 
-private extension Glia {
-    func startObservingInteractorEvents() {
+extension Glia {
```
gersonnoboa, line 267: Why it's not private?

GliaWidgets/Public/GliaError.swift
```diff
@@ -6,4 +6,5 @@ public enum GliaError: Error {
     case callVisualizerEngagementExists
     case configuringDuringEngagementIsNotAllowed
     case clearingVisitorSessionDuringEngagementIsNotAllowed
+    case startEngagementWithNoQueueIds
```
gersonnoboa, line 9: I suggest changing it to something like the error above. Starting engagement with no queue ids is not allowed. This error sounds like we are telling them to start the engagement without queue ids.

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
```diff
@@ -22,17 +23,29 @@ extension Glia {
...
         if let engagement = environment.coreSdk.getCurrentEngagement(),
             engagement.source == .callVisualizer {
             throw GliaError.callVisualizerEngagementExists
         }
 
+        // This check is needed for integrators who uses old interface
+        // and pass queue identifier throught `configuration` function,
+        // but would not pass queue ids in this method, so SDK would not override
+        // existed queue id.
+        if !queueIds.isEmpty {
```
gersonnoboa, line 45: To be honest I don't really like that when it's time to remove the deprecated methods, we need to probably change this code and the guard. Wouldn't it be easier to just have like, the whole deprecated method works on its own without assumptions, and this one works as if the old one doesn't exist at all? Otherwise it will be like "this method works in X way, unless you are using deprecated methods, in which case it works in Y way". And the behavior of startEngagement vs configure + start is already very confusing for integrators. This has the pontential of making it worse.

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
```diff
@@ -22,17 +23,29 @@ extension Glia {
...
     public func startEngagement(
         engagementKind: EngagementKind,
+        in queueIds: [String],
         theme: Theme = Theme(),
         features: Features = .all,
         sceneProvider: SceneProvider? = nil
     ) throws {
+        // `interactor?.queueIds.isEmpty == false` statement is needed for integrators who uses old interface
+        // and pass queue identifier throught `configuration` function.
+        guard !queueIds.isEmpty || interactor?.queueIds.isEmpty == false else { throw GliaError.startEngagementWithNoQueueIds }
```
EgorovEI, line 33: Looks like this line is better to place after `guard let interactor = self.interactor else { throw GliaError.sdkIsNotConfigured }` line. 
What do you think?

GliaWidgets/Sources/Interactor/Interactor.swift
```diff
@@ -155,18 +155,19 @@ extension Interactor {
...
                 coreSdkVisitorContext,
                 // shouldCloseAllQueues is `true` by default core sdk,
                 // here it is passed explicitly
                 true,
                 mediaType,
                 options
-            ) { [weak self] queueTicket, error in
-                if let error = error {
+            ) { [weak self] result in
+                switch result {
```
EgorovEI, line 166: `Result` 🔥🔥🔥

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
EgorovEI: Agree with Gerson. It must be overcomplicated for integrator to cope with these logic.

GliaWidgets/Public/Glia/Glia.swift
yurii-glia: because another file (Glia.Deprecated.swift) uses it, so we need to make it available for internal usage.

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
yurii-glia: correct
image

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
yurii-glia: @gersonnoboa @EgorovEI 
The main reason of this code is that we changed type for `queueId` inside of Interactor and requires to have queue id(s) on start engagement step. Old configure method expects queueId in signature, however, a new interface doesn't expect it in configure, however, it requires for `startEngagement`.
To keep backward compatibility we need to handle
- queue id passed through `configure` method - old approach
- queue id not passed through `configure` method - new approach
- queue id not passed through `startEngagement` but in this case SDK should not trying start engagement without queue id at all
- queue id passed through `startEngagement` and in this case SDK should initiate a new engagement with these queue id(s) only

So, this not so clear checks needed because of backward compatibility and avoid possible mistakes during integrations.
Would it make sense or would you like to have a call and have a discussion to keep backward compatibility and solve this issue?

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
yurii-glia: I prefer to validate passed values before doing anything with SDK and checking configuration.
Could you please clarify your point and tell why this validation is better after configuration check?

GliaWidgets/Public/GliaError.swift
yurii-glia: fixed - afbd35fc13f2d73f62a6ad17e2cc0d51c1624063

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
yurii-glia: fixed - 2d0861642be424fd5052a06d6b8672f002383868

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
EgorovEI: If it will simplify the process of upgrading SDK and will not confuse integrators, than we can leave it as is. But I would have a call to understand this part better to minimize possible breaks that I (or another) can bring to this code.

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
EgorovEI: My thoughts were not about priority of things to validate, but about interactor instance optionality. 
Your point sound reasonable, so let's omit my comment.

igorkravchenko (approved):

rasmustautsglia (approved):

EgorovEI (approved):

PR #761, opened 2023-09-08
Fix for SingleChoiceResponse message duplications

EgorovEI:
With introduction of visitor messages delivered by socket we are now able to receive messages from web and other devices simultaneously. For regular chat messages all changes in Widgets were done accordingly, but other functionality that involves visitor send message request-response may not take it to account, such as response and custom cards. This commit fixes duplications of sent SingleChoiceResponse messages coming from API response and sockets.

Also `sendSelectedOptionValue` internal interface was removed, because now we use `sendMessageWithPayload` one

MOB-2624


gersonnoboa (approved):
Just the SwiftLint issue

igorkravchenko (approved):

rasmustautsglia (approved):

PR #762, opened 2023-09-11
Make chat call upgrade to audio view ADA compliant

rasmustautsglia:
This PR fixes the issue that the time is partially hidden under the view's bounds.

Snapshot reference PR: https://github.com/salemove/ios-widgets-snapshots/pull/43

MOB-2484


rasmustautsglia:
Snapshots references will be updated too

EgorovEI (approved):

PR #767, opened 2023-09-13
Introduce string providing mechanism

gersonnoboa:
The proposal is that the Core SDK will return to us a dictionary with the keys and values just as they come from the backend. This is also exactly how the strings are saved in the `Localizable.strings` file, accessed through the `Localization` enum. When the Core SDK is configured and is able to provide the dictionary, we save it to a StringProviding struct. When a string is accessed through the use of `Localization`, then `Localization` knows what is the key of the string, and also specifies a fallback in case reading from the file fails. So with this same key, we can go to the StringProviding, consult the dictionary, and if the string is not there, use the file fallback, and if not use the string fallback. This means that with this simple PR, once the Core SDK provides the dictionary, we support custom locales for the whole Widgets SDK with minimal effort. The alternative is to create a model which we need to keep in sync between Core and Widgets, which seems like a big mess.

I have consulted with Karl, and this is also how it's going to be done in Android.

MOB-2282

### DIFF
--- a/GliaWidgets/Localization+StringProviding.swift
+++ b/GliaWidgets/Localization+StringProviding.swift
@@ -1,0 +1,22 @@
+import Foundation
+
+extension Localization {
+    static func tr(
+        _ table: String,
+        _ key: String,
+        _ args: CVarArg...,
+        fallback value: String,
+        stringProviding: StringProviding? = Glia.sharedInstance.stringProviding,
+        bundleManaging: BundleManaging = .live
+    ) -> String {
+        guard
+            let stringProviding,
+            let remoteString = stringProviding.getRemoteString(key)
+        else {
+            let format = bundleManaging.current().localizedString(forKey: key, value: value, table: table)
+            return String(format: format, locale: Locale.current, arguments: args)
+        }
+
+        return remoteString
+    }
+}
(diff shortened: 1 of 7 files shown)


GliaWidgets/Localization+StringProviding.swift
```diff
@@ -0,0 +1,31 @@
...
+        else {
+            let format = BundleToken.bundle.localizedString(forKey: key, value: value, table: table)
+            return String(format: format, locale: Locale.current, arguments: args)
+        }
+
+        return remoteString
+    }
+}
+
+private final class BundleToken {
```
yurii-glia, line 23: I suppose we can replace this with existed [one](https://github.com/salemove/ios-sdk-widgets/blob/71d7bc71b5827bcfc04d5c5f0b1826c616204f4a/GliaWidgets/Sources/BundleManaging/BundleManaging.swift#L3).

yurii-glia (approved):

igorkravchenko (approved):

EgorovEI (approved):

GliaWidgets/Localization+StringProviding.swift
gersonnoboa: Indeed, fixed da02274a862e6161eaf7631d5007ca19b6067237

gersonnoboa:
!squash

PR #778, opened 2023-09-21
Add landscape snapshot tests

rasmustautsglia:
This PR adds snapshot tests in landscape mode for regular and large font types.

Snapshot PR: https://github.com/salemove/ios-widgets-snapshots/pull/49

MOB-2683

PR #780, opened 2023-09-22
Add missing accessibility identifier and introduce new view modifier

igorkravchenko:
After migration to SwiftUI, Secure Conversation’s Confirmation header close button is missing 'header_close_button' identifier, which results in acceptance tests failure. Since we are constrained to iOS 13 deployment target, it made sense to add dedicated view modifier to resolve between depreciated and new methods for providing accessibility identifier to avoid deprecation warnings. This modifier will become obsolete once we bump up deployment target to iOS 14.

MOB-2694


EgorovEI (approved):
SwiftLint warnings will be fixed in this [PR](https://github.com/salemove/ios-sdk-widgets/pull/778)

rasmustautsglia (approved):

PR #782, opened 2023-09-25
#F Get company name from custom locales

EgorovEI:
In this commit, support for adding a company name from custom locales is added.
However, if the integrator has set this name through the theme, that one is used
instead of the custom locales. Also, in case the custom locale doesn't have a
string for the company name, but the integrator has specified a company name
through the configuration of the SDK, then that is used. Finally, if neither of
the previous possibilies have been used, then the local fallbacks are used.
Together, the first and the last point mean that there are no backwards
compatibility issues, as if the integrator hasn't set any name, still the
local fallback will be displayed, which is "CompanyName". If they have, then
that is used immediately without caring about configuration or custom locales.

MOB-2686

### DIFF
--- a/GliaWidgets.xcodeproj/project.pbxproj
+++ b/GliaWidgets.xcodeproj/project.pbxproj
@@ -1690,7 +1690,6 @@
 			isa = PBXGroup;
 			children = (
 				3146C9412AB1850A0047D8CC /* Resources */,
-				7552DFB22A6FBC6E0093519B /* CoreSdk */,
 				7552DFAF2A6FB37E0093519B /* ChatMessage */,
 				846A5C3729D18D220049B29F /* ScreenShareHandler */,
 				AFEF5C7229929A73005C3D8D /* SecureConversations */,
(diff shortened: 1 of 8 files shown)


EgorovEI:
!squash

EgorovEI:
!squash

GliaWidgets/Public/Configuration/Configuration.swift
```diff
@@ -16,7 +16,7 @@ public struct Configuration {
     /// Controls the visibility of the "Powered by" text and image.
     public let isWhiteLabelApp: Bool
     /// Company name. Appears during connection with operator.
-    public let companyName: String
+    public var companyName: String
```
yurii-glia, line 19: do we need to have it as var and change after configuration sdk?

igorkravchenko (approved):

yurii-glia (approved):

EgorovEI:
!squash

rasmustautsglia (approved):

PR #783, opened 2023-09-25
#I Improve snapshot test

rasmustautsglia:
This PR improves the way snapshot tests are executed. By creating an
extension to UIViewController, we can now create snapshot tests with
less code, time, and effort. And, it is much more readable. Since our
usages of snapshot tests are very limited (for now), this approach serves
those purposes very well.

This commit also updates the following:
1. Naming of files - some voiceover test files did not have a proper name.
2. Naming image references - voiceover image references didn't have a 
correct naming, leaving out the dimensions of the view. This PR fixes that
4. AccessibilityImage tests didn't have frame that was aligned with layout
and extra3LargeFont tests
5. Fixes the issue, that some landscape images came out portrait before.

Snapshots PR: https://github.com/salemove/ios-widgets-snapshots/pull/50

MOB-2696


SnapshotTests/extensions/UIViewController+Extensions.swift
```diff
@@ -0,0 +1,66 @@
...
+        line: UInt = #line
+    ) {
+        let snapshotting: Snapshotting<UIViewController, UIImage>
+        switch mode {
+        case .accessibilityImage(let precision):
+            snapshotting = .accessibilityImage(precision: precision)
+        case .image:
+            snapshotting = orientation == .portrait ? .image : .imageLandscape
+        case .extra3LargeFont:
+            snapshotting = orientation == .portrait ? .extra3LargeFontStrategy : .extra3LargeFontStrategyLandscape
```
EgorovEI, line 21: What is the `precision` value for `.image` and `.extra3LargeFont` modes?

rasmustautsglia: I believe it is 1.0 by default

rasmustautsglia: Since our custom precision has move up to 1.0 too, we can remove that from the associated value

EgorovEI (approved):
Looks great to me. More convenient

igorkravchenko (commented):
As we have discussed earlier, in my opinion, extending UIViewController to get snapshot references, makes all snapshots very specific - meaning that we will be forced to create same extension for UIView, SwiftUI.View etc, over and over. I think we should instead extend Snapshotting or SnapshotTestCase to keep snapshot `matching value` generic as it is right now in ` assertSnapshot<Value, Format>` and get desired behaviour via method overload.

yurii-glia (approved):

rasmustautsglia:
Please re-review as the complete update has been added

EgorovEI (approved):

igorkravchenko (commented):
I think `assertSnapshot` in more appropriate method name than `testSnapshot`.

rasmustautsglia:
> I think assertSnapshot in more appropriate method name than testSnapshot.

fixed [d23621c](https://github.com/salemove/ios-sdk-widgets/pull/783/commits/d23621ce99cfa4cb6cbc2af26fff2bf863d39e55)

yurii-glia (approved):

PR #793, opened 2023-10-10
#F Change priority of company name algorithm

gersonnoboa:
Before, the UiTheme took the highest priority in the algorithm to determine the company name. Now the remote company name has the highest priority, with the added constraint that it cannot be empty, as this will be the default value on the default locale. In case it is empty, the UiTheme will come afterwards, then the SDK configuration, and finally the local fallback.

MOB-2733

### DIFF
--- a/GliaWidgets/Public/Glia/Glia+StartEngagement.swift
+++ b/GliaWidgets/Public/Glia/Glia+StartEngagement.swift
@@ -48,12 +48,12 @@ extension Glia {
 
         theme.chat.connect.queue.firstText = companyName(
             using: interactor,
-            currentName: theme.chat.connect.queue.firstText
+            themeCompanyName: theme.chat.connect.queue.firstText
         )
 
         theme.call.connect.queue.firstText = companyName(
             using: interactor,
-            currentName: theme.call.connect.queue.firstText
+            themeCompanyName: theme.call.connect.queue.firstText
         )
 
         let viewFactory = ViewFactory(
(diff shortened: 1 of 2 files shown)


EgorovEI (approved):

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
```diff
@@ -81,20 +81,23 @@ extension Glia {
...
-
         let companyNameStringKey = "general.company_name"
 
-        // Company name has been set on the custom locale.
-        if let remoteCompanyName = stringProviding?.getRemoteString(companyNameStringKey) {
+        // Company name has been set on the custom locale and is not empty.
+        if
+            let remoteCompanyName = stringProviding?.getRemoteString(companyNameStringKey),
+            !remoteCompanyName.isEmpty
+        {
```
yurii-glia, line 92: didn't see this kind of formatting in our code base, would it make sense to align with other code or provide a new codebase style?

yurii-glia (approved):

igorkravchenko (approved):

rasmustautsglia (approved):

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
gersonnoboa: Fixed beb7772ce9aeb5f947e11b96840126b360471bf5

gersonnoboa:
!squash

PR #796, opened 2023-10-11
#R Connecting state issue fixes

EgorovEI:
On SDK configuration, interactor instance is created and has .none state. When Call screen is loaded, its start() method is called which executes some updates based on current interactor.state. When configure method is called before each engagement start (like we currently do in WidgetsSDK Testing app), CallViewModel.start() does nothing, because interactor.state is .none. But when WidgesSDK is configured once (for example on the app launch), the same interactor instance is used for all subsequent engagements. When you end first engagement (Chat/Audio/Video, probably CV also), interactor.state is .ended. Then if you start new Audio/Video engagement, CallViewModel.start() calls CallViewModel.call.end() method, which breaks something and connecting state becomes infinite. 

This commit fixes calling CallViewModel.call.end() on Audio/Video engagement start. 
The toggle (UISwitch) was added to Main screen, that provides the ability to disable/enable SDK configuration before each engagement.

MOB-2730

### DIFF
--- a/GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
+++ b/GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
@@ -11,6 +11,8 @@ class EngagementViewModel: CommonEngagementModel {
     let alertConfiguration: AlertConfiguration
     let environment: Environment
     let screenShareHandler: ScreenShareHandler
+    // Need to keep strong reference of `activeEngagement`,
+    // to be able to fetch survey after ending engagement
     var activeEngagement: CoreSdkClient.Engagement?
 
     private(set) var isViewActive = ObservableValue<Bool>(with: false)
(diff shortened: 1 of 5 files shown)


gersonnoboa (approved):

GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
```diff
@@ -34,6 +36,7 @@ class EngagementViewModel: CommonEngagementModel {
     }
 
     deinit {
+        print("☠️", Self.self, ObjectIdentifier(self))
```
igorkravchenko, line 39: Is this still relevant?

igorkravchenko (approved):
Left one comment about logging `deinit` of view model.

GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
EgorovEI: Oops)

EgorovEI: Fixed 2095a493dd034ee78bb1220038e6b0b6fee41f0e

EgorovEI:
!squash

PR #802, opened 2023-10-13
Add support for default values in single choice questions in survey

gersonnoboa:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2747

**What was solved?**
Default values weren't selected automatically in surveys that showed a single option. With this change, a new `defaultOption` is added to the survey and used in case the selected value is not there. Now this works as it should, and how it works in Android.

The reason why a default option was added and the default value is not just set as the selected one, is because the question view starts with props, and the prop is either selected or not. However, the question view wouldn't know if an option is selected because of user input or because of the default value. If the view is selected because of user input, then `opt.select(opt)` needs to be executed only on user interaction, but in case of default value, it needs to be run automatically as soon as the view is presented. I don't know how to make this distinction without having the default value separately.

Also, a snapshot test was added to ensure that when there's a default value, it's selected correctly.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/ViewController/Survey/Components/SingleChoiceQuestionView/Survey.SingleChoiceQuestionView.swift
+++ b/GliaWidgets/Sources/ViewController/Survey/Components/SingleChoiceQuestionView/Survey.SingleChoiceQuestionView.swift
@@ -85,14 +85,33 @@ extension Survey {
             zip(props.options, optionsStack.arrangedSubviews)
                 .forEach { opt, view in
                     guard let checkboxView = view as? CheckboxView else { return }
+
+                    let state = Self.handleSelection(with: props, option: opt)
+
                     checkboxView.props = .init(
                         title: opt.name,
-                        state: props.selected == opt ? .selected : .active
+                        state: state
                     ) {
                         opt.select(opt)
                     }
                 }
             validationError.isHidden = !props.showValidationError
+        }
+
+        static func handleSelection(
+            with props: Props,
+            option opt: Survey.Option<String>
+        ) -> CheckboxView.State {
+            // If user selected or it should be selected by default.
+            let isSelected = props.selected == opt || props.defaultOption == opt
+
+            // Trigger selection manually because the option has
+            // been selected by default, not because of user input.
+            if isSelected, props.selected == nil {
+                opt.select(opt)
+            }
+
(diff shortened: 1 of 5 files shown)


EgorovEI (approved):

rasmustautsglia (approved):

GliaWidgets/Sources/ViewController/Survey/Components/SingleChoiceQuestionView/Survey.SingleChoiceQuestionView.swift
```diff
@@ -85,9 +85,34 @@ extension Survey {
             zip(props.options, optionsStack.arrangedSubviews)
                 .forEach { opt, view in
                     guard let checkboxView = view as? CheckboxView else { return }
+
+                    var state: CheckboxView.State
```
igorkravchenko, line 89: I think it will be beneficial to extract this logic into pure static function and also cover it with unit tests. Ideally that logic should not be a part of view but view model, but there is no such for Survey, as I understand.

igorkravchenko (commented):
Left one comment about logic in view.

GliaWidgets/Sources/ViewController/Survey/Components/SingleChoiceQuestionView/Survey.SingleChoiceQuestionView.swift
yurii-glia: i may be wrong, but could this switch be simplified with if statement like this? will it not be less code and more straightforward what happens inside?
```swift
                    // If user selected or it should be selected by default.
                    let isSelected = props.selected == opt || props.defaultOption == opt

                    // Handle first selection and apply default selection
                    if isSelected, props.selected == nil {
                        opt.select(opt)
                    }
```

gersonnoboa: First I added the tests and then I exchanged it for Yurii's code, everything works c3c6c15fa0eac9f067a7a077e86d0692ca449c71

yurii-glia (approved):

gersonnoboa:
!squash

PR #829, opened 2023-11-03
Add manual locale override capability

gersonnoboa:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2798

**What was solved?**
The configuration accepts a manual locale override, which is `nil` by default. If the Core SDK returns an incorrect manual locale override, then the Widgets SDK also returns an error on configuration to let know the integrator about it. If the integrator provides a wrong site API key, this commit also makes sure to inform the integrator instead of hanging.

Also, the settings screen now includes the possibility to add a manual locale override visually. If it's empty, then it sends nil.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore
 - [x] Migration guide (If changes are needed for integrator already using the SDK - what needs to be communicated? Add underneath please)

There's a new completion that is mandatory and that gives a Result, so that the integrator knows if there has been an issue with configuration instead of figuring it out when starting the engagement. Also, when creating the configuration, the integrator can add a `manualLocaleOverride`.

**Additional info:**

**Screenshots:**

### DIFF
--- a/TestingApp/ViewController/ViewController.swift
+++ b/TestingApp/ViewController/ViewController.swift
@@ -91,7 +91,6 @@ class ViewController: UIViewController {
 
     private func showErrorAlert(using error: Error) {
         if let gliaError = error as? GliaError {
-
             switch gliaError {
             case GliaError.engagementExists:
                 alert(message: "Failed to start\nEngagement is ongoing, please use 'Resume' button")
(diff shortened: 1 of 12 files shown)


TestingApp/ViewController/ViewController.swift
```diff
@@ -403,7 +406,10 @@ extension ViewController {
 
     @IBAction private func toggleAuthentication() {
         catchingError {
-            try Glia.sharedInstance.configure(with: configuration) { [weak self] in
+            try Glia.sharedInstance.configure(
```
EgorovEI, line 409: `configure` with `queueId` is deprecated

EgorovEI (approved):

TestingApp/ViewController/ViewController.swift
gersonnoboa: Fixed, i added the result also a17158aa2793fe617819543052d53269ad7c8fd6

rasmustautsglia (approved):

gersonnoboa:
!squash

PR #838, opened 2023-11-14
Check for critical imports in swift files

rasmustautsglia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2450

**What was solved?**
This PR takes an alternative approach to Swiftlint usage and uses script instead to determine if critical imports are present.

If SwiftUI is imported, UIKit nor Foundation is needed. If not, then either of them has to be in a swift file. When there isn't one, Xcode will display a warning.

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**


gersonnoboa (commented):
Does this run only on Xcode? One of the benefits of doing it on SwiftLint was that if there's something wrong, then it would show up on GitHub as a comment. This would only show up when you are developing. However yes, SwiftLint sucks so I don't know exactly what to do.

rasmustautsglia:
Swiftlint is great for checking elements that are present. However, it has its downside, which is that it cannot detect what is absent. I didn't find a possibility to write a rule that checks for missing imports.

gersonnoboa (approved):
Well at least now the warning will be there :)

igorkravchenko (commented):
Does this change imply that UIKit has to be always imported?

rasmustautsglia:
This change implies that either UIKit or Foundation has to be always imported for .swift files. Currently, there is about 110 files that don't have either

bitrise.yml
```diff
@@ -173,6 +173,10 @@ workflows:
             make clone-snapshots-ci repo_url="https://$GITHUB_USERNAME:$GITHUB_API_TOKEN@$SNAPSHOTS_REPO"
     - cache-pull@2: {}
     - certificate-and-profile-installer@1: {}
+    - script@1:
+        inputs:
+        - content: |-
+            bundle install
```
gersonnoboa, line 179: Perhaps this change means that all workflows that build the app need to have a `bundle install` now, not only the one that runs =checks in the PRs.

Podfile
```diff
@@ -1,4 +1,6 @@
 source 'https://github.com/CocoaPods/Specs.git'
+plugin 'cocoapods-no-autoimports'
+# cocoapods-no-autoimports__pods = ['Pods-GliaWidgets.debug']
```
gersonnoboa, line 3: Is this needed for something?

Gemfile
```diff
@@ -1,3 +1,4 @@
 source "https://rubygems.org"
 
 gem "fastlane"
+gem 'cocoapods-no-autoimports'
```
gersonnoboa, line 4: GitHub is complaining about a lack of a new line here.

Podfile
rasmustautsglia: Yes it is a critical part of the plugin

EgorovEI (approved):

Gemfile
rasmustautsglia: fixed [a00e1de](https://github.com/salemove/ios-sdk-widgets/pull/838/commits/a00e1de203c793d83b3c7bb06c8f106ef53bf89d)

bitrise.yml
rasmustautsglia: This commit will be dropped. Thank you for the notice

igorkravchenko (approved):

PR #843, opened 2023-11-28
Add Logger to Widgets

igorkravchenko:
Initiate local and remote logging in Widgets through Core SDK.

MOB-2813

**Jira issue:**
https://glia.atlassian.net/browse/MOB-2813

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


GliaWidgets/Sources/CoreSDKClient/CoreSDKClient.Live.swift
```diff
@@ -54,7 +54,9 @@ extension CoreSdkClient {
             downloadSecureFile: GliaCore.sharedInstance.secureConversations.downloadFile(_:progress:completion:),
             startSocketObservation: GliaCore.sharedInstance.startSocketObservation,
             stopSocketObservation: GliaCore.sharedInstance.stopSocketObservation,
-            createSendMessagePayload: CoreSdkClient.SendMessagePayload.init(content:attachment:)
+            createSendMessagePayload: CoreSdkClient.SendMessagePayload.init(content:attachment:),
+            createLogger: { try Logger(GliaCore.sharedInstance.createLogger(externalParameters: $0))
+            }
```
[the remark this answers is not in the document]
igorkravchenko: Fixed ac95e72aa070dcec511958d3c00a722a0fcbae5f.

rasmustautsglia (approved):

gersonnoboa (approved):

GliaWidgets/Public/Glia/Glia+Logging.swift
```diff
@@ -0,0 +1,88 @@
+extension Glia {
```
yurii-glia, line 1: This structures are not public, so let's move this file to another directory please.

GliaWidgets/Public/Glia/Glia+Logging.swift
```diff
@@ -0,0 +1,88 @@
...
+                try logger.configureRemoteLogLevel(.none)
+            // In non-debug mode local logs are turned off and remote are on.
+            } else {
+                try logger.configureLocalLogLevel(.none)
+                try logger.configureRemoteLogLevel(.info)
+            }
+
+            self.loggerPhase = .configured(
+                logger
+            )
```
yurii-glia, line 66: since here is only one unnamed param, perhaps it makes sense to format inline, would it make sense?
```suggestion
            self.loggerPhase = .configured(logger)
```

GliaWidgets/Public/Glia/Glia+Logging.swift
```diff
@@ -0,0 +1,88 @@
...
+    /// collected logs could be transferred to configured logger, but
+    /// this is out of scope for current implementation.
+    struct NotConfiguredLogger: CoreSdkClient.Logging {
+        var environment: Environment
+        var localLogger: (CoreSdkClient.Logging)?
+        var remoteLogger: (CoreSdkClient.Logging)?
+        var oneTime: CoreSdkClient.Logging { self }
+
+        func error(_ object: @autoclosure () -> Any, file: String, function: String, line: Int) {
+            environment.print(object())
```
yurii-glia, line 21: do we need to pass an `object` closure instead of an execution function and pass it as a value?
in sense of optiomization I think it is better to pass closure because in this case, Core SDK will be able to verify whether should it be logged or not. otherwise might be case when complex `object` with a bunch of formation will be executed here, but filter out on Core Sdk level.

GliaWidgets/Sources/CoreSDKClient/Logger/Logger.Interface.swift
```diff
@@ -0,0 +1,82 @@
...
+
+        var localLogger: Logger? {
+            localLoggerClosure()
+        }
+        var remoteLogger: Logger? {
+            remoteLoggerClosure()
+        }
+
+        func error(_ object: @autoclosure () -> Any, file: String = #fileID, function: String = #function, line: Int = #line) {
+            errorClosure(object(), file, function, line)
```
yurii-glia, line 37: the same question as above about passing closure or value. would we able to optimize this?

GliaWidgets/Public/Glia/Glia+Logging.swift
igorkravchenko: For this particular case it is not needed because this happens in `NotConfiguredLogger` which is just placeholder and wrapper around `print` function.

GliaWidgets/Public/Glia/Glia+Logging.swift
igorkravchenko: Moved it 1de9cac1347aebd466b6bf8a392de67800d6b4c2.

GliaWidgets/Public/Glia/Glia+Logging.swift
igorkravchenko: Updated ecd53cbe08144a47a78655580bfa1c224abd0e44.

igorkravchenko:
!squash

PR #852, opened 2023-12-06
Log extra events for debugging

igorkravchenko:
~UPD: Awaiting CoreSDK release to allow configuration of Logger in Glia instance.~

MOB-2897

**Jira issue:**
https://glia.atlassian.net/browse/MOB-2897

**What was solved?**
Add logging of events for debugging. Most of the events correspond to events added in Android Widgets SDK.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Public/Glia/Glia.OpaqueAuthentication.swift
+++ b/GliaWidgets/Public/Glia/Glia.OpaqueAuthentication.swift
@@ -50,7 +50,10 @@ extension Glia {
         }
 
         return .init(
-            authenticateWithIdToken: { idToken, accessToken, callback in
+            authenticateWithIdToken: { [weak self] idToken, accessToken, callback in
+                self?.loggerPhase.logger.prefixed(Self.self).info(
+                    "Authenticate. Is external access token used: \(accessToken != nil)"
+                )
                 auth.authenticate(
                     with: .init(rawValue: idToken),
                     externalAccessToken: accessToken.map { .init(rawValue: $0) }
@@ -67,7 +70,8 @@ extension Glia {
                     callback(result.mapError(Glia.Authentication.Error.init) )
                 }
             },
-            deauthenticateWithCallback: { callback in
+            deauthenticateWithCallback: { [weak self] callback in
+                self?.loggerPhase.logger.prefixed(Self.self).info("De-authenticate")
                 auth.deauthenticate { result in
                     switch result {
                     case .success:
--- a/GliaWidgets/Public/Glia/Glia.swift
+++ b/GliaWidgets/Public/Glia/Glia.swift
@@ -85,7 +85,8 @@ public class Glia {
             getCurrentEngagement: environment.coreSdk.getCurrentEngagement,
             eventHandler: onEvent,
             orientationManager: environment.orientationManager,
-            proximityManager: environment.proximityManager
+            proximityManager: environment.proximityManager,
+            log: loggerPhase.logger
         )
     )
     var rootCoordinator: EngagementCoordinator?
@@ -101,13 +102,24 @@ public class Glia {
     init(environment: Environment) {
         self.environment = environment
         self.theme = Theme()
-        self.loggerPhase = .notConfigured(
-            .init(
-                NotConfiguredLogger(
-                    environment: .init(print: environment.print)
-                )
-            )
-        )
+
+        do {
+            let logger = try environment.coreSdk.createLogger(CoreSdkClient.Logger.loggerParameters)
+            if environment.conditionalCompilation.isDebug() {
+            // In debug mode local and remote logs are turned off.
+                try logger.configureLocalLogLevel(.none)
+                try logger.configureRemoteLogLevel(.none)
+            // In non-debug mode local logs are turned off and remote are on.
+            } else {
+                try logger.configureLocalLogLevel(.none)
+                try logger.configureRemoteLogLevel(.info)
+            }
+
+            loggerPhase = .configured(logger)
+        } catch {
+            environment.print("Unable to configure logger: '\(error)'.")
+            self.loggerPhase = .notConfigured(.notConfigured)
+        }
     }
 
     /// Setup SDK using specific engagement configuration without starting the engagement.
@@ -151,9 +163,9 @@ public class Glia {
             switch result {
             case .success:
                 defer {
-                    self.setupLogging()
+                    self.loggerPhase.logger.prefixed(Self.self).info("Initialize Glia Widgets SDK")
                     if uiConfig != nil {
-                        self.loggerPhase.logger.remoteLogger?.oneTime.prefixed(Self.self)
+                        self.loggerPhase.logger.remoteLogger?.prefixed(Self.self)
                             .info("Setting Unified UI Config")
                     }
                 }
@@ -225,6 +237,7 @@ public class Glia {
     /// because `GliaError.clearingVisitorSessionDuringEngagementIsNotAllowed` will occur otherwise.
     ///
     public func clearVisitorSession(_ completion: @escaping (Result<Void, Error>) -> Void) {
+        loggerPhase.logger.prefixed(Self.self).info("Clear visitor session")
         guard environment.coreSdk.getCurrentEngagement() == nil else {
             completion(.failure(GliaError.clearingVisitorSessionDuringEngagementIsNotAllowed))
             return
@@ -301,6 +314,8 @@ public class Glia {
 
     /// Ends active engagement if existing and closes Widgets SDK UI (includes bubble).
     public func endEngagement(_ completion: @escaping (Result<Void, Error>) -> Void) {
+        loggerPhase.logger.prefixed(Self.self).info("End engagement by integrator")
+
         defer {
             onEvent?(.ended)
             rootCoordinator = nil
@@ -361,8 +376,14 @@ extension Glia {
                     self?.callVisualizer.offerScreenShare(
                         from: operators ?? [],
                         configuration: Theme().alertConfiguration.screenShareOffer,
-                        accepted: { answer(true) },
-                        declined: { answer(false) }
+                        accepted: { [weak self] in
+                            self?.loggerPhase.logger.prefixed(Self.self).info("Screen sharing accepted by visitor")
+                            answer(true)
+                        },
+                        declined: {
+                            self?.loggerPhase.logger.prefixed(Self.self).info("Screen sharing declined by visitor")
+                            answer(false)
+                        }
                     )
                 }
             case let .upgradeOffer(offer, answer):
@@ -384,9 +405,12 @@ extension Glia {
                 if case .ended = state {
                     self?.callVisualizer.endSession()
                     self?.onEvent?(.ended)
+                    self?.loggerPhase.logger.prefixed(Self.self).info("Engagement ended")
                 } else if case .engaged = state {
+                    self?.loggerPhase.logger.prefixed(Self.self).info("New Call visualizer engagement loaded")
                     self?.callVisualizer.handleEngagementRequestAccepted()
                     self?.onEvent?(.started)
+                    self?.loggerPhase.logger.prefixed(Self.self).info("Engagement started")
                 }
             default:
                 break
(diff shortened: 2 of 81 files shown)


GliaWidgets/Sources/Interactor/Interactor.swift
```diff
@@ -135,6 +135,7 @@ extension Interactor {
                 self?.state = .ended(.byError)
                 failure(error)
             case .success(let ticket):
+                self?.environment.log.prefixed(Self.self).info("Queue for engagement stopped due to error or empty queue")
```
EgorovEI, line 138: Should it be in the `failure` case?

rasmustautsglia (approved):

GliaWidgets/Sources/LockIsolated/LockIsolated.swift
```diff
@@ -0,0 +1,100 @@
+import Foundation
+
+/// A generic wrapper for isolating a mutable value with a lock.
+@dynamicMemberLookup
+final class LockIsolated<Value>: @unchecked Sendable {
```
gersonnoboa, line 5: The indentation is wrong on this file

gersonnoboa (approved):
I'm a little lost though :)

GliaWidgets/Sources/LockIsolated/LockIsolated.swift
igorkravchenko: Fixed fc7a7400893c7c46ab9eab3df3a22340a5df1b88.

GliaWidgets/Public/Glia/Glia.OpaqueAuthentication.swift
```diff
@@ -50,7 +50,10 @@ extension Glia {
         }
 
         return .init(
-            authenticateWithIdToken: { idToken, accessToken, callback in
+            authenticateWithIdToken: { [weak self] idToken, accessToken, callback in
+                self?.loggerPhase.logger.prefixed(Self.self).info(
+                    "Authenticate. Is external access token used: \(accessToken != nil)"
```
yurii-glia, line 55: - is it authentication result closure or it is just function. For me it looks like this log should be inside of the callback to be sure that authentication is succeeded.
- I'd propose to put the value inside of single quote to be sure that inside no spaces
```swift
 "Authenticate. Is external access token used: '\(accessToken != nil)'."
```
- `accessToken` is used for internal usage of integrator services. For debug purposes it is more useful to use `idToken` instead, since there is payload (with name, email, etc.). At the same time name, email, etc. may be sensitive data so we need to confirm that it is allowed to log.

GliaWidgets/Public/Glia/Glia.OpaqueAuthentication.swift
```diff
@@ -67,7 +70,8 @@ extension Glia {
                     callback(result.mapError(Glia.Authentication.Error.init) )
                 }
             },
-            deauthenticateWithCallback: { callback in
+            deauthenticateWithCallback: { [weak self] callback in
+                self?.loggerPhase.logger.prefixed(Self.self).info("De-authenticate")
```
yurii-glia, line 74: would it make sense to move this log inside of closure and log result of deauthentication?

GliaWidgets/Public/Glia/Glia.swift
```diff
@@ -152,8 +184,9 @@ public class Glia {
             case .success:
                 defer {
                     self.setupLogging()
+                    self.loggerPhase.logger.prefixed(Self.self).info("Initialize Glia Widgets SDK")
```
yurii-glia, line 187: this is a configuration function, not initialisation, so perhaps it makes sense to change this message to something like "Configure Glia Widgets SDK"?

GliaWidgets/SecureConversations/ChatTranscript/SecureConversations.TranscriptModel+GVA.swift
```diff
@@ -118,6 +118,7 @@ private extension SecureConversations.TranscriptModel {
                     in: self.pendingSection
                 )
             case .failure:
+                environment.log.prefixed(Self.self).info("Show Unexpected error Dialog")
```
yurii-glia, line 121: would it make sense to log error also?

GliaWidgets/SecureConversations/Welcome/SecureConversations.WelcomeViewModel.swift
```diff
@@ -128,13 +128,14 @@ private extension SecureConversations.WelcomeViewModel {
...
             queueIds
-        ) { [weak self, alertConfiguration = environment.alertConfiguration] result in
+        ) { [weak self, alertConfiguration = environment.alertConfiguration, environment] result in
             self?.sendMessageRequestState = .waiting
 
             switch result {
             case .success:
                 self?.delegate?(.confirmationScreenRequested)
             case .failure:
+                environment.log.prefixed(Self.self).info("Show Unexpected error Dialog")
```
yurii-glia, line 138: would not it be beneficial to log specific error here?

GliaWidgets/SecureConversations/Welcome/SecureConversations.WelcomeViewModel.swift
```diff
@@ -151,6 +152,7 @@ private extension SecureConversations.WelcomeViewModel {
             case let .success(site):
                 self?.isAttachmentsAvailable = site.allowedFileSenders.visitor
             case .failure:
+                self?.environment.log.prefixed(Self.self).info("Show Unexpected error Dialog")
```
yurii-glia, line 155: would not it be beneficial to log specific error here?

GliaWidgets/Sources/CallVisualizer/CallVisualizer.swift
```diff
@@ -78,6 +80,8 @@ public final class CallVisualizer {
     /// - Parameter source: The current viewController to present from.
     ///
     public func showVisitorCodeViewController(from source: UIViewController) {
+        environment.log.prefixed(Self.self).info("Show visitor code")
+        environment.log.prefixed(Self.self).info("Show Visitor Code Dialog")
```
yurii-glia, line 84: do we need to have this duplication?

GliaWidgets/Sources/CallVisualizer/CallVisualizer.swift
```diff
@@ -102,6 +106,8 @@ public final class CallVisualizer {
         into container: UIView,
         onEngagementAccepted: @escaping () -> Void
     ) {
+        environment.log.prefixed(Self.self).info("Show visitor code")
+        environment.log.prefixed(Self.self).info("Show Visitor Code Dialog")
```
yurii-glia, line 110: duplicaiton?

GliaWidgets/Sources/Coordinators/EngagementCoordinator/EngagementCoordinator.swift
```diff
@@ -197,6 +205,7 @@ extension EngagementCoordinator {
             updateProps: { viewController.props = $0 },
             onError: { [weak self] _ in
                 guard let self else { return }
+                self.environment.log.prefixed(Self.self).info("Show Unexpected error Dialog")
```
yurii-glia, line 208: would it be beneficial to know error of Unexpected error dialog?

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -239,7 +240,7 @@ class ChatViewModel: EngagementViewModel {
                         self.action?(.scrollToBottom(animated: true))
                     case .failure:
                         guard let self else { return }
-
+                        self.environment.log.prefixed(Self.self).info("Show Unexpected error Dialog")
```
yurii-glia, line 243: do we need to log error here?

yurii-glia (commented):
- Would it be more clear to use `.` in the end of sentence in logs?
-  Would it make sense to use `Deinit` instead of `Destroy`  in logs to align with function name?
- Do we need to have init log also similar to deinit?

@igorkravchenko  @gersonnoboa what do you think about these?

igorkravchenko:
> * Would it be more clear to use `.` in the end of sentence in logs?
> * Would it make sense to use `Deinit` instead of `Destroy`  in logs to align with function name?
> * Do we need to have init log also similar to deinit?
> 
> @igorkravchenko @gersonnoboa what do you think about these?

@yurii-glia  It makes sense. However the idea was to have identical logs as much as possible for both platforms. So what we can do is to create tickets proposing these additions.
> Do we need to have init log also similar to deinit

It think it makes it makes more sense to log creation in method that instantiates object rather than in init itself, because then we have context. That is how it is done right now.

GliaWidgets/Sources/CallVisualizer/CallVisualizer.swift
igorkravchenko: Log on line 83, seems to be not present in Android, so it should be removed here as well. It is present in comparison table though 03ec82f55b8134c57ab9d2d986b5a986b84fdb08.

GliaWidgets/Sources/CallVisualizer/CallVisualizer.swift
igorkravchenko: Removed 03ec82f55b8134c57ab9d2d986b5a986b84fdb08.

GliaWidgets/Public/Glia/Glia.OpaqueAuthentication.swift
igorkravchenko: We log here the fact if access token is used or not, so it is boolean value. As stated [here](https://github.com/salemove/ios-sdk-widgets/pull/852#issuecomment-1854455410) proposed change has to be done on both platforms and can not be a part of this PR.

GliaWidgets/Public/Glia/Glia.OpaqueAuthentication.swift
igorkravchenko: It makes sense. But ss stated [here](https://github.com/salemove/ios-sdk-widgets/pull/852#issuecomment-1854455410) proposed change has to be done on both platforms and can not be a part of this PR.

GliaWidgets/Public/Glia/Glia.swift
igorkravchenko: [Same](https://github.com/salemove/ios-sdk-widgets/pull/852#issuecomment-1854455410).

GliaWidgets/SecureConversations/ChatTranscript/SecureConversations.TranscriptModel+GVA.swift
igorkravchenko: Yes. But has to be done on both platforms.

GliaWidgets/SecureConversations/Welcome/SecureConversations.WelcomeViewModel.swift
igorkravchenko: https://github.com/salemove/ios-sdk-widgets/pull/852#issuecomment-1854455410

GliaWidgets/SecureConversations/Welcome/SecureConversations.WelcomeViewModel.swift
igorkravchenko: https://github.com/salemove/ios-sdk-widgets/pull/852#issuecomment-1854455410

GliaWidgets/Sources/Coordinators/EngagementCoordinator/EngagementCoordinator.swift
igorkravchenko: https://github.com/salemove/ios-sdk-widgets/pull/852#issuecomment-1854455410

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
igorkravchenko: https://github.com/salemove/ios-sdk-widgets/pull/852#issuecomment-1854455410

GliaWidgets/Sources/CallVisualizer/CallVisualizer.swift
andrews-moc: Sorry, there is no "Show visitor code" on Android. Most probably, indeed, to avoid duplication. I just missed to remove this event from the spreadsheet.

So you can remove "Show visitor code".

yurii-glia (approved):

PR #863, opened 2024-01-02
Fix local video being shown when onHold

rasmustautsglia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2943

**What was solved?**
When the operator had put the visitor on hold, and then requested a two-way video, the operator was able to see the visitor's video even while on hold. This PR addresses the issue and only allows the video to be shown when the engagement is not on hold.

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
+++ b/GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
@@ -236,36 +236,6 @@ class CallViewModel: EngagementViewModel, ViewModel {
         }
     }
 
-    private func handleAudioStreamError(_ error: CoreSdkClient.SalemoveError) {
-        switch error.error {
-        case let mediaError as CoreSdkClient.MediaError:
-            switch mediaError {
-            case .permissionDenied:
-                showSettingsAlert(
-                    with: alertConfiguration.microphoneSettings
-                )
-            default:
-                showAlert(for: error)
-            }
-        default:
-            showAlert(for: error)
-        }
-    }
-
-    private func handleVideoStreamError(_ error: CoreSdkClient.SalemoveError) {
-        switch error.error {
-        case let mediaError as CoreSdkClient.MediaError:
-            switch mediaError {
-            case .permissionDenied:
-                showSettingsAlert(with: alertConfiguration.cameraSettings)
-            default:
-                showAlert(for: error)
-            }
-        default:
-            showAlert(for: error)
-        }
-    }
-
     func showSnackBarIfNeeded() {
         environment.fetchSiteConfigurations { [weak self] result in
             switch result {
@@ -285,9 +255,9 @@ class CallViewModel: EngagementViewModel, ViewModel {
         case .videoStreamAdded(let stream):
             call.updateVideoStream(with: stream)
         case .audioStreamError(let error):
-            handleAudioStreamError(error)
+            handleStreamError(error, with: alertConfiguration.microphoneSettings)
         case .videoStreamError(let error):
-            handleVideoStreamError(error)
+            handleStreamError(error, with: alertConfiguration.cameraSettings)
         case .upgradeOffer(let offer, answer: let answer):
             offerMediaUpgrade(offer, answer: answer)
         case .updateOffer(let offer):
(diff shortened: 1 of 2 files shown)


rasmustautsglia:
Body length violations made me move 2 unrelated method to an extension.

EgorovEI (approved):
Are we able to add unit test for this case?

GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
```diff
@@ -301,6 +271,36 @@ class CallViewModel: EngagementViewModel, ViewModel {
...
+            case .permissionDenied:
+                showSettingsAlert(
+                    with: alertConfiguration.microphoneSettings
+                )
+            default:
+                showAlert(for: error)
+            }
+        default:
+            showAlert(for: error)
+        }
```
yurii-glia, line 287: would it make sense to use the `where` keyword to simplify the statement?
```swift
        switch error.error {
        case let mediaError as CoreSdkClient.MediaError where mediaError == .permissionDenied:
            showSettingsAlert(
                with: alertConfiguration.microphoneSettings
            )
        default:
            showAlert(for: error)
        }
```

GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
```diff
@@ -367,7 +367,11 @@ extension CallViewModel {
     private func showLocalVideo(with stream: CoreSdkClient.VideoStreamable) {
         action?(.switchToVideoMode)
         action?(.setLocalVideo(stream.getStreamView()))
-        stream.playVideo()
+        stream.onHold = { isOnHold in
+            if !isOnHold {
+                stream.playVideo()
+            }
```
yurii-glia, line 373: perhaps I missed something, however, I see that we `play` stream, but don't pause it, should we do it?

GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
yurii-glia: perhaps we even can union two function to avoid boilerplate and have more universal function like.
```swift
    private func handleStreamError(
        _ error: CoreSdkClient.SalemoveError,
        with message: SettingsAlertConfiguration
    ) {
        switch error.error {
        case let mediaError as CoreSdkClient.MediaError where mediaError == .permissionDenied:
            showSettingsAlert(
                with: message
            )
        default:
            showAlert(
                for: error
            )
        }
    }
```
what do you think?

igorkravchenko (commented):
Is it possible to add unit tests for added logic?

GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
rasmustautsglia: No, because `pause` and `resume` are called when onHold is triggered. But with this situation, video stream was added when visitor was on hold , not taking into account that during on hold we are not allowed to show the video. If operator resumes the call, so does the SDK, because that logic works correctly. And, if operator puts the visitor on hold again, that is executed correctly too

gersonnoboa (approved):
I like Yurii's suggestions but other than that, it's fine

rasmustautsglia:
Unit tests added [778011e](https://github.com/salemove/ios-sdk-widgets/pull/863/commits/778011e82cfac2499bdd2d693221d3254c2d8442)

GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
yurii-glia: then this closure looks a little bit strange in that we handle only one case but closure provides two possible cases. perhaps we can rethink & rework it later. 
or leave comment for now that it is expected that we ignore `on hold` case.

yurii-glia (approved):

EgorovEI (approved):

rasmustautsglia:
!squash

igorkravchenko (approved):

PR #865, opened 2024-01-03
Fix error showing up in Xcode console as warning

gersonnoboa:
The error was about a table view not being shown yet but trying to lay out
its visible cells. This was fixed by using `async` instead of `asyncIfNeeded`,
which allows enough time for the table view to be shown.

MOB-2971

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/Interactor/Interactor.swift
+++ b/GliaWidgets/Sources/Interactor/Interactor.swift
@@ -96,7 +96,7 @@ class Interactor {
             .forEach {
                 let handler = $0.1
 
-                environment.gcd.mainQueue.asyncIfNeeded {
+                environment.gcd.mainQueue.async {
                     handler(event)
                 }
             }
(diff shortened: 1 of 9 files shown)


igorkravchenko (commented):
Seems like this change has to be reflected in unit tests - it is required to overwrite `async ` instead of `asyncIfNeeded`.

gersonnoboa:
@igorkravchenko fixed, my apologies

igorkravchenko (approved):

yurii-glia:
3 tests failed, are they related to changes?

gersonnoboa:
@yurii-glia I just removed the "fix" for the constraints as it was making the snapshot tests fail. I think I have spent too much time on such an issue and finding a fix would take even more.

EgorovEI (approved):

yurii-glia (approved):

PR #869, opened 2024-01-16
Add tests for Secure Conversations coordinator

gersonnoboa:
Correct actions and creation of view controllers in the coordinator has been covered with unit tests.

MOB-2914

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgetsTests/SecureConversations/Coordinator/SecureConversations.CoordinatorTests.swift
+++ b/GliaWidgetsTests/SecureConversations/Coordinator/SecureConversations.CoordinatorTests.swift
@@ -1,0 +1,212 @@
+import Foundation
+import XCTest
+@testable import GliaWidgets
+
+final class SecureConversationsCoordinatorTests: XCTestCase {
+    var navigationPresenter = NavigationPresenter(with: NavigationController())
+    var coordinator: SecureConversations.Coordinator!
+
+    override func setUp() {
+        coordinator = createCoordinator()
+    }
+
+    func createCoordinator(
+        initialScreen: SecureConversations.InitialScreen = .welcome
+    ) -> SecureConversations.Coordinator {
+        return SecureConversations.Coordinator(
+            messagingInitialScreen: initialScreen,
+            viewFactory: .mock(),
+            navigationPresenter: navigationPresenter,
+            environment: .mock
+        )
+    }
+
+    // Start
+    func test_startGeneratesWelcomeViewController() {
+        let viewController = coordinator.start() as? SecureConversations.WelcomeViewController
+
+        XCTAssertNotNil(viewController)
+    }
+
+    func test_startGeneratesChatViewController() {
+        let coordinator = createCoordinator(initialScreen: .chatTranscript)
+        let viewController = coordinator.start() as? ChatViewController
+
+        XCTAssertNotNil(viewController)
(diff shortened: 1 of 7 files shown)


GliaWidgetsTests/SecureConversations/Coordinator/SecureConversations.CoordinatorTests.swift
```diff
@@ -0,0 +1,212 @@
...
+        _ = coordinator.start()
+        coordinator.viewModel?.delegate?(.confirmationScreenRequested)
+        let confirmationViewController = navigationPresenter.viewControllers
+            .last as? SecureConversations.ConfirmationViewController
+
+        XCTAssertNotNil(confirmationViewController)
+    }
+
+    func test_mediaPickerRequested() {
+        let welcomeViewController = coordinator.start() as! SecureConversations.WelcomeViewController
```
igorkravchenko, line 79: It is possible to avoid force casting, by using  `try XCTUnwrap(coordinator.start() as?SecureConversations.WelcomeViewController)` and marking `test_mediaPickerRequested` method throwing.

igorkravchenko (commented):
Looks good. I would recommend to avoid force casting by using `XCTUnwrap`.

GliaWidgetsTests/SecureConversations/Coordinator/SecureConversations.CoordinatorTests.swift
gersonnoboa: Fix bfce3ead403f998eef6bcf9d8de660a98d89ce91

GliaWidgetsTests/SecureConversations/Welcome/SecureConversations.WelcomeViewController.Mock.swift
```diff
@@ -0,0 +1,37 @@
+#if DEBUG
```
EgorovEI, line 1: Do we need it in test target?

EgorovEI (approved):

igorkravchenko (approved):

GliaWidgetsTests/SecureConversations/Welcome/SecureConversations.WelcomeViewController.Mock.swift
gersonnoboa: @igorkravchenko I see that sometimes mocks are in the tests, but sometimes they are in the widgets themselves, and thus they have this `#if DEBUG` macro. Is there more of a recommended way to do this? Asking you because you are the environmental expert :)

EgorovEI: If I'm not mistaken, we discussed that would be good to have (or even move existing) mocks in test target. 
The only reason to keep them in WidgetsSDK target that I see is if we use some of them in both unit tests and snapshot tests targets to avoid duplications. But for this case we also can introduce additional target

gersonnoboa: Thank you. I removed the macros and kept the mocks in the test target.

PR #872, opened 2024-01-18
Add unit tests for ChatCoordinator

gersonnoboa:
Correct actions and creation of view controllers in the coordinator has been covered with unit tests.

MOB-2920

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


EgorovEI (approved):

GliaWidgetsTests/Sources/Coordinators/Chat/ChatCoordinatorTests.swift
```diff
@@ -0,0 +1,400 @@
...
+            viewModel.delegate?(
+                .mediaUpgradeAccepted(
+                    offer: mediaOffer,
+                    answer: { _, _ in}
+                )
+            )
+        default: XCTFail()
+        }
+
+        coordinator.delegate = { event in
```
igorkravchenko, line 121: What if delegate closure is not called for some reason (if it is `nil` for example) or called unexpected number of times? I think these cases will not be caught by this test.
Instead, we can try define array where we are to populate it with events delivered by delegate closure, then we can assert the contents of array. Does it make sense?

rasmustautsglia (approved):

igorkravchenko (commented):
Left comment about delegate events asserted in this style:
```
 coordinator.delegate = { event in
 /// Asserts of event goes here. 
}
```
This kind of test is fragile, because delegate closure may never be called or called multiple times. Instead:

```
var listToBeAsserted: Array<SomeDelegateEvent> = []

 coordinator.delegate = { event in
   listToBeAsserted.append(event)
}
// Asserts of event(s) in listToBeAsserted go here.
```

This will allow us to know that listToBeAsserted is not empty, and that its contents is expected.

GliaWidgetsTests/Sources/Coordinators/Chat/ChatCoordinatorTests.swift
gersonnoboa: Makes sense, will change these tests and then will open another PR in which I address this also for the secure conversations tests. Thank you.

gersonnoboa: Fixed df25ce2bc7f83589d7f52bc5992a6ca18c966e3d

igorkravchenko (approved):

gersonnoboa:
!squash

PR #877, opened 2024-01-22
POC for Localization as a dependency

rasmustautsglia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2881

**What was solved?**
This PR introduces a new structure `Localizable2` (placeholder name, because `Localizable` is still in use). This new approach allows localizable strings to be mocked, which enhances our testing capabilities. 

In order to avoid circular reference, the old `.tr` method couldn't be used, because it was using `Glia.sharedInstance.stringProvidingPhase` as a default value for `stringProviding`. In the new method, this is swapped for a closure that is called when the string is rendered and receives the current stringProvidingPhase from the SDK.

In the moment when passing down the closure didn't make any sense, the method is called and only the portion of the Localization object is injected to the view. You can see it in `VisitorCodeCoordinator()` where VisitorCodeViewController is initialized.

NB! This PR is a proof of concept and only includes the implementation of CallVisualizer's Visitor Code. Upon approval of this approach, new tickets will be made to further cover the SDK with Localizable2. When the entire SDK has made the transition to the new method, Localizable2 will be renamed to Localizable, and the old object will be deleted or deprecated.

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**


EgorovEI (approved):
Looks good to me

GliaWidgets/Localization2/Localization2.swift
```diff
@@ -0,0 +1,47 @@
+import Foundation
+
+struct Localization2 {
+    let callVisualizer: CallVisualizer
+}
+
+extension Localization2 {
```
gersonnoboa, line 7: This part seems like out of place considering the rest of the file's content.

GliaWidgets/Localization2/Localization2+Live.swift
```diff
@@ -0,0 +1,60 @@
+import Foundation
+
+extension Localization2 {
+    static func live(stringProviding: @escaping () -> StringProvidingPhase?) -> Self {
+        .init(callVisualizer: .init(
+            screenSharing: .init(
+                message: Localization2.tr(
```
gersonnoboa, line 7: So on the final version, this will look more like `message: Localization.ScreenSharing.message` rather than this? Because if the `tr` itself is here, we lose the autogenerated code that gets generated from the strings file, which would mean duplication and we would be fixing a problem by creating another.

gersonnoboa (commented):
In the final version, there's no need to deprecate `Localization` when it gets replaced by `Localization2`, because `Localization` is internal, unlike the previous strings which were public. This was exactly the reason why it was done like that, so that if changes were needed, we don't have to deprecate stuff.

GliaWidgets/Localization2/Localization2+Live.swift
rasmustautsglia: Autogeneration in the old fashion wasn't the envisioned outcome.

GliaWidgets/Localization2/Localization2.swift
rasmustautsglia: fixed [7ebe6c3](https://github.com/salemove/ios-sdk-widgets/pull/877/commits/7ebe6c35b8bd118e7ad5dd256847be668c74df43)

GliaWidgets/Localization2/Localization2+Live.swift
rasmustautsglia: No final version looks exactly like it is implemented here

gersonnoboa: Okay, if everyone else is fine with this approach, then I will not complain :) I find it a huge detriment though to manage this manually.

gersonnoboa (approved):

igorkravchenko (commented):
As of PoC it looks good. This will serve as example that it is possible to approach Localization as dependency. I also think that it makes sense to keep the PR as draft, for further refinement. But it would be beneficial to explore code generation instead of manually constructing structs that carry localized text, because it is time consuming and error prone process.

PR #888, opened 2024-02-16
Add MAXIMUM_UPLOADS_PER_MESSAGE variable

EgorovEI:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2646

**What was solved?**
Adding this variable gives a way to adjust maximum amount of files that can be attached to a single message. This is needed to make the test checking whether the attachment button is disabled when maximum number of file is attached shorter. This variable will be passed through `processInfo` from acceptance test.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Public/Glia/Glia.swift
+++ b/GliaWidgets/Public/Glia/Glia.swift
@@ -48,6 +48,10 @@ public protocol SceneProvider: AnyObject {
     func windowScene() -> UIWindowScene?
 }
 
+private extension Glia {
+    static let maximumUploads = 25
+}
+
 /// Glia's engagement interface.
 public class Glia {
     /// A singleton to access the Glia's interface.
@@ -58,6 +62,18 @@ public class Glia {
     public var onEvent: ((GliaEvent) -> Void)?
 
     var stringProvidingPhase: StringProvidingPhase = .notConfigured
+
+    lazy var maximumUploads: Int = {
+        var value = Self.maximumUploads
+        /// `MAXIMUM_UPLOADS_PER_MESSAGE` is used in acceptance tests to decrease the time
+        /// for uploading maximum number of files in a single message.
+        if let stringValue = environment.processInfo.value(for: .maximumUploads) {
+            /// Round to 1, because the attachment button doesn't turns to disabled state
+            /// if the value is set to 0.
+            value = max(Int(stringValue) ?? 1, 1)
+        }
+        return value
+    }()
 
     public lazy var callVisualizer = CallVisualizer(
         environment: .init(
--- a/GliaWidgetsTests/SecureConversations/ChatTranscript/TranscriptModel.Environment.Failing.swift
+++ b/GliaWidgetsTests/SecureConversations/ChatTranscript/TranscriptModel.Environment.Failing.swift
@@ -73,6 +73,10 @@ extension SecureConversations.TranscriptModel.Environment {
             fail("\(Self.self).createSendMessagePayload")
             return .mock()
         },
-        log: .failing
+        log: .failing,
+        maximumUploads: {
+            fail("\(Self.self).maximumUploads")
+            return 2
+        }
     )
 }
--- a/GliaWidgets.xcodeproj/project.pbxproj
+++ b/GliaWidgets.xcodeproj/project.pbxproj
@@ -437,6 +437,10 @@
 		8491AF652AAB281500CC3E72 /* ChatViewModelTests+CustomCard.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8491AF642AAB281500CC3E72 /* ChatViewModelTests+CustomCard.swift */; };
 		8491AF672AB8707600CC3E72 /* ChatViewModelTests+Transferring.swift in Sources */ = {isa = PBXBuildFile; fileRef = 8491AF662AB8707600CC3E72 /* ChatViewModelTests+Transferring.swift */; };
 		84A318A12869ECFC00CA1DE5 /* Unavailable.swift in Sources */ = {isa = PBXBuildFile; fileRef = 84A318A02869ECFC00CA1DE5 /* Unavailable.swift */; };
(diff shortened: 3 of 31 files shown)


GliaWidgets/Public/Glia/Glia.swift
```diff
@@ -59,6 +63,19 @@ public class Glia {
 
     var stringProvidingPhase: StringProvidingPhase = .notConfigured
 
+    var maximumUploads: Int {
+        var value = Self.maximumUploads
+        /// `MAXIMUM_UPLOADS_PER_MESSAGE` is used in acceptance tests to decrease the time
+        /// for uploading maximum number of files in a single message.
+        let env = ProcessInfo.processInfo.environment
```
igorkravchenko, line 70: Is it possible to create managed dependency for ProcessInfo? This will allow you to test reading values from it via unit tests.

igorkravchenko (commented):
Maybe it would be beneficial to create managed dependency for `ProcessInfo.processInfo`?

GliaWidgets/Public/Glia/Glia.swift
```diff
@@ -48,6 +48,10 @@ public protocol SceneProvider: AnyObject {
     func windowScene() -> UIWindowScene?
 }
 
+private extension Glia {
+    static let maximumUploads = 25
```
gersonnoboa, line 52: Where does this value come from? Is it a backend limitation or it's client imposed?

GliaWidgets/Public/Glia/Glia.swift
```diff
@@ -59,6 +63,19 @@ public class Glia {
 
     var stringProvidingPhase: StringProvidingPhase = .notConfigured
 
+    var maximumUploads: Int {
+        var value = Self.maximumUploads
+        /// `MAXIMUM_UPLOADS_PER_MESSAGE` is used in acceptance tests to decrease the time
+        /// for uploading maximum number of files in a single message.
+        let env = ProcessInfo.processInfo.environment
+        if let stringValue = env["MAXIMUM_UPLOADS_PER_MESSAGE"] {
+            /// Round to 1, because the attachment button doesn't turns to disabled state
```
gersonnoboa, line 72: I don't understand the issue deeply, but isn't this a workaround rather than fixing the button's behavior?

GliaWidgetsTests/SecureConversations/ChatTranscript/TranscriptModel.Environment.Failing.swift
```diff
@@ -73,6 +73,10 @@ extension SecureConversations.TranscriptModel.Environment {
             fail("\(Self.self).createSendMessagePayload")
             return .mock()
         },
-        log: .failing
+        log: .failing,
+        maximumUploads: {
+            fail("\(Self.self).maximumUploads")
+            return 2
```
gersonnoboa, line 79: Was 2 chosen in these mocks and failings for a reason?

GliaWidgets/Public/Glia/Glia.swift
EgorovEI: Honestly I don't know where it initially came from, we have this limitation for a long time

GliaWidgets/Public/Glia/Glia.swift
EgorovEI: There was no issue, I just want to make `maximumUploads` value adjustable from acceptance tests. Because now uploading 25 files just to check whether attachmentButton turns to disabled state is a bit overhead. With 25 files the test take 10min, but with 5 files - 3mins. 

The problem with rounding to `0` exists, if we pass 0 or negative value to `MAXIMUM_UPLOADS_PER_MESSAGE` from acceptance tests. If we round to `0`, attachmentButton will be still enabled, due to missing any update events on that time. Nevertheless, we don't need to have this value set to `0`. That's why I didn't fix "rounding to 0"

GliaWidgetsTests/SecureConversations/ChatTranscript/TranscriptModel.Environment.Failing.swift
EgorovEI: No reasons

GliaWidgets/Public/Glia/Glia.swift
EgorovEI: Added 19d21ea308800f1cf9868494d88118003477e0ca

GliaWidgets/Public/Glia/Glia.swift
yurii-glia: yes, it is backend limitation, however, this value doesn't come form backend.

GliaWidgets/Public/Glia/Glia.swift
yurii-glia: process information can't be changed after process is run, so I'd propose to change a little bit this getter and make it as `lazy var` and set value once instead of doing all this stuff for each get. would it make sense for you?

GliaWidgetsTests/SecureConversations/ChatTranscript/TranscriptModel.Environment.Failing.swift
yurii-glia: i suppose this is failing mock, so in fact the return value may be random, because unit test will be failed anyway.

rasmustautsglia (approved):

GliaWidgets/Sources/ProcessInfoHandling/ProcessInfoHandling.Implementation.swift
```diff
@@ -0,0 +1,7 @@
+import Foundation
```
igorkravchenko, line 1: I think it should be `ProcessInfoHandling.Live.swift`.

igorkravchenko (approved):
Looks good to me. Left one comment about **ProcessInfoHandling.Implementation.swift**. Since it just uses live instance, it should have **Live** suffix. If it would contain logic that has to be tested, then it had to have **Implementation** suffix.

GliaWidgets/Public/Glia/Glia.swift
EgorovEI: Made it `lazy` b1447e9e5f7bafe28a9551b5ce7f4dc6f06a67e3

EgorovEI:
!squash

PR #892, opened 2024-02-22
CallVisualizer views do not close upon ending engagement

rasmustautsglia:
MOB-3114

**Jira issue:**
https://glia.atlassian.net/browse/MOB-3114

**What was solved?**
CV was broken and the views didn't close when CV engagement had ended, leaving the views in a limbo where they didn't function but were present. This was caused by not observing interaction events, and checking if CV engagement was present before acting on specific interaction event. This was contradicting because if engagement had ended, current engagement couldn't be CV anymore.
**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/CallVisualizer/CallVisualizer.swift
+++ b/GliaWidgets/Sources/CallVisualizer/CallVisualizer.swift
@@ -86,7 +86,6 @@ public final class CallVisualizer {
         environment.log.prefixed(Self.self).info("Show Visitor Code Dialog")
         delegate?(.visitorCodeIsRequested)
         coordinator.showVisitorCodeViewController(by: .alert(source))
-        startObservingInteractorEvents()
     }
 
     /// Show VisitorCode embedded view for current Visitor.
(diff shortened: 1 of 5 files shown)


GliaWidgets/Sources/CallVisualizer/CallVisualizer.swift
```diff
@@ -151,14 +153,17 @@ extension CallVisualizer {
 extension CallVisualizer {
     func startObservingInteractorEvents() {
         environment.interactorProviding()?.addObserver(self) { [weak self] event in
+            if case .stateChanged(.ended(.byOperator)) = event {
```
EgorovEI, line 156: Could you please move this statement to `Glia.startObservingInteractorEvents()`?
This is needed because there we also send `onEvent` actions to integrators. And since we log `.info("Engagement ended")` in Glia, we can change wording there to `"Call visualizer engagement ended"` (this one is more descriptive) and remove logging the same event from `CallVisualizer. startObservingInteractorEvents()`

What do you think?

PR #893, opened 2024-02-22
Call visualizer views did not close on engagement ending

rasmustautsglia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3114

**What was solved?**
CV was broken and the views didn't close when CV engagement had ended, leaving the views in limbo where they didn't function but were present. This was caused by not observing interaction events, and checking if CV engagement was present before acting on specific interaction events. This was contradicting because if the engagement had ended, the current engagement couldn't be CV anymore.

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Public/Glia/Glia.swift
+++ b/GliaWidgets/Public/Glia/Glia.swift
@@ -104,7 +104,9 @@ public class Glia {
             proximityManager: environment.proximityManager,
             log: loggerPhase.logger,
             fetchSiteConfigurations: environment.coreSdk.fetchSiteConfigurations,
-            snackBar: environment.snackBar
+            snackBar: environment.snackBar,
+            coreSdk: environment.coreSdk,
+            operatorRequestHandlerService: operatorRequestHandlerService
         )
     )
     var rootCoordinator: EngagementCoordinator?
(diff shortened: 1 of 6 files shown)


EgorovEI (approved):

igorkravchenko (commented):
Is it obvious when this issue was introduced? 
Also if this is a bug fix, is it possible add unit test that ensures that the issue no longer happens?

rasmustautsglia:
> Is it obvious when this issue was introduced?
> Also if this is a bug fix, is it possible add unit test that ensures that the issue no longer happens?

I'm not exactly sure when it was broken but there were 2 factors:
1. `startObservingInteractorEvents` wasn't called at all, meaning that we didn't observe interaction events in CallVisualizer
2. Even if that method was called, nothing would have happened because the guard statement within that method would have returned early because of no engagement being present.

`test_screenSharingIsStoppedWhenCallVisualizerEngagementIsEnded` has been modified and will fail if this logic is broken again.

rasmustautsglia:
Unit test added, assuring that interactor observer is successfully added

[b6baf38](https://github.com/salemove/ios-sdk-widgets/pull/893/commits/b6baf38109083cc32e156537ad77154f7772e29f)

igorkravchenko (approved):

gersonnoboa (approved):

rasmustautsglia:
!squash

PR #896, opened 2024-02-26
Address video stream memory leak

igorkravchenko:
MOB-3139

**Jira issue:**
https://glia.atlassian.net/browse/MOB-3139

**What was solved?**
Address video stream memory leak by weakly capturing in visitor on hold callback.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


EgorovEI (approved):

GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
```diff
@@ -323,12 +323,12 @@ extension CallViewModel {
...
         action?(.switchToVideoMode)
         action?(.setLocalVideo(stream.getStreamView()))
-        stream.onHold = { isOnHold in
+        stream.onHold = { [weak stream] isOnHold in
             // onHold case is not handled deliberately because it
             // is handled elsewhere. This logic may need some adjustments
             // in future
             if !isOnHold {
-                stream.playVideo()
+                stream?.playVideo()
```
gersonnoboa, line 331: Should there be a log in case the stream has been deallocated? Could be useful for debugging maybe?

gersonnoboa (approved):

GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
igorkravchenko: I think it makes sense. But since we agreed on logs consistency between  mobile platforms, this has to be done in separate tickets for both platforms.

rasmustautsglia (approved):

yurii-glia (approved):

igorkravchenko:
!squash

PR #897, opened 2024-02-29
Enable visitor video by default

rasmustautsglia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3146

**What was solved?**
This PR enables visitor video by default unless the visitor is on hold, in which case visitor video will be enabled only after the hold is lifted.
**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
+++ b/GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
@@ -323,18 +323,8 @@ extension CallViewModel {
     private func showLocalVideo(with stream: CoreSdkClient.VideoStreamable) {
         action?(.switchToVideoMode)
         action?(.setLocalVideo(stream.getStreamView()))
-        stream.onHold = { [weak stream, environment] isOnHold in
-            // onHold case is not handled deliberately because it
-            // is handled elsewhere. This logic may need some adjustments
-            // in future
-            if !isOnHold {
-                if let stream = stream {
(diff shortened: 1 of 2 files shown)


yurii-glia (commented):
will not this change fix issue with play video stream by default but break on hold handling?

igorkravchenko (commented):
Also be aware about [this PR](https://github.com/salemove/ios-sdk-widgets/pull/896) merged recently.

rasmustautsglia:
> will not this change fix issue with play video stream by default but break on hold handling?

No, it works better because `steam.onHold` is nil by default so the block is not called upon launch, so we had no way playing the video. The good thing is, it doesn't break onHold logic because the observable observes onHold within Call object.

> Also be aware about https://github.com/salemove/ios-sdk-widgets/pull/896 merged recently.
I rebased the branch on development and checked. Will make adjustments to Call object and remove the implementation created in CallViewModel for weak onHold.

> Could we also add unit tests for that case?
Will create/edit tests for confidence

igorkravchenko (commented):
I think commit should contain ticket number and PR description should have proper link to ticket, so that Jira would index PR correctly.

rasmustautsglia:
> I think commit should contain ticket number and PR description should have proper link to ticket, so that Jira would index PR correctly.

Good catch!

igorkravchenko (approved):

PR #903, opened 2024-03-07
Define accessibilityId for video file view

EgorovEI:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2247

**What was solved?**
accessibilityId for video file view is needed for acceptance tests

**Release notes:**

 - [ ] Feature
 - [x] Ignore

### DIFF
--- a/GliaWidgets/Sources/Download/FileDownload.swift
+++ b/GliaWidgets/Sources/Download/FileDownload.swift
@@ -28,7 +28,7 @@ class FileDownload {
 
         var accessibilityString: String {
             switch self {
-            case .none: return "none"
+            case .none: return "download"
             case .downloading: return "downloading"
             case .downloaded: return "downloaded"
             case .error: return "error"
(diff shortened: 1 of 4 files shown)


EgorovEI:
!squash

GliaWidgets/Sources/Download/FileDownload.swift
```diff
@@ -28,7 +28,7 @@ class FileDownload {
 
         var accessibilityString: String {
             switch self {
-            case .none: return "none"
+            case .none: return "download"
```
gersonnoboa, line 31: Is there a reason for the deviation between the case and the accessbility string?

dukhovnyi (commented):
I'd propose to change commit message to more clear.
Perhaps something like this:
```
Update accessibility identifiers for video file view
Build accessibility identifiers based on file state and
sender of file (visitor or operator).

MOB-2247
```

would it make sense?

igorkravchenko (approved):

rasmustautsglia (approved):

gersonnoboa (approved):

PR #926, opened 2024-04-10
Add unit tests for disappeared bubble window fix and remove waiting

igorkravchenko:
MOB-3226

**Jira issue:**
https://glia.atlassian.net/browse/MOB-3226

**What was solved?**
Remove waiting strategy since it is unnecessary complex and not testable. The fix for now relies on proper window filtering. Unit tests are added.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

 - [x] Logging necessary for future troubleshooting of customer issues added?

**Screenshots:**


gersonnoboa (approved):
It would have been good to have the change for the waiting strategy in its own commit, in case something adverse is found and we have to revert the commit.

igorkravchenko:
> It would have been good to have the change for the waiting strategy in its own commit, in case something adverse is found and we have to revert the commit.

It is available in the original  commit d27e325d0e6de6bf17b1e622e3f2374b0cbdbbbc, so it will be possible to restore it. The main issue with it that I was not able to properly test it, because non async tests would affect async one.

rasmustautsglia (approved):

igorkravchenko:
!squash

PR #928, opened 2024-04-12
Add tests for CallVisualizerCoordinator

gersonnoboa:
The delegates inside the coordinators are not tested because each coordinator itself tests the delegate methods functioning through their own set of tests. Also, the coordinators in this class are all private, so testing it from outside would require moving the coordinators to internal properties, which I would say is not desirable.

MOB-2917

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgetsTests/Sources/CallVisualizer/Coordinator/CallVisualizerCoordinatorTests.swift
+++ b/GliaWidgetsTests/Sources/CallVisualizer/Coordinator/CallVisualizerCoordinatorTests.swift
@@ -1,0 +1,139 @@
+import Foundation
+import XCTest
+@testable import GliaWidgets
+
+final class CallVisualizerCoordinatorTests: XCTestCase {
+    var coordinator: CallVisualizer.Coordinator!
+    var viewController = UIViewController()
+
+    override func setUp() {
+        coordinator = .init(environment: .mock)
+    }
+
+    func test_showVisitorCodeViewController() throws {
+        let scene = try XCTUnwrap(UIApplication.shared.connectedScenes.first as? UIWindowScene)
+        let window = scene.windows.first
+        let oldRootViewController = window?.rootViewController
+        window?.rootViewController = viewController
+        defer { window?.rootViewController = oldRootViewController }
+
+        coordinator.showVisitorCodeViewController(by: .alert(viewController))
+        
+        XCTAssertTrue(viewController.presentedViewController is CallVisualizer.VisitorCodeViewController)
+    }
+
+    func test_handleAcceptedUpgrade() {
+        var calledEvents: [CallVisualizer.Coordinator.DelegateEvent] = []
+        coordinator.environment.eventHandler = { calledEvents.append($0) }
+        coordinator.handleAcceptedUpgrade()
+
+        XCTAssertTrue(calledEvents.contains(.maximized))
+    }
+
+    func test_handleEngagementRequestAccepted() throws {
+        let site = CoreSdkClient.Site(
+            id: .mock, defaultOperatorPicture: nil,
+            alwaysUseDefaultOperatorPicture: false,
+            allowedFileSenders: try .mock(),
+            maskingRegularExpressions: [],
+            visitorAppDefaultLocale: "",
+            mobileConfirmDialogEnabled: false,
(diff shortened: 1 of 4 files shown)


GliaWidgetsTests/Sources/CallVisualizer/Coordinator/CallVisualizerCoordinatorTests.swift
```diff
@@ -0,0 +1,141 @@
...
+            mobileObservationVideoFps: try videoFps(),
+            mobileObservationEnabled: true
+        )
+
+        coordinator.environment.fetchSiteConfigurations = { callback in
+            callback(.success(site))
+        }
+
+        let answer = Command<Bool> { boolean in
+            XCTAssertTrue(boolean)
```
igorkravchenko, line 51: This assert will only be called if `answer` command is called. To avoid missing this command call in unit tests I think it would be better to introduce empty array of bools, and append `boolean ` to it and assert it out of the closure like this:

```
var answers: [Bool] = []
let answer = Command<Bool> { boolean in
  answers.append(boolean)
}

XCTAssertEqual(answers, [true])
```

GliaWidgetsTests/Sources/CallVisualizer/Coordinator/CallVisualizerCoordinatorTests.swift
```diff
@@ -0,0 +1,141 @@
...
```
igorkravchenko, line 84: Same idea about asserting in callback.

igorkravchenko (changes requested):
Added suggestions about avoiding asserting inside closures.

GliaWidgetsTests/Sources/CallVisualizer/Coordinator/CallVisualizerCoordinatorTests.swift
gersonnoboa: Fixed 21298a0a635189b8bb2191b3dc1b34e4e621e7be

GliaWidgetsTests/Sources/CallVisualizer/Coordinator/CallVisualizerCoordinatorTests.swift
gersonnoboa: I changed it here, because I realized that indeed this assertion wasn't even running, but the problem is that to run it, I need to manually press the AlertViewController because the answer command depends on what the user presses. Thus I believe that because showing the AlertViewController is what the coordinator takes care of, then that should be enough for the test. 21298a0a635189b8bb2191b3dc1b34e4e621e7be let me know if it makes sense.

igorkravchenko: Make sense.

GliaWidgetsTests/Sources/CallVisualizer/Coordinator/CallVisualizerCoordinatorTests.swift
```diff
@@ -0,0 +1,140 @@
...
+
+        XCTAssertTrue(coordinator.environment.presenter.getInstance()?.presentedViewController is AlertViewController)
+    }
+
+    func test_end() {
+        var isCalled = false
+        
+        coordinator.environment.screenShareHandler.stop = { _ in
+            isCalled = true
+        }
```
yurii-glia, line 96: would it be beneficial to have a counter here to know about multiple calls if this happens?
```swift
        var stopCallCounter = 0
        
        coordinator.environment.screenShareHandler.stop = { _ in
            stopCallCounter += 1
        }
        // ...
        XCAssertTrue(stopCallCounter)
```

gersonnoboa: Fixed 52f4f8d4ae0dc6303b92c6934efda5240eeaa6e2

yurii-glia (approved):

igorkravchenko (approved):

gersonnoboa:
!squash

PR #935, opened 2024-05-08
Add ability to flip camera from front to back

igorkravchenko:
MOB-3290

**Jira issue:**
https://glia.atlassian.net/browse/MOB-3290

**What was solved?**
Add ability to flip camera from front to back.
Tests, styling, accessibility are to be added in subsequent PRs.

Note this PR is on hold because it awaits Core SDK release with Flip Camera API.

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**


GliaWidgets/Sources/ViewModel/Call/CallViewModel+CameraFlip.swift
```diff
@@ -0,0 +1,48 @@
...
+        }
+
+        do {
+            let cameraDeviceManager = try environment.cameraDeviceManager()
+            let devices = cameraDeviceManager.cameraDevices().filter { $0.facing == .front || $0.facing == .back }
+            // Camera flip only makes sense when there are at least two camera devices.
+            guard devices.count >= 2 else {
+                flipCameraCallback = nil
+                return
+            }
```
EgorovEI, line 18: I'd add array of facing cases and rely on its count. Does it make sense for you?
```suggestion
            let cameraDeviceManager = try environment.cameraDeviceManager()
            let cameraDeviceFacingTypes: [PutCorrectFacingType] = [.front, .back]
            let devices = cameraDeviceManager.cameraDevices().filter { cameraDeviceFacingTypes.contains($0) }
            // Camera flip only makes sense when there are at least two camera devices.
            guard devices.count >= cameraDeviceFacingTypes.count else {
                flipCameraCallback = nil
                return
            }
```

EgorovEI (approved):

GliaWidgets/Sources/Extensions/UIButton+Extensions.swift
```diff
@@ -0,0 +1,24 @@
...
+
+        if let context = UIGraphicsGetCurrentContext() {
+            context.setFillColor(color.cgColor)
+            context.fill(CGRect(origin: .zero, size: minimumSize))
+        }
+
+        let colorImage = UIGraphicsGetImageFromCurrentImageContext()
+        UIGraphicsEndImageContext()
+
+        self.clipsToBounds = true
```
gersonnoboa, line 21: NIT: I personally like the use of `self` in this situation, but I believe it doesn't follow the rest of the codebase style.

Edit: I do see usages of this so it's up to you what you do.

GliaWidgets/Sources/Extensions/UIImage+Extensions.swift
```diff
@@ -54,4 +54,24 @@ extension UIImage {
             orientation: self.imageOrientation
         )
     }
+
+    func image(withTintColor color: UIColor) -> UIImage {
```
gersonnoboa, line 58: Is there a reason why this is needed instead of setting the tintColor of the imageView ?

GliaWidgets/Sources/View/Call/Video/FlipCameraButton.swift
```diff
@@ -0,0 +1,126 @@
...
+        tap: nil
+    ) {
+        didSet {
+            renderProps()
+        }
+    }
+
+    private var renderedActiveState: FlipCameraButtonStyle.State = .nop {
+        didSet {
+            guard renderedActiveState != oldValue else {
```
gersonnoboa, line 38: I don't understand, if the renderedActiveState starts with 0 let's say, and you change the value from 0 to 1, it means that renderedActiveState (0) is the same as oldValue (0), so it will return and not render the state 1.

GliaWidgets/Sources/ViewModel/Call/CallViewModel+CameraFlip.swift
```diff
@@ -0,0 +1,48 @@
...
+                if let currentDevice = cameraDeviceManager.currentCameraDevice(),
+                    // Find corresponding index for active camera
+                    let currentIndex = devices.firstIndex(of: currentDevice) {
+                    // and increase it by one.
+                    let nextIndex = currentIndex.advanced(by: 1)
+                    // If next index is valid use it for next device selection, otherwise use first device.
+                    let nextDevice = devices.indices.contains(nextIndex) ? devices[nextIndex] : devices[0]
+                    cameraDeviceManager.setCameraDevice(nextDevice)
+                } else {
+                    environment.log.warning("Unable to change camera device.")
```
gersonnoboa, line 31: Maybe here you could add more info, like the number of cameras on the device and to which index we are going.

GliaWidgets/Sources/ViewModel/Call/CallViewModel+CameraFlip.swift
```diff
@@ -0,0 +1,48 @@
...
+
+            flipCameraCallback = Cmd { [environment] in
+                // Get currently active camera device.
+                if let currentDevice = cameraDeviceManager.currentCameraDevice(),
+                    // Find corresponding index for active camera
+                    let currentIndex = devices.firstIndex(of: currentDevice) {
+                    // and increase it by one.
+                    let nextIndex = currentIndex.advanced(by: 1)
+                    // If next index is valid use it for next device selection, otherwise use first device.
+                    let nextDevice = devices.indices.contains(nextIndex) ? devices[nextIndex] : devices[0]
```
gersonnoboa, line 28: Cool, this would mean that if there's a third camera at some point, this functionality wouldn't need changes, correct? Although then the name "flip" wouldn't be correct.

GliaWidgets/Sources/Extensions/UIImage+Extensions.swift
igorkravchenko: Yes, the reason is that `UIButton.setImage(_ image: UIImage?, for state: UIControl.State)` requires `UIImage` instance. Since image colour is specified via styling for different button states, multiple images with different colors should be provided.

GliaWidgets/Sources/View/Call/Video/FlipCameraButton.swift
igorkravchenko: Yes, the idea is to avoid excessive re-rendering if value is the same, so we render if new value is different from old one. That is especially important, because we rely on background image generation with specific color for button background states and the same true for icon image used for button.

gersonnoboa: Okay I understand now, `renderedActiveState` at this point already has the new value. For some reason when I read it, I thought that at this point it would have the old value, and thus checking the old value against `oldValue` didn't make much sense. Apologies.

GliaWidgets/Sources/ViewModel/Call/CallViewModel+CameraFlip.swift
igorkravchenko: On one hand it makes sense, to limiting camera devices to have both `front` and back `facing` to allow true meaning of flip. On the other hand we would discard changing cameras if there are two `back` cameras while `front` one is unavailable and vice versa. I would prefer for now to allow changing of camera even if there are only two or more same facing devices. But it would be good to hear from others about it. cc: @AndriiHorishniiMOC @yurii-glia @gersonnoboa @rasmustautsglia

GliaWidgets/Sources/ViewModel/Call/CallViewModel+CameraFlip.swift
igorkravchenko: Added more information to this message b1dfae82fddc5847f01d1779564f153e1482b067.

GliaWidgets/Sources/ViewModel/Call/CallViewModel+CameraFlip.swift
igorkravchenko: Yes, still for now there are only `front` and `back` facing devices that are taken into account. Technically if two or more cameras will have only `front` facing,  toggling will also happen between them. But to be consistent with feature name and android implementation I think we can leave 'flip' name for now.

GliaWidgets/Sources/ViewModel/Call/CallViewModel+CameraFlip.swift
gersonnoboa: I also like the approach of going through the cameras instead even though the meaning of "flipping" gets thrown away by this implementation. But right now, if Apple decides to switch between two cameras at the back for some reason, we would be able to support it without changing the SDK, which is a nice thing to have.

GliaWidgets/Sources/Extensions/UIButton+Extensions.swift
igorkravchenko: I think using `self` explicitly brings more clarity.

gersonnoboa (approved):

GliaWidgets/Sources/ViewModel/Call/CallViewModel+CameraFlip.swift
yurii-glia: Agree with @igorkravchenko  & @gersonnoboa

igorkravchenko:
!squash

GliaWidgets/Sources/ViewModel/Call/CallViewModel+CameraFlip.swift
```diff
@@ -0,0 +1,86 @@
...
+                // Find corresponding index for active camera
+                let currentIndex = currentDevice.flatMap { devices.firstIndex(of: $0) }
+                // and increase it by one.
+                let nextIndex = currentIndex.flatMap { $0.advanced(by: 1) }
+
+                if let nextIndex {
+                    // If next index is valid use it for next device selection, otherwise use first device.
+                    let nextDevice = devices.indices.contains(nextIndex) ? devices[nextIndex] : devices[0]
+                    cameraDeviceManager.setCameraDevice(nextDevice)
+                    // We need to call `setCameraFlipVisible` to let
```
kadriiiii, line 51:
```suggestion
                    // Call `setCameraFlipVisible` to let
```

GliaWidgets/Sources/ViewModel/Call/CallViewModel+CameraFlip.swift
```diff
@@ -0,0 +1,86 @@
...
+                    // accessibility label be propagated to button via
+                    // props.
```
kadriiiii, line 53:
```suggestion
                    // properties.
```

GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
```diff
@@ -325,6 +325,17 @@ extension CallViewModel {
         action?(.setLocalVideo(stream.getStreamView()))
         guard call.isVisitorOnHold.value == false else { return }
         stream.playVideo()
+        // For now camera manager is available only when corresponding
```
kadriiiii, line 328:
```suggestion
        // For now, camera manager is available only when corresponding
```

GliaWidgets/Sources/ViewModel/Call/CallViewModel.swift
```diff
@@ -325,6 +325,17 @@ extension CallViewModel {
...
+        // local stream is available, that's why camera button presence
```
kadriiiii, line 329: "that's" to "that is" to be consistent with support docs.

```suggestion
        // local stream is available, that is why camera button presence
```

GliaWidgets/Sources/Extensions/UIButton+Extensions.swift
```diff
@@ -0,0 +1,24 @@
+import UIKit
+
+extension UIButton {
+    /// Sets the background color to use for the specified button state.
```
kadriiiii, line 4:
```suggestion
    /// Sets the background color to be used for the specified button state.
```

GliaWidgets/Sources/View/Call/Video/VideoStreamView.swift
```diff
@@ -76,4 +114,50 @@ class VideoStreamView: UIView {
...
+                accessibility: .nop
+            )
+        }
+
+        flipButton.props = props
+    }
+
+    func renderLocalVideoAccessibility() {
+        // We need to change accessibility behaviour
+        // in case if flip camera button is present,
```
kadriiiii, line 142:
```suggestion
        // in case the flip camera button is present,
```

GliaWidgets/Sources/View/Call/Video/VideoStreamView.swift
```diff
@@ -76,4 +114,50 @@ class VideoStreamView: UIView {
...
+                tap: nil,
+                accessibility: .nop
+            )
+        }
+
+        flipButton.props = props
+    }
+
+    func renderLocalVideoAccessibility() {
+        // We need to change accessibility behaviour
```
kadriiiii, line 141:
```suggestion
        // Change accessibility behaviour
```

GliaWidgets/Sources/View/Call/Video/VideoStreamView.swift
```diff
@@ -76,4 +114,50 @@ class VideoStreamView: UIView {
...
+        // in case if flip camera button is present,
+        // to allow it to be recognizable by screen reader.
+        if flipCameraAccessibilityLabelWithTap != nil {
+            // Parent container can not be accessibility element,
+            // if child needs to be accessibility element as well,
+            // but siblings within same containers can.
+            self.isAccessibilityElement = false
+            // So we make `streamView` (which is sibling to
```
kadriiiii, line 149:
```suggestion
            // Thus make `streamView` (which is sibling to
```

GliaWidgets/Sources/View/Call/Video/VideoStreamView.swift
```diff
@@ -76,4 +114,50 @@ class VideoStreamView: UIView {
...
+            // flip camera button) an accessibility element.
+            streamView?.isAccessibilityElement = true
+            // Also we forward `accessibilityLabel` from parent to
```
kadriiiii, line 152:
```suggestion
            // Also, forward `accessibilityLabel` from parent to
```

GliaWidgets/Sources/View/Call/Video/VideoStreamView.swift
```diff
@@ -76,4 +114,50 @@ class VideoStreamView: UIView {
...
+            // `streamView`, so that it would be read instead.
+            streamView?.accessibilityLabel = accessibilityLabel
+            // Finally we assign `streamView` and `flipButton` to
```
kadriiiii, line 155:
```suggestion
            // Finally assign `streamView` and `flipButton` to
```

kadriiiii (commented):
Few styling suggestions :)

kadriiiii (approved):
Thank you Igor!

yurii-glia:
@igorkravchenko  there tests failed, would you be able to fix them?

igorkravchenko:
> @igorkravchenko there tests failed, would you be able to fix them?

@yurii-glia this PR refers to the source code of unreleased Core SDK, so it is expected that tests/build will fail. Once we release Core SDK with flip camera, I assume that tests will start to pass.

yurii-glia:
@igorkravchenko  Converted to Draft while Core SDK release.

igorkravchenko:
!squash

PR #939, opened 2024-05-15
Update GliaCoreSDK to 1.4.4

EgorovEI:


GliaWidgets.podspec
```diff
@@ -19,5 +19,5 @@ Pod::Spec.new do |s|
   }
   s.exclude_files         = ['GliaWidgets/Window/**']
 
-  s.dependency 'GliaCoreSDK', '1.4.3'
+  s.dependency 'GliaCoreSDK', '1.4.4'
```
yurii-glia, line 22: I suppose similar change should be done in `templates/GliaWidgets.podspec`. Does it?

EgorovEI: Correct, and it's done

yurii-glia (approved):

PR #950, opened 2024-05-20
Fix crash occurring on onHold event during LO

EgorovEI:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3332

**What was solved?**
When the visitor is put on-hold during LO, the crash occurs, because `OnHoldOverlayVisualEffectView.draw` method is called on background thread by `layer?.render(in:)` in `WindowCapturer.native`. Since we need to keep screen capturing on background thread, `draw` method execution was wrapped into main queue. This issue is related only to native implementation for WindowCapturer, and is not reproducible for flutter side.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

### DIFF
--- a/GliaWidgetsTests/Sources/OnHoldOverlayVisualEffectView/OnHoldOverlayVisualEffectViewTests.swift
+++ b/GliaWidgetsTests/Sources/OnHoldOverlayVisualEffectView/OnHoldOverlayVisualEffectViewTests.swift
@@ -1,0 +1,20 @@
+import XCTest
+@testable import GliaWidgets
+
+final class OnHoldOverlayVisualEffectViewTests: XCTestCase {
+
+    func test_overlayViewDrawIsExecutedOnMainQueue() {
+        enum Call { case mainQueueAsync }
+        var calls: [Call] = []
+        var gcd = GCD.failing
+        gcd.mainQueue.async = { _ in
+            calls.append(.mainQueueAsync)
+        }
+        let environment = OnHoldOverlayVisualEffectView.Environment(gcd: gcd)
+        let view = OnHoldOverlayVisualEffectView(environment: environment)
+
+        view.draw(.zero)
+
+        XCTAssertEqual(calls, [.mainQueueAsync])
+    }
+}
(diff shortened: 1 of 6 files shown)


EgorovEI:
I have not found another way to fix the crash. If you have any ideas, please share them

igorkravchenko (commented):
Is possible to add test that just asserts that `environment.gcd.mainQueue.async` gets called?

gersonnoboa (approved):

GliaWidgetsTests/Sources/OnHoldOverlayVisualEffectView/OnHoldOverlayVisualEffectViewTests.swift
```diff
@@ -0,0 +1,23 @@
...
+final class OnHoldOverlayVisualEffectViewTests: XCTestCase {
+
+    func test_overlayViewDrawIsExecutedOnMainQueue() {
+        enum Call { case mainQueueAsync, globalQueueAsync }
+        var calls: [Call] = []
+        var gcd = GCD.failing
+        gcd.mainQueue.async = { _ in
+            calls.append(.mainQueueAsync)
+        }
+        gcd.globalQueue.async = { _ in
```
igorkravchenko, line 13: Why `gcd.globalQueue.async` also is overwritten if it is not used?

igorkravchenko (commented):
Left on comment regarding GCD global queue usage.

GliaWidgetsTests/Sources/OnHoldOverlayVisualEffectView/OnHoldOverlayVisualEffectViewTests.swift
EgorovEI: Agree that globalQueue check is redundant. I added this check to be sure that it's not called.

EgorovEI: Fixed 3c62e8478011e44cdeec2e7fa053456067aeef94

igorkravchenko (approved):

EgorovEI:
!squash

PR #954, opened 2024-05-21
Fix local video shape after device orientation change

igorkravchenko:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3338

**What was solved?**
Fix local video shape after device orientation change by evaluating current device orientation and deriving width and height for the video.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/View/Call/CallView.swift
+++ b/GliaWidgets/Sources/View/Call/CallView.swift
@@ -418,13 +418,15 @@ class CallView: EngagementView {
 // MARK: Local Video
 
 extension CallView {
+    static let localVideoSideMultiplier = 0.3
+
     private func setLocalVideoFrame(isVisible: Bool) {
         if isVisible {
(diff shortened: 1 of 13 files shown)


GliaWidgets/Sources/View/Call/CallView.swift
```diff
@@ -440,17 +442,30 @@ extension CallView {
...
+            height = min(size.width, size.height)
+        } else {
+            width = min(size.width, size.height)
+            height = max(size.width, size.height)
+        }
+
+        let newSize = CGSize(
+            width: width * 0.3,
+            height: height * 0.3
         )
```
EgorovEI, line 461: looks like it can be replaced with 
```swift
let newSize = localVideoSize(
   for: environment.uiDevice.orientation(),
   from: size
)

gersonnoboa (approved):

PR #958, opened 2024-05-23
Remove `singleChoice` check for Custom Card

EgorovEI:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3359

**What was solved?**
Since `attachment.type == .singleChoice` check for Custom Cards does not exist on Android side, that was decided to align the platform and remove it from iOS side. So now, evaluation of cardType does not take into account `attachment.type` if metadata exists.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

### DIFF
--- a/GliaWidgetsTests/ChatMessage/ChatMessageTests.swift
+++ b/GliaWidgetsTests/ChatMessage/ChatMessageTests.swift
@@ -23,4 +23,15 @@ final class ChatMessageTests: XCTestCase {
         )
         XCTAssertEqual(msg.cardType, .customCard)
     }
+
+    func testCardType__metadataWithoutSingleChoice() throws {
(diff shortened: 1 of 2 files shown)


GliaWidgetsTests/ChatMessage/ChatMessageTests.swift
```diff
@@ -23,4 +23,15 @@ final class ChatMessageTests: XCTestCase {
         )
         XCTAssertEqual(msg.cardType, .customCard)
     }
+
```
igorkravchenko, line 26: Maybe it makes sense to also add test for `. choiceCard` as well, since it is a part of the changes?

igorkravchenko (commented):
Looks good, left one question about additional test.

gersonnoboa (approved):

igorkravchenko (approved):

PR #959, opened 2024-05-24
Filter out 401 error when sending a message failed

rasmustautsglia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3357

**What was solved?**
When sending a message failed, Internal Error was always returned, making it impossible to determine the cause and, if needed, skip any action. Since 401 errors are handled from elsewhere, 401 case needs to be skipped.
**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
+++ b/GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
@@ -458,7 +458,7 @@ extension ChatViewModel {
             var messagePayload = outgoingMessage.payload
             messagePayload.content = messageTextTemp
             messagePayload.attachment = attachment
-            interactor.send(messagePayload: messagePayload) { [weak self, environment] result in
+            interactor.send(messagePayload: messagePayload) { [weak self] result in
                 guard let self else { return }
 
                 switch result {
@@ -473,17 +473,10 @@ extension ChatViewModel {
                         )
                         self.action?(.scrollToBottom(animated: true))
                     }
-
-                case .failure:
-                    environment.log.prefixed(Self.self).info("Message send exception")
-                    environment.log.prefixed(Self.self).info("Show Unexpected error Dialog")
-                    self.showAlert(
-                        with: self.alertConfiguration.unexpectedError,
-                        dismissed: nil
-                    )
+                case let .failure(error):
+                    handleSendMessageError(error)
                 }
             }
-
         case .enqueued:
             handle(pendingMessage: outgoingMessage)
 
@@ -493,6 +486,20 @@ extension ChatViewModel {
         }
 
         messageText = ""
+    }
+
+    func handleSendMessageError(_ error: CoreSdkClient.GliaCoreError) {
+        switch error.error {
+        case let authError as CoreSdkClient.Authentication.Error where authError == .expiredAccessToken:
+            break
+        default:
+            environment.log.prefixed(Self.self).info("Message send exception")
+            environment.log.prefixed(Self.self).info("Show Unexpected error Dialog")
+            showAlert(
+                with: alertConfiguration.unexpectedError,
+                dismissed: nil
+            )
+        }
     }
 
     func handle(pendingMessage: OutgoingMessage) {
--- a/GliaWidgetsTests/Sources/InteractorTests.swift
+++ b/GliaWidgetsTests/Sources/InteractorTests.swift
@@ -423,6 +423,40 @@ class InteractorTests: XCTestCase {
         XCTAssertEqual(callbacks, [.sendMessageWithAttachment])
     }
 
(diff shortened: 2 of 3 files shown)


GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -474,13 +474,29 @@ extension ChatViewModel {
...
+                        default:
+                            environment.log.prefixed(Self.self).info("Message send exception")
+                            environment.log.prefixed(Self.self).info("Show Unexpected error Dialog")
+                            self.showAlert(
+                                with: self.alertConfiguration.unexpectedError,
+                                dismissed: nil
+                            )
+                        }
+                    default:
+                        print("WAZAAA")
```
gersonnoboa, line 492: https://www.youtube.com/watch?v=lgJsNW_zNUI

gersonnoboa (commented):
Also now this method is very long so SwiftLint is complaining

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
rasmustautsglia: Fixed [a727937](https://github.com/salemove/ios-sdk-widgets/pull/959/commits/a727937c6cd96f2f241718c9ba03862639769906)

gersonnoboa (approved):

EgorovEI (approved):

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -495,6 +490,30 @@ extension ChatViewModel {
...
+                )
+            }
+        default:
+            environment.log.prefixed(Self.self).info("Message send exception")
+            environment.log.prefixed(Self.self).info("Show Unexpected error Dialog")
+            self.showAlert(
+                with: self.alertConfiguration.unexpectedError,
+                dismissed: nil
+            )
+        }
```
yurii-glia, line 514: could we simplify this switch? perhaps to something like this?

```swift
        switch error.error {
        case let authenticationError as CoreSdkClient.Authentication.Error where authenticationError == .expiredAccessToken:
                break
        default:
            environment.log.prefixed(Self.self).info("Message send exception")
            environment.log.prefixed(Self.self).info("Show Unexpected error Dialog")
            self.showAlert(
                with: self.alertConfiguration.unexpectedError,
                dismissed: nil
            )
        }
```

rasmustautsglia: fixed [6a2873a](https://github.com/salemove/ios-sdk-widgets/pull/959/commits/6a2873a734ab1214de900b20b3908782b8fc09b3)

rasmustautsglia: Can't and will not merge before CoreSDK release because `expiredAccessToken` has an associated value. However for testing purposes I had Core connected directly and this issue didn't emerge

igorkravchenko (commented):
I think there should be also corresponding test that ensures that the error is filtered out.

rasmustautsglia:
Unit test added [178b6f7](https://github.com/salemove/ios-sdk-widgets/pull/959/commits/178b6f741edc31bfa9e093ea0e5f9f5747177c6e)

GliaWidgetsTests/Sources/InteractorTests.swift
```diff
@@ -423,6 +423,32 @@ class InteractorTests: XCTestCase {
...
+            result(.failure(expectedError))
+        }
+        let interactor = Interactor.mock(environment: interactorEnv)
+
+        interactor.send(messagePayload: .mock(content: "mock-message")) { result in
+            switch result {
+            case .success:
+                XCTFail()
+            case let .failure(error):
+                switch error.error {
```
igorkravchenko, line 442: Since this assertion happens in callback, there's potential probability that this callback will be never called or called later where test already has finished its execution. I would suggest to store result in the array declared on line 436, the assert contents of array out of callback. This will assert that callback is called expected number of times, and also that result is also expected.

igorkravchenko (commented):
Left one more comment regarding unit test.

GliaWidgetsTests/Sources/InteractorTests.swift
rasmustautsglia: Added callbacks to improve the test [a9ed2df](https://github.com/salemove/ios-sdk-widgets/pull/959/commits/a9ed2dfd0e961615034cc77c5e4e3e6c682bc6d5)

igorkravchenko (approved):

rasmustautsglia:
!squash

PR #968, opened 2024-06-03
Create Alert Manager

rasmustautsglia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3373

**What was solved?**
Because of 409 error handling via global handler, there were many duplicate alerts being presented. Alert Manager is a solution going forward, that gets rid of that problem, but also allows further improvements in error handling, without breaking UI.

Alert Manager brings everything related to alerts to one place. With a single interface, every alert will be handled by present() method. An alert can be presented in a root view, or in a separate key window, depending on the use case.

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**


GliaWidgets/Sources/AlertManager/AlertInputType.swift
```diff
@@ -0,0 +1,68 @@
...
+        case (.operatorEndedEngagement, .operatorEndedEngagement):
+            return true
+        case (.unsupportedGvaBroadcastError, .unsupportedGvaBroadcastError):
+            return true
+        case (.unavailableMessageCenter, .unavailableMessageCenter):
+            return true
+        case (.unavailableMessageCenterForBeingUnauthenticated, .unavailableMessageCenterForBeingUnauthenticated):
+            return true
+        case (.mediaUpgrade, .mediaUpgrade):
+            return false
```
yurii-glia, line 59: I think it would be nice to compare `operators` & `offer`, since a few media upgrades may happen at the same time. should we?

GliaWidgets/Sources/AlertManager/AlertManager.swift
```diff
@@ -0,0 +1,140 @@
...
+        self.environment = environment
+        self.viewFactory = viewFactory
+        composer = .init(
+            environment: .init(log: environment.log),
+            theme: viewFactory.theme
+        )
+    }
+}
+
+// MARK: - Public Methods
```
yurii-glia, line 25: That methods are not public.

GliaWidgets/Sources/AlertManager/AlertManager.swift
```diff
@@ -0,0 +1,140 @@
...
+        alertViewController = nil
+        alertWindow?.isHidden = true
+        alertWindow = nil
+        currentAlert = nil
+    }
+}
+
+// MARK: - KeyWindow
+private extension AlertManager {
+    func createAlertWindow() {
```
yurii-glia, line 116: could elaborate a bit, for what purposes would we like to have a separate window for alerts?

rasmustautsglia: These are for mediaUprades and screensharing that need to be displayed in any view of the app, not limited to our Glia views

GliaWidgets/Sources/AlertManager/AlertInputType.swift
rasmustautsglia: I returned them as false(not the same) regardless if they are actually different or not. The reason:

Let's say media upgrade is sent and then right after that another same media upgrade is sent. Since the same alerts are filtered/blocked, the most recent one is not shown. If I accept the offer, then nothing happens because GliaHub has already cancelled that request, and waits for the response on the most recent one. That is why we can't filter them out. And the same goes for screensharing offer.

GliaWidgets/Sources/AlertManager/AlertManager.swift
rasmustautsglia: fixed [763c33b](https://github.com/salemove/ios-sdk-widgets/pull/968/commits/763c33b953fde15ccfb8400a27fb76f00b1e2646)

GliaWidgets/Sources/AlertManager/AlertManager.swift
rasmustautsglia: See the implementation here https://github.com/salemove/ios-sdk-widgets/blob/master/GliaWidgets/Sources/OperatorRequestHandlerService/OperatorRequestHandlerService.swift

GliaWidgets/Sources/AlertManager/AlertInputType.swift
yurii-glia: agree with everything your said and I suppose if two the same alerts come to this == function they will not be filtered out because they all fully the same, so the same alert may appear twice. So, would it make sense to compare `operators` and `offer` also to remove duplications in filter function?

rasmustautsglia: I don't think it makes sense to compare `operators` and `offer` because we intentionally don't want to filter out duplications. The reason: If we do filter out the most recent one (i.e mediaUpgrade), then we are displaying the old one. If we accept the old one, then nothing happens on GliaHub side, because Gliahub only cares about the latest request. To guarantee that the latest is always shown, we have to let through all mediaUpgrades and screenShareOffers in order to guarantee proper functionality.

GliaWidgetsTests/SecureConversations/Welcome/SecureConversations.WelcomeViewModelSpec.swift
```diff
@@ -688,10 +697,11 @@ extension SecureConversationsWelcomeViewModelTests {
         )
 
         XCTAssertEqual(viewModel.availabilityStatus, .unavailable(.unauthenticated))
-        XCTAssertEqual(
-            alertConfiguration?.message,
-            viewModel.environment.alertConfiguration.unavailableMessageCenterForBeingUnauthenticated.message
-        )
+        if case .unavailableMessageCenterForBeingUnauthenticated = alertInputType {
```
igorkravchenko, line 700: I think this can be simplified by introducing `Bool` constant before `if/else` and assign value to it In `if/else` branches, then `XCTAssertTrue` it once.

igorkravchenko (approved):
Left some minor comments regarding documentation and unit tests.

GliaWidgets/Sources/AlertManager/AlertManager.swift
```diff
@@ -0,0 +1,212 @@
...
+    }
+
+    func isAlertReplacable(offer: Replaceable) -> Bool {
+        guard let alertViewController else { return true }
+        switch alertViewController.isReplaceable(with: offer) {
+        case true:
+            return true
+        case false:
+            return false
+        }
```
EgorovEI, line 174: Does it make sense?
```suggestion
        alertViewController.isReplaceable(with: offer) ?? true
```

GliaWidgetsTests/Coordinator/RootCoordinator.Environment.Failing.swift
```diff
@@ -89,12 +89,12 @@ extension EngagementCoordinator.Environment {
...
         snackBar: .failing,
-        operatorRequestHandlerService: .failing,
         maximumUploads: {
             fail("\(Self.self).maximumUploads")
             return 2
         },
         cameraDeviceManager: { .failing },
-        flipCameraButtonStyle: .nop
+        flipCameraButtonStyle: .nop,
+        alertManager: .mock()
```
EgorovEI, line 98: I'd recommend to create `failing` instance

rasmustautsglia: added [c08f255](https://github.com/salemove/ios-sdk-widgets/pull/968/commits/c08f255d539b691b212f88c3695b2f272a154aad)

GliaWidgets/Sources/AlertManager/AlertManager.swift
rasmustautsglia: fixed [c08f255](https://github.com/salemove/ios-sdk-widgets/pull/968/commits/c08f255d539b691b212f88c3695b2f272a154aad)

GliaWidgetsTests/SecureConversations/Welcome/SecureConversations.WelcomeViewModelSpec.swift
rasmustautsglia: fixed [c08f255](https://github.com/salemove/ios-sdk-widgets/pull/968/commits/c08f255d539b691b212f88c3695b2f272a154aad)

GliaWidgets/Sources/AlertManager/AlertTypeComposer.swift
```diff
@@ -0,0 +1,307 @@
+import UIKit
+
+extension AlertManager {
+    struct AlertTypeComposer {
```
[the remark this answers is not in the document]
rasmustautsglia: added [c08f255](https://github.com/salemove/ios-sdk-widgets/pull/968/commits/c08f255d539b691b212f88c3695b2f272a154aad)

GliaWidgets/Sources/AlertManager/AlertManager.swift
```diff
@@ -0,0 +1,212 @@
+import UIKit
+import Foundation
+
+final class AlertManager {
```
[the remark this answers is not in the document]
rasmustautsglia: added [c08f255](https://github.com/salemove/ios-sdk-widgets/pull/968/commits/c08f255d539b691b212f88c3695b2f272a154aad)

GliaWidgets/Sources/AlertManager/Alert/AlertViewController.swift
```diff
@@ -185,6 +134,19 @@ class AlertViewController: UIViewController, Replaceable {
                 accepted: accepted,
                 declined: declined
             )
+        case .systemAlert:
+            // Should never be called
+            return makeMessageAlertView(
+                with: .mock(),
```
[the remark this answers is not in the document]
rasmustautsglia: fixed [c08f255](https://github.com/salemove/ios-sdk-widgets/pull/968/commits/c08f255d539b691b212f88c3695b2f272a154aad)

GliaWidgets/Sources/AlertManager/Alert/AlertViewController.swift
```diff
@@ -185,6 +134,19 @@ class AlertViewController: UIViewController, Replaceable {
                 accepted: accepted,
                 declined: declined
             )
+        case .systemAlert:
+            // Should never be called
```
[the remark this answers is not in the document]
rasmustautsglia: fixed [c08f255](https://github.com/salemove/ios-sdk-widgets/pull/968/commits/c08f255d539b691b212f88c3695b2f272a154aad)

GliaWidgets/Sources/AlertManager/AlertInputType.swift
EgorovEI: If AlertManager is called to present `mediaUpgrade` from several places with the same `offer` that will help to not show it multiple times. Or I misunderstood something?

GliaWidgetsTests/Sources/AlertManager/AlertManager.Failing.swift
```diff
@@ -0,0 +1,16 @@
+import Foundation
+@testable import GliaWidgets
+
+extension AlertManager {
+    static let failing = AlertManager(
+        environment: .failing,
+        viewFactory: .mock()
```
EgorovEI, line 7: `failing` entities should use only `failing` properties

igorkravchenko: @EgorovEI Ideally yes. But that also depends if it is possible to create such failing entity without drastically rewriting it. For managed  dependencies (structs with closures) it is super easy, but for classes it's not always straightforward. 

@rasmustautsglia Since `AlertManager` is a class (reference type) it's not safe to have static let of this type, because it's internal state may affect other tests. I would recommend changing it to function instead, to always return new instance of the class. Also if viewFactory can't be easily adjusted to be failing, I would leave it as non-default parameter, like so:
```
static func failing(viewFactory: ViewFactory, environment: Environment = .failing) -> AlertManager {
...
```

igorkravchenko (commented):
Left comment about static failing instance.

GliaWidgets/Sources/AlertManager/AlertInputType.swift
rasmustautsglia: `mediaUpgrade` can only be called to present via interactor which is technically one place. The idea of filtering some alerts out is for all the other alert types, but more specifically for errors, such as 409 error.

EgorovEI (approved):
LGTM

yurii-glia (approved):

PR #974, opened 2024-06-12
Improve environment initialization logic

rasmustautsglia:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-3395

**What was solved?**
Currently, Environments have been initialized in views, models, coordinators etc. This takes quite a bit of space and makes the entire codebase hard to read. To improve that, environment initialization has been moved to each corresponding Environment object, into a static method. 

Benefits
1.  The files we use the most, are shorter and easier to read, with thousands of lines moved to less used files.
2. By observing the environment, we can clearly see where it is called from - better overview

NOTE: There are practically no changes to the logic, just code being moved around

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**


igorkravchenko (commented):
There are some merge conflicts. Overall looks fine to me.

yurii-glia:
and rebase is required, because of 44 commits in the PR.

PR #995, opened 2024-06-27
[HOLD] Show bubble when engagement is restored

igorkravchenko:
**Jira issue:**
https://glia.atlassian.net/browse/MOB-2648

**What was solved?**
Show bubble when engagement is restored to align behaviour with Android Widgets.
A few points to pay attention to:
- After configuration if there's ongoing engagement, bubble view will show up because engagement gets restored
In order to align the behaviour with Android, if there's existing engagement we resume it when `Glia.sharedInstance.startEngagement(engagementKind:in:features:sceneProvider:)` is called.
- Now we also provide `features` parameter with `Glia.sharedInstance.configure` method.

Also #1011  was added into this PR since it makes changes to configuration as well.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**


EgorovEI (approved):

igorkravchenko:
!squash

GliaWidgets/Public/Glia/Glia+RestoreEngagement.swift
```diff
@@ -0,0 +1,61 @@
...
+        // be appeared again.
+        createdInteractor.skipLiveObservationConfirmations = true // TODO: snack bar has to be shown, but not dialog
+
+        // Apply company name to theme and get the modified theme.
+        let modifiedTheme = applyCompanyName(using: configuration, theme: theme)
+
+        let viewFactory = ViewFactory(
+            with: modifiedTheme,
+            messageRenderer: messageRenderer,
+            environment: .init(
```
rasmustautsglia, line 25: Can environment be initialized in a static method within `viewFactory` environment's extension?

igorkravchenko: Updated to use it 07047ccf370e05a605a583b15d64237dd3d6e307.

igorkravchenko: Addressed this in #996 via a02dcc8b22d9355ca40af7e846e873106bf90f18.

rasmustautsglia (approved):

PR #1005, opened 2024-07-22
Fix auto-enqueuing for authenticated visitor

EgorovEI:
**Jira issue:**
MOB-3442

**What was solved?**
Previously, if an authenticated visitor has chat history, enqueuing was postponed until a visitor sends a message. This behaviour was broken after adding LO acknowledge.
This commit fixes it.
So now:
- unauthenticated visitors, if LO acknowledge is enabled, will be enqueued once they accept LO acknowledge
- unauthenticated visitors, if LO acknowledge is disabled, will be automatically enqueued
- authenticated visitors, if they do not have chat history and LO acknowledge is disabled, will be enqueued automatically
- authenticated visitors, if they do not have chat history and if LO acknowledge is enabled, will be enqueued once they accept LO acknowledge
- authenticated visitors, if LO acknowledge is enabled and if they have chat history, will be enqueued only after sending a message and accepting LO acknowledge
- authenticated visitors, if LO acknowledge is disabled and if they have chat history, will be enqueued after sending a message

**Release notes:**

 - [ ] Feature
 - [ ] Ignore
 **We should explicitly highlight that the fix will be available in versions 2.4.4 and 2.7.0+**

**Additional info:**

### DIFF
--- a/GliaWidgets/Sources/Coordinators/EngagementCoordinator/EngagementCoordinator.swift
+++ b/GliaWidgets/Sources/Coordinators/EngagementCoordinator/EngagementCoordinator.swift
@@ -71,7 +71,6 @@ class EngagementCoordinator: SubFlowCoordinator, FlowCoordinator {
                 showsCallBubble: false
             )
             engagement = .chat(chatViewController)
-            interactor.state = .enqueueing(.text)
             navigationPresenter.setViewControllers(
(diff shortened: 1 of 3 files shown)


EgorovEI:
!squash

igorkravchenko (approved):

PR #1024, opened 2024-08-28
GVA QuickReply conteiner does not close automatically

rasmustautsglia:
**What was solved?**
If send button is clicked, quick reply container is closed.

MOB-3525
**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**


igorkravchenko (commented):
I think this is a good candidate for unit test.

EgorovEI (approved):

igorkravchenko (approved):

rasmustautsglia:
!squash

PR #1025, opened 2024-08-28
Update bubble visibility through 'startEngagement' and `configure`

igorkravchenko:
**What was solved?**
In TestingApp take into account selected settings regarding showing and hiding bubble from SettingsViewController and add UI for overwrite bubble visibility through `startEngagement`.
Also repurpose Settings to pass `features` to `configure` instead of `startEngagement` method.

MOB-3526

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


rasmustautsglia (approved):

igorkravchenko:
!squash

PR #1040, opened 2024-09-10
Release 2.8.0

EgorovEI:


igorkravchenko (approved):
@EgorovEI Since we do not have SPM integration check flow automated yet, it would be necessary to check it manually.

EgorovEI:
> @EgorovEI Since we do not have SPM integration check flow automated yet, it would be necessary to check it manually.

I manually integrated source code to Cortex app and there is integration issue

PR #1049, opened 2024-09-20
Update Sample integration project with 2.8.0 SDK version

ykyivskyi-gl:
MOB-3598

**What was solved?**
Update Glia Sample Integration project with 2.8.0 SDK version to show usage example of Call Visualizer, Secure Conversation and Authentication. 

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

<div style="display: flex; justify-content: space-around;">
Screenshot 2024-09-20 at 17 24 55
Screenshot 2024-09-20 at 17 24 55
</div>


EgorovEI (approved):

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
```diff
@@ -5,86 +5,120 @@ import SwiftUI
...
+                print("Error has happened: \(error.localizedDescription)")
+            }
+        }
+    }
+    
+    // Show the visitor code to display code and start engagements with live observation
+    func showVisitorCode() {
+        var topViewController: UIViewController?
+        for scene in UIApplication.shared.connectedScenes {
+            if let windowScene = scene as? UIWindowScene {
```
igorkravchenko, line 78: Perhaps nesting can be reduced by using `guard` instead of `if`.

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
```diff
@@ -5,86 +5,120 @@ import SwiftUI
...
+                for window in windowScene.windows {
```
igorkravchenko, line 79: This can take advantage of `where`: `for window in windowScene.windows where window.isKeyWindow` to remove inner `if`. 

I wonder if we need to also break from the loop when key window is found to avoid more iterations.
Also it seems possible to have several key windows at once, that means `topViewController` could be assigned several times. But finding first one key window and break from loop is likely enough for now.

rasmustautsglia (approved):

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
```diff
@@ -98,13 +132,47 @@ class ContentViewPresenter: ObservableObject {
...
+            case .failure(let error):
+                print("Error has happened: \(error.localizedDescription)")
+            }
+        }
+    }
+    
+    // Configure listening of Glia's events
+    func eventListener() {
+        Glia.sharedInstance.onEvent = { event in
+            switch event {
```
igorkravchenko, line 162: I think adding comment here will be helpful - something like "This is an example of how events are handled in Glia Widgets", otherwise this code does not seem to do much.

igorkravchenko (commented):
Left some minor suggestions.

Example/Glia Sample Integration/Presenters/ContentViewPresenter.swift
ykyivskyi-gl: Agree. 

Commit https://github.com/salemove/ios-sdk-widgets/pull/1049/commits/a103b7aba199f55f823e5eb20fe7f39ce13c2feb

igorkravchenko (approved):

ykyivskyi-gl:
!squash

ykyivskyi-gl:
!squash

PR #1054, opened 2024-09-30
Localization and error style for visitor message

EgorovEI:
MOB-3597

**What was solved?**
This PR adds localization, extends VisitorChatMessageStyle and RemoteConfiguration to customize visitor message error state.

Business logic and tests will be added in further PRs.

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**


EgorovEI:
!squash

rasmustautsglia (approved):

PR #1055, opened 2024-09-30
Implement EngagementLauncher logic

ykyivskyi-gl:
**What was solved?**
This PR implements EngagementLauncher logic and splits start engagement logic to two phases(setup and coordinator start). 

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Public/Glia/Glia+EngagementLauncher.swift
+++ b/GliaWidgets/Public/Glia/Glia+EngagementLauncher.swift
@@ -8,7 +8,18 @@ extension Glia {
     ///
     /// - Returns:
     ///   - `EngagementLauncher` instance.
-    public func getEngagementLauncher(queueIds: [String]?) -> EngagementLauncher {
-        .init(queueIds: queueIds)
+    public func getEngagementLauncher(queueIds: [String]?) throws -> EngagementLauncher {
+        let parameters = try getEngagementParameters(in: queueIds ?? [])
+        return try EngagementLauncher { [weak self] engagementKind, sceneProvider in
+            try self?.resolveEngangementState(
+                engagementKind: engagementKind,
+                sceneProvider: sceneProvider,
+                configuration: parameters.configuration,
+                interactor: parameters.interactor,
+                features: parameters.features,
+                viewFactory: parameters.viewFactory,
+                ongoingEngagementMediaStreams: parameters.ongoingEngagementMediaStreams
+            )
+        }
     }
 }
--- a/GliaWidgets/Public/Glia/Glia+StartEngagement.swift
+++ b/GliaWidgets/Public/Glia/Glia+StartEngagement.swift
@@ -24,39 +24,39 @@ extension Glia {
         in queueIds: [String] = [],
         sceneProvider: SceneProvider? = nil
     ) throws {
+        let parameters = try getEngagementParameters(in: queueIds)
+
+        try resolveEngangementState(
+            engagementKind: engagementKind,
+            sceneProvider: sceneProvider,
+            configuration: parameters.configuration,
+            interactor: parameters.interactor,
+            features: parameters.features,
+            viewFactory: parameters.viewFactory,
+            ongoingEngagementMediaStreams: parameters.ongoingEngagementMediaStreams
+        )
+    }
+
+    /// Set up and returns parameters needed to start or restore engagement
+    func getEngagementParameters(in queueIds: [String] = []) throws -> EngagementParameters {
         // In order to align behaviour between platforms,
         // `GliaError.engagementExists` is no longer thrown,
         // instead engagement is getting restored.
-        guard let configuration = self.configuration else { throw GliaError.sdkIsNotConfigured }
+        guard let configuration else {
+            throw GliaError.sdkIsNotConfigured
+        }
+
+        guard let interactor else {
+            loggerPhase.logger.prefixed(Self.self).warning("Interactor is missing")
+            throw GliaError.sdkIsNotConfigured
+        }
 
         // Interactor is initialized during configuration, which means that queueIds need
         // to be set in interactor when startEngagement is called.
-        self.interactor?.setQueuesIds(queueIds)
+        interactor.setQueuesIds(queueIds)
+
         // It is assumed that `features` to be provided from `configure` or via deprecated `startEngagement` method.
         let features = self.features ?? []
-
-        if let engagement = environment.coreSdk.getCurrentEngagement() {
-            if engagement.source == .callVisualizer {
-                throw GliaError.callVisualizerEngagementExists
-            } else {
-                guard let interactor else {
-                    loggerPhase.logger.prefixed(Self.self).warning("Interactor is missing")
-                    return
-                }
-                if let rootCoordinator {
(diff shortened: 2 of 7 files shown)


GliaWidgets/Public/Glia/Glia+StartEngagement.swift
```diff
@@ -24,40 +24,38 @@ extension Glia {
...
+            sceneProvider: sceneProvider,
+            configuration: parameters.configuration,
+            interactor: interactor,
+            features: parameters.features,
+            viewFactory: parameters.viewFactory,
+            ongoingEngagementMediaStreams: parameters.ongoingEngagementMediaStreams
+        )
+    }
+
+    // Set up and returns parameters those are needed to start or restore engagement
```
igorkravchenko, line 45:
```suggestion
    // Sets up and returns parameters needed to start or restore engagement
```

igorkravchenko (commented):
Left some comments about unit tests and code comments.

rasmustautsglia (commented):
Apart from unit tests, looks good

GliaWidgets/Public/Glia/Glia+EngagementLauncher.swift
```diff
@@ -8,7 +8,23 @@ extension Glia {
     ///
     /// - Returns:
     ///   - `EngagementLauncher` instance.
-    public func getEngagementLauncher(queueIds: [String]?) -> EngagementLauncher {
-        .init(queueIds: queueIds)
+    public func getEngagementLauncher(queueIds: [String]?) throws -> EngagementLauncher {
+        guard let interactor else {
```
EgorovEI, line 12: I'd add configuration check before everything, because `interactor` is created on configuration step. So if the SDK is not configured, throwing `internalError` will confuse integrators.
```swift
guard let configuration = self.configuration else { throw GliaError.sdkIsNotConfigured }
```

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
```diff
@@ -24,40 +24,38 @@ extension Glia {
         in queueIds: [String] = [],
         sceneProvider: SceneProvider? = nil
     ) throws {
+        guard let interactor else {
```
EgorovEI, line 27: The first step on startEngagement call should be configuration check, which you moved into `getEngagementParameters`. If we keep as is, then integrator will not receive any exception if the SDK is not configured.
I'd propose to move `guard let interactor else` check inside `getEngagementParameters` method right below configuration check.

```swift
guard let configuration = self.configuration else { throw GliaError.sdkIsNotConfigured }

guard let interactor else {
   loggerPhase.logger.prefixed(Self.self).warning("Interactor is missing")
   throw GliaError.sdkIsNotConfigured
}

File: GliaWidgets/Public/Glia/Glia+StartEngagement.swift [hunk 4]
```diff
@@ -175,4 +207,11 @@ extension Glia {
             onEvent?(.maximized)
         }
     }
+
+    struct EngagementParameters {
```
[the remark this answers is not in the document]
ykyivskyi-gl: Added https://github.com/salemove/ios-sdk-widgets/pull/1055/commits/35b99fb60b80b3827f49db315f47b1c344efdea9

GliaWidgets/Public/Glia/Glia+EngagementLauncher.swift
```diff
@@ -8,7 +8,23 @@ extension Glia {
     ///
     /// - Returns:
     ///   - `EngagementLauncher` instance.
-    public func getEngagementLauncher(queueIds: [String]?) -> EngagementLauncher {
-        .init(queueIds: queueIds)
+    public func getEngagementLauncher(queueIds: [String]?) throws -> EngagementLauncher {
```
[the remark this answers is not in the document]
ykyivskyi-gl: Agree on that and the rest.

Added https://github.com/salemove/ios-sdk-widgets/pull/1055/commits/35b99fb60b80b3827f49db315f47b1c344efdea9

GliaWidgets/Public/Glia/Glia+EngagementLauncher.swift
ykyivskyi-gl: Totally agree.

Fixed https://github.com/salemove/ios-sdk-widgets/pull/1055/commits/35b99fb60b80b3827f49db315f47b1c344efdea9

GliaWidgets/Public/Glia/Glia+StartEngagement.swift
ykyivskyi-gl: Totally agree.

Fixed https://github.com/salemove/ios-sdk-widgets/commit/35b99fb60b80b3827f49db315f47b1c344efdea9

GliaWidgets/Public/Glia/Glia+EngagementLauncher.swift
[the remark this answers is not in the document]
ykyivskyi-gl: Agree, as this method uses functions that used for startEngagement and they are already tested, I added tests for happy flow only.

https://github.com/salemove/ios-sdk-widgets/pull/1055/commits/fbf00f8d1dafd282b3902a9c6b88b2b3bf1a14c2

EgorovEI (approved):

GliaWidgetsTests/Sources/EngagementLauncher/EngagementLauncherTests.swift
```diff
@@ -0,0 +1,50 @@
+import XCTest
+
+@testable import GliaWidgets
+
+final class EngagementLauncherTests: XCTestCase {
+
+    func test_startChat() throws {
+        var chatEngagement: EngagementKind?
+        var engagementLauncher = EngagementLauncher { engagementKind, _ in
```
igorkravchenko, line 9: Why this is instantiated on its own? Does integrator instantiate `EngagementLauncher` like that? As I understand there is dedicated method on the SDK to get `EngagementLauncher` instance, so it seems logically to use it. This will make sure that property `Glia.engagement` has correct value, making the test much deeper. Does it make sense? Also the name of the test can provide more information on what is tested.

igorkravchenko (commented):
Looks good, but I think `EngagementLauncherTests` can be improved by making tests a bit deeper involving property `engagement` on `Glia` instance to be asserted instead of local variables, like  `chatEngagement` etc.

GliaWidgetsTests/Sources/EngagementLauncher/EngagementLauncherTests.swift
ykyivskyi-gl: Good catch. So this tests remains, because EngagementLauncher here as a unit. I added tests for get/start engagement launcher in scope of Glia.

https://github.com/salemove/ios-sdk-widgets/pull/1055/commits/b9452a76d2e2c1b46170b675b1431e9750bcc66d

igorkravchenko (approved):
Looks good to me 👍

ykyivskyi-gl:
!squash

PR #1065, opened 2024-10-09
Remove successfully sent message from pending list to omit resending

igorkravchenko:
Address issue, where pending message is attempted to be sent after operator transferring, by removing pending outgoing message once it was successfully delivered. Previously this resulted in 'Something went wrong' message, because API request was failing for already sent message, but now such request is not even executed.

MOB-3656

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
+++ b/GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
@@ -204,6 +204,11 @@ class ChatViewModel: EngagementViewModel {
                     guard let self else { return }
                     switch result {
                     case let .success(message):
+                        // When pending message is successfully delivered,
+                        // it has to be removed from the `pendingMessages` list, to avoid
+                        // situation where it gets sent again, for example after
+                        // transfer to another operator.
+                        removePendingMessage(by: message.id)
                         self.replace(
                             outgoingMessage,
                             uploads: [],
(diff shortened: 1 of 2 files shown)


EgorovEI (approved):

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -910,6 +915,10 @@ extension ChatViewModel {
     ) -> Bool {
         return section.item(after: row) == nil
     }
+
+    func removePendingMessage(by messageId: String) {
+        pendingMessages.removeAll { $0.payload.messageId.rawValue.uppercased() == messageId.uppercased() }
```
ykyivskyi-gl, line 920: Are we sure that it is safe to alter messageId?

igorkravchenko: @ykyivskyi-gl  Message ID, represented by `UUID` format, is case insensitive, but here it is uppercased for comparison. This approach is also used [here](https://github.com/salemove/ios-sdk-widgets/blob/4fab8cae530bb766337a1df54166eefe062d53b1/GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift#L406), so it is not something invented just now. Also message ID generation is something that we have control on Core SDK side, so I think it is safe to change message ID case for equality check. This is necessary because in some places message ID is represented as `Foundation.UUID` thus uppercased string, but in other places is is a raw `String` when received from backend and decoded as `String`.

ykyivskyi-gl: Got the idea, thanks for explaining!

ykyivskyi-gl (approved):

rasmustautsglia (approved):

PR #1066, opened 2024-10-13
Integrate queues monitor and engagement launcher into EntryWidget

ykyivskyi-gl:
**What was solved?**

This PR integrates queues monitor and engagement launcher into EntryWidget. Unit tests will be added as a part of MOB-3669. 
There are TODOs left to handle errors on View and handle View height update when count of available channels changes(@rasmustautsglia).

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
+++ b/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
@@ -1,4 +1,6 @@
+import Combine
 import Foundation
+import GliaCoreSDK
 import UIKit
 import SwiftUI
 
@@ -7,9 +9,7 @@ public final class EntryWidget: NSObject {
     private var embeddedView: UIView?
     private var queueIds: [String] = []
 
-    // Channels will become dynamic in the subsequent PRs
-    private var channels: [Channel] = [.chat, .audio, .video, .secureMessaging]
-    private let theme: Theme
+    @Published private var channels: [Channel] = []
     private let sizeConstraints: SizeConstraints = .init(
         singleCellHeight: 72,
         singleCellIconSize: 24,
@@ -20,11 +20,23 @@ public final class EntryWidget: NSObject {
         dividerHeight: 1
     )
 
-    init(theme: Theme) {
-        self.theme = theme
+    private let environment: Environment
+
+    private var cancellables = CancelBag()
+
+    init(environment: Environment) {
+        self.environment = environment
+        super.init()
+
+        environment.queuesMonitor.$state
+            .sink(receiveValue: handleQueuesMonitorUpdates(state:))
+            .store(in: &cancellables)
     }
 
     public func show(by presentation: EntryWidget.Presentation) {
+        defer {
+            environment.queuesMonitor.startMonitoring(queuesIds: queueIds)
+        }
         switch presentation {
         case let .sheet(parentViewController):
             showSheet(in: parentViewController)
@@ -38,14 +50,8 @@ public final class EntryWidget: NSObject {
         hostedViewController = nil
         embeddedView?.removeFromSuperview()
         embeddedView = nil
-    }
 
-    func createEntryWidgetView(
-        queueIds: [String],
-        channels: [Channel] = [.chat, .audio, .video]
-    ) {
-        self.queueIds = queueIds
-        self.channels = channels
+        environment.queuesMonitor.stopMonitoring()
     }
 }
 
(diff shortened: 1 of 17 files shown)


GliaWidgets/Sources/EntryWidget/EntryWidget.swift
```diff
@@ -172,7 +172,50 @@ extension EntryWidget: UIViewControllerTransitioningDelegate {
...
+        case .updated(let queues):
+            let availableChannels = resolveAvailableChannels(from: queues)
+            self.channels = availableChannels
+        case .failed(let error):
+            // TODO: Handle error on EntryWidgetView
+            print(error)
+        }
+    }
+
+    func resolveAvailableChannels(from queues: [Queue]) -> [Channel] {
```
igorkravchenko, line 194: As I understand, usage of `Set` makes sure that channels are unique.
Is it necessary to return `Array` here, instead of a `Set` used inside? If conversation to `Array` needed for consumer of the `channels`, it probably makes sense to convert at consumer side, otherwise it is not clear why we convert it here to `Array`.

TestingApp/ViewController/ViewController.swift
```diff
@@ -333,7 +333,7 @@ extension ViewController {
             ) { result in
                 switch result {
                 case .success:
-                    self.entryWidget = Glia.sharedInstance.getEntryWidget(queueIds: [""])
+                    self.entryWidget = try? Glia.sharedInstance.getEntryWidget(queueIds: [""])
```
igorkravchenko, line 336: Maybe error should not be ignored here? We can at least show some feedback regarding it.
```suggestion
               catchingError { [weak self] in
                    self?.entryWidget = try Glia.sharedInstance.getEntryWidget(queueIds: [""])
               }
```

igorkravchenko (commented):
Left some minor questions/suggestions.

rasmustautsglia (approved):

rasmustautsglia:
Besides Igor's comments, looks good to me

GliaWidgets/Sources/EntryWidget/EntryWidget.swift
ykyivskyi-gl: On UI it will be needed to sort the Channels' list, so that's why it's returning `Array`.

igorkravchenko: Does it make sense to return sorted array to preserve some kind of oder?

ykyivskyi-gl: I don't know which order is required on the UI side, I will left a TODO comment so @rasmustautsglia will sort it in a way needed

TestingApp/ViewController/ViewController.swift
ykyivskyi-gl: Agree. https://github.com/salemove/ios-sdk-widgets/pull/1066/commits/7efd8ef93c02f5b31d0bea788c38b7453cc83ed2

GliaWidgets/Sources/QueuesMonitor/QueuesMonitor.Mock.swift
```diff
@@ -0,0 +1,11 @@
+import Foundation
```
EgorovEI, line 1: Please wrap the file content into #if DEBUG preprocessor macros

GliaWidgets/Sources/EntryWidget/EntryWidget.swift
```diff
@@ -20,11 +20,23 @@ public final class EntryWidget: NSObject {
...
+    init(environment: Environment) {
+        self.environment = environment
+        super.init()
+
+        environment.queuesMonitor.$state
+            .sink(receiveValue: handleQueuesMonitorUpdates(state:))
+            .store(in: &cancellables)
     }
 
     public func show(by presentation: EntryWidget.Presentation) {
```
EgorovEI, line 36: Please add documentation for public methods and class itself

ykyivskyi-gl: The documentation should be added along with UI completion by @rasmustautsglia . I was only integrating queues observation logic.

EgorovEI (approved):

ykyivskyi-gl:
!squash

PR #1067, opened 2024-10-14
Add remote configuration for secure conversations 2.0

igorkravchenko:
Add unified remote configuration for SC 2.0. Related styles are to be added naturally with views that need them.

MOB-3671

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration.swift
+++ b/GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration.swift
@@ -222,4 +222,17 @@ extension RemoteConfiguration {
         let title: Text?
         let message: Text?
     }
+
+    struct SecureConversations: Codable {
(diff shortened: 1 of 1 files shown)


rasmustautsglia (approved):

EgorovEI (commented):
You added only new structure, but didn't add it into RemoteConfiguration object and didn't add applying it on top of Theme object. Is it expected?

igorkravchenko:
> You added only new structure, but didn't add it into RemoteConfiguration object and didn't add applying it on top of Theme object. Is it expected?

@EgorovEI  Yes it is expected - this will be added when styles are created. Since styles should be naturally created with view implementation, it does not seem reasonable/obvious to create styles without the views that take advantage of them.

PR #1088, opened 2024-10-23
Fix sending message with attachments

EgorovEI:
**What was solved?**
This PR fixes the case when OutgoingMessage stored in messagesSection did not include attachment object after sending. It led to the situation when the attachment was missed if initial send message request failed.
Now failed-to-send messages with attachments can be successfully resent.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**


EgorovEI:
!squash

rasmustautsglia (approved):

igorkravchenko (approved):

PR #1091, opened 2024-10-24
Hide the "Secure Messaging" if visitor is not authenticated

rasmustautsglia:
This PR checks for authentication status to determine if secure messaging will be displayed in the available media type section in Entry Widget

MOB-3658

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
+++ b/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
@@ -15,7 +15,7 @@ public final class EntryWidget: NSObject {
     private var cancellables = CancelBag()
     private let environment: Environment
 
-    @Published private var viewState: ViewState = .loading
+    @Published private(set) var viewState: ViewState = .loading
 
     private let sizeConstraints: SizeConstraints = .init(
         singleCellHeight: 72,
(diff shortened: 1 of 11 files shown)


GliaWidgets/Sources/EntryWidget/EntryWidget.swift
```diff
@@ -98,6 +98,10 @@ private extension EntryWidget {
                 }
             }
         }
+        if !environment.isAuthenticated() {
```
igorkravchenko, line 101: I think this can be covered with test.

igorkravchenko (commented):
Looks good. I'd add a few tests regarding logic.

GliaWidgets/Sources/EntryWidget/EntryWidget.swift
rasmustautsglia: tests added [059deb7](https://github.com/salemove/ios-sdk-widgets/pull/1091/commits/059deb77207eb3de7d86317697927d8a54e47970)

ykyivskyi-gl (approved):

EgorovEI (approved):

rasmustautsglia:
!squash

PR #1099, opened 2024-10-28
Replace unavailability dialog with banner view on Transcript screen

igorkravchenko:
MOB-3739

**What was solved?**
According to new SC 2.0 design we use 'send message unavailability' view instead of dialog. These changes remove dialog and show banner instead.

**Note**: Message entry placeholder and corresponding buttons look currently active even though they are disabled. It is required to update styling to match the design to reflect disabled state - this is WIP.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/SecureConversations/ChatTranscript/SecureConversations.ChatWithTranscriptModel.swift
+++ b/GliaWidgets/SecureConversations/ChatTranscript/SecureConversations.ChatWithTranscriptModel.swift
@@ -69,6 +69,15 @@ extension SecureConversations.ChatWithTranscriptModel {
         }
     }
 
+    var isSecureConversationsAvailable: Bool {
+        switch self {
+        case .chat:
+            false
+        case let .transcript(model):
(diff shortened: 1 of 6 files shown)


DavDo:
The SC bottom banner should always be visible.

ykyivskyi-gl (approved):

igorkravchenko:
> The SC bottom banner should always be visible.

@DavDo updated this logic in https://github.com/salemove/ios-sdk-widgets/pull/1101

DavDo:
> > The SC bottom banner should always be visible.
> 
> @DavDo updated this logic in #1101

Could you please also update the attached screenshot for consistency?

igorkravchenko:
> Could you please also update the attached screenshot for consistency or add the screenshot to #1101 ?

Updated for both.

rasmustautsglia (approved):

PR #1100, opened 2024-10-28
Improved the Entry Widget Style structures

AndriiHorishniiMOC:
https://glia.atlassian.net/browse/MOB-3745

**What was solved?**
Improved the Entry Widget Style structures
Fixed and added missed attributes.
Added custom redacted placeholder that is not transparent by default.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/SwiftUI/Extensions/View+Extensions.swift
+++ b/GliaWidgets/SwiftUI/Extensions/View+Extensions.swift
@@ -103,4 +103,27 @@ extension SwiftUI.View {
             )
         }
     }
+
+    @ViewBuilder
+    func redactedPlaceholder(
+        _ colorType: ColorType,
+        font: UIFont? = nil,
+        cornerRadius: CGFloat = 2
+    ) -> some View {
+        let padding: CGFloat = if let font = font {
+            font.pointSize * 0.22
+        } else {
+            0
+        }
+
+        self
+            .opacity(0)
+            .overlay(
+                RoundedRectangle(cornerRadius: cornerRadius)
+                    .setColor(Color.baseShade)
+                    .frame(maxWidth: .infinity, maxHeight: .infinity)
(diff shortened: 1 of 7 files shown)


GliaWidgets/SwiftUI/Extensions/View+Extensions.swift
```diff
@@ -103,4 +103,26 @@ extension SwiftUI.View {
...
         }
     }
+
+    @ViewBuilder
+    func redactedPlaceholder(
+        _ colorType: ColorType,
+        font: UIFont? = nil,
+        cornerRadius: CGFloat = 2
+    ) -> some View {
+        let padding: CGFloat = if let font = font {
```
ykyivskyi-gl, line 113: Where `padding` is used?

ykyivskyi-gl (commented):
Please make a commit message in Past Tense:
`Improved the Entry Widget Style structures`
->
`Improve the Entry Widget Style structures`

GliaWidgets/SwiftUI/Extensions/View+Extensions.swift
igorkravchenko: It seems on line 124

igorkravchenko (approved):

GliaWidgets/SwiftUI/Extensions/View+Extensions.swift
AndriiHorishniiMOC: Yes, it is in line 124.
It is because the placeholder is lower than the Text view height. Do make it similar to the original `.redacted(reason: .placeholder)` we should make it a bit lower.

ykyivskyi-gl (approved):

rasmustautsglia (approved):

PR #1107, opened 2024-11-06
Bump up deployment target from iOS 13 to 14

AndriiHorishniiMOC:
**What was solved?**
Bump up deployment target from iOS 13 to 14.

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**


rasmustautsglia (approved):

igorkravchenko (approved):
I think it would be better to reword "Devs want to deprecate iOS 13 support" to "Bump up deployment target from iOS 13 to 14"

PR #1108, opened 2024-11-07
Fix media upgrade dialog configuration

EgorovEI:
MOB-3768

**What was solved?**
After adding AlertManager class configuring alert dialogs was broken. As a result for Audio, 1-way Video, 2-way Video offers the SDK always showed 2-way Video media upgrade dialog. That was only UI configuration issue, accepting the offer resulted in establishing right media type. This commit fixes alert configuration issue.

Acceptance tests will be added in separate task later.

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

 - [x] **Did you add new logging?** We would like the logging between platforms to be similar. Refer to **Logging from iOS SDKs** → **Things to consider for newly added logs** in Confluence for more information.


igorkravchenko (commented):
Is it possible to unit test it? Basically to assert `configuration`?

EgorovEI:
> Is it possible to unit test it? Basically to assert `configuration`?

Added tests bd4b0e2418b9550bf67a020cd2b99c73a3756db3
I also updated acceptance tests ticket to adding units tests for AlertManager and AlertTypeComposer instead.

igorkravchenko (approved):

EgorovEI:
!squash

ykyivskyi-gl (approved):

PR #1113, opened 2024-11-11
Disable send message and pick media buttons instead of hiding them

igorkravchenko:
MOB-3781

**What was solved?**
According to new design for SC 2.0 buttons are not longer hidden but disabled.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


GliaWidgets/Sources/ViewController/Chat/ChatViewController.swift
```diff
@@ -124,7 +124,7 @@ final class ChatViewController: EngagementViewController, PopoverPresenter {
             case .setMessageText(let text):
                 view?.messageEntryView.messageText = text
             case .sendButtonHidden(let hidden):
```
EgorovEI, line 126: Does it make sense to rename `.sendButtonHidden` also?

igorkravchenko: Good point f2aad2984a8e5f46ae00b659a95490ea27254cfa.

EgorovEI (approved):

igorkravchenko:
!squash

PR #1118, opened 2024-11-12
Add logs and telemetry for Entry Widget

rasmustautsglia:
**What was solved?**
This PR adds logs to EntryWidget, and aligning iOS with Android logs. Because logger was added as a dependency to QueueMonitor, it was necessary to move queueMonitor from Glia environment to Glia's class property due to Logger being initialized during GliaSDK initialization. Because of that some changes can be seen in tests as well.

MOB-3481

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Public/Glia/Glia+EngagementLauncher.swift
+++ b/GliaWidgets/Public/Glia/Glia+EngagementLauncher.swift
@@ -10,6 +10,7 @@ extension Glia {
     ///   - `EngagementLauncher` instance.
     public func getEngagementLauncher(queueIds: [String]) throws -> EngagementLauncher {
         let parameters = try getEngagementParameters(in: queueIds)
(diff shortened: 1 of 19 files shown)


igorkravchenko (commented):
Since logger is designed as dependency, we can try to add unit tests, that ensure that expected logs are in place.

rasmustautsglia:
fixed [b1c382a](https://github.com/salemove/ios-sdk-widgets/pull/1118/commits/b1c382a6d185db657ee725e8353dea93096791ab)

igorkravchenko (approved):

EgorovEI (approved):
Please fix introduced linter warnings

rasmustautsglia:
!squash

PR #1122, opened 2024-11-18
SecureConversation business logic

EgorovEI:
MOB-3628

**What was solved?**
This commit introduces next logic:
- If pending secure conversation exists, ChatTranscript screen opens and Leave Current Conversation dialog presents. Then if user press Leave, then EngagementCoordinator changes its engagement kind and starts proper flow: Chat/Audio/Video 

This commit also adds unit tests.

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

### DIFF
--- a/GliaWidgets/Sources/Coordinators/EngagementCoordinator/EngagementCoordinator.swift
+++ b/GliaWidgets/Sources/Coordinators/EngagementCoordinator/EngagementCoordinator.swift
@@ -1,12 +1,41 @@
 import UIKit
 import Foundation
 
+extension EngagementCoordinator {
+    /// `EngagementLaunching` is used to start one of two possible flows:
+    /// - `direct` case is used to start regular engagement flow with single `EngagementKind`. 
+    /// In this case `EngagementCoordinator` starts regular engagement.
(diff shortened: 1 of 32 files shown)


GliaWidgets/Sources/Coordinators/EngagementCoordinator/EngagementCoordinator.swift
```diff
@@ -1,12 +1,33 @@
 import UIKit
 import Foundation
 
+extension EngagementCoordinator {
+    enum EngagementLaunching {
```
igorkravchenko, line 5: I think it is worth documenting this enum for clarity.

igorkravchenko (commented):
I would add some documentation to `EngagementLaunching` type.

GliaWidgets/Sources/Coordinators/EngagementCoordinator/EngagementCoordinator.swift
EgorovEI: Added 67482f9aa893c5192af9423c2cfa1e2b107c02d1

igorkravchenko (approved):

rasmustautsglia (approved):

PR #1134, opened 2024-11-28
Fix memory leak in Chat module

EgorovEI:
MOB-3843

**What was solved?**
This PR introduces:
- Fixing strong reference cycle in ChatView;
- DismissManager struct allowing to take UIKit dismiss completion block execution under control;
- new unit test;
- getting rid of expectation usage in EngagementCoordinator tests.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**


EgorovEI:
!squash

igorkravchenko (approved):

ykyivskyi-gl (approved):

PR #1137, opened 2024-11-30
Isolate pending interaction logic in separate model

igorkravchenko:
MOB-3850

**What was solved?**
Introduce 'PendingInteraction' model to determine pending interaction for SC 2.0.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/SecureConversations/SecureConversations.PendingInteraction.swift
+++ b/GliaWidgets/SecureConversations/SecureConversations.PendingInteraction.swift
@@ -1,0 +1,79 @@
+import Combine
+
+extension SecureConversations {
+    final class PendingInteraction: ObservableObject {
+        @Published private(set) var pendingStatus = false
+        @Published private(set) var unreadMessageCount = 0
+        @Published private(set) var hasPendingInteraction = false
+        private let environment: Environment
+        private(set) var pendingStatusCancellationToken: String?
+        private(set) var unreadMessageCountCancellationToken: String?
+
+        init(environment: Environment) {
+            self.environment = environment
+            self.pendingStatusCancellationToken = environment.observePendingSecureConversationsStatus { [weak self] result in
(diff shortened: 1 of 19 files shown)


GliaWidgets/SecureConversations/SecureConversations.PendingInteraction.swift
```diff
@@ -0,0 +1,80 @@
+import Combine
+
+extension SecureConversations {
+    final class PendingInteraction: ObservableObject {
+        private(set) var cancellables: Set<AnyCancellable> = []
```
ykyivskyi-gl, line 5: There as "sugar" typealias CancelBag, let's decide whether to use or remove it at all.

ykyivskyi-gl (approved):
Good job. Left minor comment

GliaWidgets/SecureConversations/SecureConversations.PendingInteraction.swift
```diff
@@ -0,0 +1,79 @@
+import Combine
+
+extension SecureConversations {
+    final class PendingInteraction: ObservableObject {
+        @Published private(set) var pendingStatus = false
+        @Published private(set) var unreadMessageCount = 0
+        @Published private(set) var hasPendingInteraction = false
+        let environment: Environment
```
EgorovEI, line 8: I see that you don't use `environment` property directly. Based on that, does it make sense to make it `private`?

igorkravchenko: Makes sense 1e8962cc29e69274952b0867a44adb75771438d2.

EgorovEI (approved):

rasmustautsglia (approved):

igorkravchenko:
!squash

PR #1146, opened 2024-12-05
Update SC availability taking into account 'transferring' state

igorkravchenko:
MOB-3654

**What was solved?**
Add 'transferring' and 'engagement.capabilities.text == true' to be used for SC availability.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/SecureConversations/SecureConversations.Availability.swift
+++ b/GliaWidgets/SecureConversations/SecureConversations.Availability.swift
@@ -40,8 +40,16 @@ extension SecureConversations {
             let filteredQueues = fetchedQueues.filter(defaultPredicate)
 
             // Check if matched queues match support `messaging` and
-            // have status other than `closed`
+            // have status other than `closed`.
             guard !filteredQueues.isEmpty else {
+                // In case of "transferred SC" we should treat SC as available.
+                if let engagement = environment.getCurrentEngagement(),
+                   engagement.status == .transferring,
+                   engagement.capabilities?.text == true {
+                    completion(.success(.available(.transferred)))
+                    return
+                }
+
                 environment.log.warning("Provided queue IDs do not match with queues that have status other than closed and support messaging.")
                 completion(.success(.unavailable(.emptyQueue)))
                 return
@@ -49,7 +57,7 @@ extension SecureConversations {
             let queueIds = filteredQueues.map { $0.id }
 
             environment.log.info("Secure Messaging is available in queues with IDs: \(queueIds).")
-            completion(.success(.available(queueIds: queueIds)))
+            completion(.success(.available(.queues(queueIds: queueIds))))
         }
 
         private var defaultPredicate: (CoreSdkClient.Queue) -> Bool {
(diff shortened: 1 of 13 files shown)


EgorovEI (approved):

GliaWidgets/SecureConversations/SecureConversations.Availability.swift
```diff
@@ -40,8 +40,16 @@ extension SecureConversations {
...
 
             // Check if matched queues match support `messaging` and
-            // have status other than `closed`
+            // have status other than `closed`.
             guard !filteredQueues.isEmpty else {
+                // In case of "transferred SC" we should treat SC as available.
+                if let engagement = environment.getCurrentEngagement(),
+                   engagement.status == .transferring,
+                   engagement.capabilities?.text == true {
+                    completion(.success(.available(queueIds: [])))
```
ykyivskyi-gl, line 49: is it intentional to pass an empty array of `queueIds`?

ykyivskyi-gl (approved):

GliaWidgets/SecureConversations/SecureConversations.Availability.swift
igorkravchenko: Yes, because in case of transferring there are no queues with messaging. This is also visible from the check above:
```
 guard !filteredQueues.isEmpty else {
```

igorkravchenko: @EgorovEI  @ykyivskyi-gl @rasmustautsglia  Perhaps it make sense to return different result case for clarity?

EgorovEI: Sure, having dedicated enum case for this case would be much more understandable

igorkravchenko: @ykyivskyi-gl @EgorovEI @rasmustautsglia  Added 'transferred' available state 9cdb97d94df8155f86048825625e93529756582a.

EgorovEI (approved):

rasmustautsglia (approved):

GliaWidgets/SecureConversations/SecureConversations.Availability.swift
```diff
@@ -85,6 +98,15 @@ extension SecureConversations.Availability {
                 return false
             }
         }
+
+        var isAvailable: Bool {
+            switch self {
```
ykyivskyi-gl, line 103: Just a matter of readability for me, `self == .available` looks a bit straightforward.

ykyivskyi-gl (approved):
Looks good. Just an optional comment,

GliaWidgets/SecureConversations/SecureConversations.Availability.swift
igorkravchenko: @ykyivskyi-gl  It has associated values, so compiler needs honest `switch` there, `self == .available` will not compile.

ykyivskyi-gl: Hm, indeed

igorkravchenko:
!squash

PR #1147, opened 2024-12-06
Handle ongoing engagement in EntryWidget

rasmustautsglia:
**What was solved?**
This PR handles ongoing engagement in EntryWidget for both core engagements and call visualizer.

MOB-3806
**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Public/Glia/Glia+EntryWidget.swift
+++ b/GliaWidgets/Public/Glia/Glia+EntryWidget.swift
@@ -27,6 +27,10 @@ extension Glia {
                 hasPendingInteraction: { [weak self] in
                     guard let self else { return false }
                     return pendingInteraction?.hasPendingInteraction ?? false
+                },
+                currentInteractor: { [weak self] in
+                    guard let self else { return nil }
+                    return interactor
                 }
             )
         )
--- a/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
+++ b/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
@@ -17,9 +17,10 @@ public final class EntryWidget: NSObject {
     private let environment: Environment
     @Published private var unreadSecureMessageCount: Int?
     private(set) var unreadSecureMessageSubscriptionId: String?
-
+    private var engagementCancellable: AnyCancellable?
     @Published var viewState: ViewState = .loading
     @Published private(set) var availableEngagementTypes: [EntryWidget.EngagementType] = []
+    private var ongoingEngagement: Engagement?
 
     // MARK: - Initialization
 
@@ -43,6 +44,13 @@ public final class EntryWidget: NSObject {
         Publishers.CombineLatest(environment.queuesMonitor.$state, $unreadSecureMessageCount)
             .sink(receiveValue: handleQueuesMonitorUpdates(state:unreadSecureMessagesCount:))
             .store(in: &cancellables)
+
+        engagementCancellable = environment.currentInteractor()?.$currentEngagement
+            .sink { [weak self] engagement in
+                guard let self else { return }
+                ongoingEngagement = engagement
+                handleQueuesMonitorUpdates(state: environment.queuesMonitor.state, unreadSecureMessagesCount: unreadSecureMessageCount)
+            }
     }
 
     deinit {
@@ -96,6 +104,8 @@ extension EntryWidget {
         case .loading, .error, .offline:
             // 4 gives the desired fixed size
             mediaTypesCount = 4
+        case .ongoingEngagement:
+            mediaTypesCount = 1
         }
         var appliedHeight: CGFloat = 0
         appliedHeight += configuration.sizeConstraints.sheetHeaderHeight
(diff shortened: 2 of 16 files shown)


GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
```diff
@@ -132,6 +132,7 @@ class EngagementViewModel: CommonEngagementModel {
                 }
                 self.engagementAction?(.showAlert(.operatorEndedEngagement(action: { [weak self] in
                     self?.endSession()
+                    self?.interactor.currentEngagement = nil
```
EgorovEI, line 135: The source of truth for `interactor.currentEngagement` is `coreSdkClient.getCurrentEngagement`. So `interactor` should update `currentEngagement` itself based on CoreSDK events, like engagement ended, media upgrades and so on.
Is it reasonable?

EgorovEI: I guess having something like `currentEngagementProducer` that will emit events on each change for `currentEngagement` in CoreSDK, might be even more useful and then we could get rid of `currentEngagement` property in `Interactor` at all.
@igorkravchenko @rasmustautsglia @ykyivskyi-gl what do you think?

rasmustautsglia: I will answer all your comments here since they are all linked.

Good points. If we can rely on core SDK for current engagement updates, then there is no need for any of this logic indeed. This seems like a good approach @EgorovEI

ykyivskyi-gl: I agree. We should keep updating currentEngagement from CoreSDK only. Let's see if you could use Combine instead of ReactiveSwift for emitting events.

GliaWidgets/Public/Glia/Glia+EntryWidget.swift
```diff
@@ -27,7 +27,8 @@ extension Glia {
                 hasPendingInteraction: { [weak self] in
                     guard let self else { return false }
                     return pendingInteraction.hasPendingInteraction
-                }
+                },
+                currentInteractor: { self.interactor }
```
igorkravchenko, line 31: Does `self` here create retain cycle?

GliaWidgets/Sources/EntryWidget/EntryWidget.swift
```diff
@@ -43,6 +44,13 @@ public final class EntryWidget: NSObject {
         Publishers.CombineLatest(environment.queuesMonitor.$state, $unreadSecureMessageCount)
             .sink(receiveValue: handleQueuesMonitorUpdates(state:unreadSecureMessagesCount:))
             .store(in: &cancellables)
+
+        engagementCancellable = environment.currentInteractor()?.$currentEngagement
+            .sink { [weak self] engagement in
+                guard let self = self else { return }
```
igorkravchenko, line 50:
```suggestion
                guard let self else { return }
```

GliaWidgets/Sources/EntryWidget/EntryWidget.swift
```diff
@@ -43,6 +44,13 @@ public final class EntryWidget: NSObject {
...
+                self.ongoingEngagement = engagement
```
igorkravchenko, line 51:
```suggestion
                ongoingEngagement = engagement
```

GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
igorkravchenko: @EgorovEI do you mean introducing currentEngagementProducer on CoreSDK side?

igorkravchenko (commented):
Left some minor suggestions and comments.

GliaWidgets/Sources/EntryWidget/EntryWidget.swift
rasmustautsglia: fixed [88ccb46](https://github.com/salemove/ios-sdk-widgets/pull/1147/commits/88ccb467e346474ad984ce69e3be3a3e470949fe)

GliaWidgets/Sources/EntryWidget/EntryWidget.swift
rasmustautsglia: fixed [88ccb46](https://github.com/salemove/ios-sdk-widgets/pull/1147/commits/88ccb467e346474ad984ce69e3be3a3e470949fe)

GliaWidgets/Public/Glia/Glia+EntryWidget.swift
rasmustautsglia: fixed [88ccb46](https://github.com/salemove/ios-sdk-widgets/pull/1147/commits/88ccb467e346474ad984ce69e3be3a3e470949fe)

EgorovEI (approved):

igorkravchenko (approved):

PR #1148, opened 2024-12-11
Create PendingInteraction model after configuration

igorkravchenko:
Make sure PendingInteraction model is created after configuration to prevent subscription for unread message count and pending SC status from failure, because 'pubsub' is created only when Core SDK is configured.

MOB-3871

**What was solved?**

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/SecureConversations/SecureConversations.PendingInteraction.swift
+++ b/GliaWidgets/SecureConversations/SecureConversations.PendingInteraction.swift
@@ -9,20 +9,33 @@ extension SecureConversations {
         private(set) var pendingStatusCancellationToken: String?
         private(set) var unreadMessageCountCancellationToken: String?
 
-        init(environment: Environment) {
+        init(environment: Environment) throws {
             self.environment = environment
-            self.pendingStatusCancellationToken = environment.observePendingSecureConversationsStatus { [weak self] result in
+            let pendingStatusCancellationToken = environment.observePendingSecureConversationsStatus { [weak self] result in
                 guard let self else { return }
                 // At this point it is enough to know if there is a pending conversation,
                 // so no need to handle error.
                 pendingStatus = (try? result.get()) ?? false
             }
-            self.unreadMessageCountCancellationToken = environment.observeSecureConversationsUnreadMessageCount { [weak self] result in
+
+            guard pendingStatusCancellationToken != nil else {
+                throw Error.subscriptionFailure(.pendingStatus)
+            }
+
+            self.pendingStatusCancellationToken = pendingStatusCancellationToken
+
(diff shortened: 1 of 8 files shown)


GliaWidgets/SecureConversations/SecureConversations.PendingInteraction.swift
```diff
@@ -29,6 +30,14 @@ extension SecureConversations {
                     hasPending || unreadCount > 0
                 }
                 .assign(to: &$hasPendingInteraction)
+
+            if self.pendingStatusCancellationToken == nil {
+                print("Unable to subscribe to pending status updates")
```
ykyivskyi-gl, line 35: should we use `log` instead `print`?

ykyivskyi-gl (approved):

EgorovEI (approved):

GliaWidgets/SecureConversations/SecureConversations.PendingInteraction.swift
igorkravchenko: It seems to me that this has to be a part of SDK configuration and affect SDK configuration completion result. So instead of adding logs I added error handling 047da9354611452871c37af00c4dd4ed50eb9c20.

GliaWidgets/SecureConversations/SecureConversations.PendingInteraction.swift
```diff
@@ -9,21 +9,34 @@ extension SecureConversations {
         private(set) var pendingStatusCancellationToken: String?
```
EgorovEI, line 9: Since you added error throwing, does it mean that properties storing token can be non-optional?

rasmustautsglia (approved):

GliaWidgets/SecureConversations/SecureConversations.PendingInteraction.swift
igorkravchenko: @EgorovEI  Seems like it is not possible, because `weak self` has to be in capture list of the closure, and because of that subscription tokens must be initialised with some values, either `nil` or empty string.
```
let pendingStatusCancellationToken = environment.observePendingSecureConversationsStatus { [weak self] result in
     // in order to access weak `self`, all of it's value must be initialised.
     // ....
     pendingStatus = (try? result.get()) ?? false
    }
```

EgorovEI: Oh, I see. Then let's leave them optional

igorkravchenko:
!squash

PR #1154, opened 2024-12-17
Create interface for resuming call visualizer

rasmustautsglia:
**What was solved?**

Up until now Call Visualizer screen sharing or video call screen could
only be opened through bubble tap. Entry Widget requires to open Call
Visualizer by a tap on call visualizer media type, if CV session is ongoing. This PR allows to do just that, by providing an interface to
resume the view.

MOB-3887
**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
+++ b/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
@@ -114,6 +114,31 @@ extension EntryWidget {
 
         return appliedHeight
     }
+
+    func mediaTypeSelected(_ mediaTypeItem: MediaTypeItem) {
+        if let configurationAction = configuration.mediaTypeSelected {
+            configurationAction(mediaTypeItem)
+            return
+        }
+        hideViewIfNecessary {
+            do {
+                switch mediaTypeItem.type {
+                case .chat:
+                    try self.environment.engagementLauncher.startChat()
+                case .audio:
+                    try self.environment.engagementLauncher.startAudioCall()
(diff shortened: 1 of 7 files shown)


EgorovEI (approved):

GliaWidgets/Sources/EntryWidget/EntryWidget.swift
```diff
@@ -206,7 +206,7 @@ private extension EntryWidget {
                 case .secureMessaging:
                     try self.environment.engagementLauncher.startSecureMessaging()
                 case .callVisualizer:
-                    break
+                    Glia.sharedInstance.callVisualizer.resume()
```
igorkravchenko, line 209: Why singleton has to be used here?

igorkravchenko (commented):
It is not clear why there's usage of singleton (Glia.sharedInstance)? Such practice tends to lead to untestable code (or hardly testable), so I'd like to make sure that there is a justified reason for it.

GliaWidgets/Sources/EntryWidget/EntryWidget.swift
rasmustautsglia: Callback added [478149b](https://github.com/salemove/ios-sdk-widgets/pull/1154/commits/478149b78108faab8c81a1748697e8301b278675)

igorkravchenko (approved):

rasmustautsglia:
!squash

PR #1161, opened 2024-12-19
Add a state property to determine resume behavior

rasmustautsglia:
**What was solved?**

If video call and screensharing were not active in Call Visualizer session
calling resume() resulted in screen sharing view being opened. An unexpected behavior. To make it much more readable, state property was introduced.

In addition, some code was moved around to make the class more readable.

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/CallVisualizer/Coordinator/CallVisualizer.Coordinator.swift
+++ b/GliaWidgets/Sources/CallVisualizer/Coordinator/CallVisualizer.Coordinator.swift
@@ -8,7 +8,7 @@ extension CallVisualizer {
         init(environment: Environment) {
             self.environment = environment
             self.bubbleView = environment.viewFactory.makeBubbleView()
-
+            self.state = .initial
             bubbleView.tap = { [weak self] in
                 self?.resume()
             }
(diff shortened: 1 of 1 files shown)


ykyivskyi-gl (approved):

GliaWidgets/Sources/CallVisualizer/Coordinator/CallVisualizer.Coordinator.swift
```diff
@@ -153,16 +50,126 @@ extension CallVisualizer {
...
+            handleEngagementRequestOutcomeNil(answer: answer)
+            return
+        }
+        handleEngagementRequestOutcomeTimeout(answer: answer)
+    }
+
+    func handleEngagementRequestOutcomeTimeout(answer: Command<Bool>) {
+        environment.alertManager.dismissCurrentAlert()
+        environment.gcd.mainQueue.asyncAfterDeadline(.now() + 0.5) {
+            // Will be swapped for localized string
```
igorkravchenko, line 115: What about this? Is there a ticket to provide localised strings?

igorkravchenko (commented):
Left one comment about localised string, which is not related probably to these changes, but still it has to be addressed sooner or later.

GliaWidgets/Sources/CallVisualizer/Coordinator/CallVisualizer.Coordinator.swift
rasmustautsglia: Ticket for it has been created and comment updated [ad3bf2e](https://github.com/salemove/ios-sdk-widgets/pull/1161/commits/ad3bf2e94dbeec8044b15a5b2a7be28472e02139)

igorkravchenko (approved):

rasmustautsglia:
!squash

PR #1169, opened 2024-12-24
Fix handling engagement ended flow

EgorovEI:
MOB-3897

**What was solved?**
This PR introduces `endedEngagement` property in Interactor for storing engagement ended by operator to fetch a survey

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

### DIFF
--- a/GliaWidgets/Public/Glia/Glia.swift
+++ b/GliaWidgets/Public/Glia/Glia.swift
@@ -437,10 +437,7 @@ public class Glia {
             return
         }
 
-        interactor?.endSession(
-            success: { completion(.success(())) },
-            failure: { completion(.failure($0)) }
-        )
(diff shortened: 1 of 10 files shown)


ykyivskyi-gl (approved):
LGTM. Just a small note, we keep expanding the state of interactor, seems that it is good candidate for refactoring.

igorkravchenko (approved):

EgorovEI:
> LGTM. Just a small note, we keep expanding the state of interactor, seems that it is good candidate for refactoring.

@ykyivskyi-gl thanks for bringing this reasonable point. Since refactoring requires additional team discussion and preparations, I would keep it as is for now. Now it works well, but still would be good to refactor it in near future. I created the ticket for that and assigned to me.

PR #1174, opened 2024-12-31
Log ongoing engagement event in Entry Widget

rasmustautsglia:
**What was solved?**
Both platforms now also log when ongoing engagement

MOB-3909
**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
+++ b/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
@@ -148,6 +148,7 @@ private extension EntryWidget {
         unreadSecureMessagesCount: Int?
     ) {
         if let ongoingEngagement {
+            environment.log.info("Preparing items based on ongoing engagement")
             if ongoingEngagement.source == .callVisualizer {
                 viewState = .ongoingEngagement(.callVisualizer)
             } else if ongoingEngagement.source == .coreEngagement {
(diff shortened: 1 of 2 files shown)


EgorovEI (approved):

GliaWidgets/Sources/EntryWidget/EntryWidget.swift
```diff
@@ -148,6 +148,7 @@ private extension EntryWidget {
         unreadSecureMessagesCount: Int?
     ) {
         if let ongoingEngagement {
+            environment.log.info("Preparing items based on ongoing engagement")
```
igorkravchenko, line 151: We can benefit from adding test for this log - I think this is easy to do since logger is a dependency.
Also I'd like to ensure that this log was added to our logging events document to keep consistency with Android.

rasmustautsglia: Why do we need to test this log? What is the benefit?

The log itself is added to the docs, having the exact same wording on both platforms.

igorkravchenko: In my opinion logging is not something to be taken lightly, because it is a way for us to have insight on how things work in outside world. 

The benefit of having at least single test for logging in current case, is ensuring that logs are created only in certain condition - ongoing engagement, that is the logic we are testing. 

As side note - we don't usually test logs, but in this case it is something that is cheap to add and useful.

igorkravchenko: One more thought - if log is not worthy a test, maybe it shouldn't be added at all?

igorkravchenko (commented):
For this case I'd still add at least single test just ti ensure that logging is executed in certain condition. Let me know if you need help with this.

igorkravchenko (approved):

rasmustautsglia:
!squash

PR #1178, opened 2025-01-07
Add snapshot tests for SC's unified UI

ykyivskyi-gl:
**What was solved?**
This PR adds snapshot tests for SC's unified UI. Related PR https://github.com/salemove/ios-widgets-snapshots/pull/88

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/SnapshotTests/SnapshotTestCase.swift
+++ b/SnapshotTests/SnapshotTestCase.swift
@@ -93,6 +93,24 @@ extension SnapshotTestCase {
     }
 }
 
+extension SnapshotTestCase {
+    /// Retrieves local configuration file for testing Unified Customization. Uses `json` file extension.
+    ///
+    /// - parameter fileName: Name of the file that should be retrieved
+    ///
+    /// - Returns: Parsed `RemoteConfiguration` object if file exists locally and `nil` if it does not.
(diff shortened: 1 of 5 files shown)


rasmustautsglia (approved):
I think we should emphasize it somewhere that if Unified Customization schema is changed or updated, then this particular example should also be maintained

SnapshotTests/SnapshotTestCase.swift
```diff
@@ -93,6 +93,19 @@ extension SnapshotTestCase {
     }
 }
 
+extension SnapshotTestCase {
+    func retrieveRemoteConfiguration(_ fileName: String? = "MockConfiguration") -> RemoteConfiguration? {
```
igorkravchenko, line 97: I think it is worth documenting this method, stating that this is used to test remote unified customisation.
Alternatively the method name could be changed to something like `retrieveRemoteConfigurationForUnifiedCustomization`.

igorkravchenko (approved):
Looks good. Left some comment about method name.

ykyivskyi-gl:
!squash

PR #1179, opened 2025-01-08
Do not cancel enqueuing if new enqueue is the same engagement type

rasmustautsglia:
**What was solved?**
Currently, if the visitor is enqueuing and decides to initiate engagement again, then that will cancel their queue ticket and create a new one. This is really bad from a UX perspective because a person would potentially lose a great spot in a long queue, even though their reason was to just resume the existing one.

The new solution would check for the engagement kind and if the current one matches the new one, then instead of canceling the old, the old one is resumed.

There were technical difficulties to achieve that because currently, the SDK converts EngagementKind to simplified mediaType, which makes this upgrade impossible. This PR improves the approach significantly and allows the SDK to rely on EngagementKind as long as possible and only use mediaType when absolutely needed. This PR lays a pathway forward to continue removing mediaType from places where EngagementKind is sufficient.

MOB-3931
**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
+++ b/GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
@@ -61,6 +61,14 @@ extension Glia {
         viewFactory: ViewFactory,
         ongoingEngagementMediaStreams: Engagement.Media?
     ) throws {
+        /// If during enqueued state the visitor initiates another engagement, we avoid cancelling the queue
+        /// ticket and adding a new one, by monitoring the new engagement kind. If the engagement kind matches
+        /// the current enqueued engagement kind, then we resume the old, and do not start a new one
+        if case let .enqueued(_, enqueudEngagementKind) = interactor.state, enqueudEngagementKind == engagementKind, let rootCoordinator {
+            rootCoordinator.maximize()
+            return
+        }
+
         if let engagement = environment.coreSdk.getCurrentEngagement() {
             if engagement.source == .callVisualizer {
                 throw GliaError.callVisualizerEngagementExists
--- a/GliaWidgets/Public/Glia/Glia.swift
+++ b/GliaWidgets/Public/Glia/Glia.swift
@@ -37,6 +37,10 @@ extension EngagementKind {
             return false
         }
     }
+
+    var mediaType: CoreSdkClient.MediaType {
+        .init(engagementKind: self)
+    }
 }
 
 extension SecureConversations {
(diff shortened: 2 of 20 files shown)


ykyivskyi-gl (approved):
LGTM. Just a minor comment

GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
```diff
@@ -61,6 +61,15 @@ extension Glia {
         viewFactory: ViewFactory,
         ongoingEngagementMediaStreams: Engagement.Media?
     ) throws {
+        switch interactor.state {
```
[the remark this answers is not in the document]
rasmustautsglia: fixed [a7d2822](https://github.com/salemove/ios-sdk-widgets/pull/1179/commits/a7d2822d6ec230e0e63a09a5cf57ca3daf1e9cc0)

GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
```diff
@@ -61,6 +61,11 @@ extension Glia {
         viewFactory: ViewFactory,
         ongoingEngagementMediaStreams: Engagement.Media?
     ) throws {
+        if case let .enqueued(_, enqueudEngagementKind) = interactor.state, enqueudEngagementKind == engagementKind, let rootCoordinator {
```
igorkravchenko, line 64: I think comment here will be helpful to understand the reason why we need to early out is such case.
Ideally to add uni test here as well (if possible).

GliaWidgets/Public/Glia/Glia.swift
```diff
@@ -37,6 +37,21 @@ extension EngagementKind {
             return false
         }
     }
+
+    var mediaType: CoreSdkClient.MediaType {
```
igorkravchenko, line 41: It seems like the logic for enqueueing is no longer dependant on `MediaType`, but on `EngagementKind`?
Maybe this mapping can encapsulated in `CoreSdkClient.MediaType` extension, and used (`MediaType(self)`) here? Like so:
Something like:
```swift
extension CoreSdkClient.MediaType {
    init(_ kind: EngagementKind) {
        switch kind {
            ....
        }
    }
}
```
Also it would make sense to add unit test for it.

igorkravchenko (commented):
Looks good. Left two small suggestions.

GliaWidgets/Public/Glia/Glia.swift
rasmustautsglia: CoreSDKClient.MediaType extended [18d25dc](https://github.com/salemove/ios-sdk-widgets/pull/1179/commits/18d25dcc689f4b46be4c94c785f8db76c0a1ce7b)

rasmustautsglia:
Refactored GliaViewContorller for testing purposes and removed redundant code. Added test [a58f210](https://github.com/salemove/ios-sdk-widgets/pull/1179/commits/a58f210a49bcabd4b311b247f6d88d2bb1c645c7)

igorkravchenko (approved):

rasmustautsglia:
!squash

PR #1184, opened 2025-01-17
Show only one available engagement type in Entry Widget when queueing

rasmustautsglia:
**What was solved?**
Show only one available engagement type in Entry Widget when queueing
When visitor is enqueued, instead of showing all the options in EntryWidget,
We show only the one that the visitor is enqueued to.

MOB-3950
**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
+++ b/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
@@ -13,14 +13,13 @@ public final class EntryWidget: NSObject {
     private var embeddedView: UIView?
     private var queueIds: [String]
     private let configuration: Configuration
-    private var cancellables = CancelBag()
     private let environment: Environment
     @Published private var unreadSecureMessageCount: Int?
     private(set) var unreadSecureMessageSubscriptionId: String?
-    private var engagementCancellable: AnyCancellable?
     @Published var viewState: ViewState = .loading
-    @Published private(set) var availableEngagementTypes: [EntryWidget.EngagementType] = []
     private var ongoingEngagement: Engagement?
+    private var interactorState: InteractorState = .none
+    private var cancellables = CancelBag()
 
     // MARK: - Initialization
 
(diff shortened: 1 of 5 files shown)


GliaWidgets/Sources/EntryWidget/EntryWidget.swift
```diff
@@ -51,6 +52,13 @@ public final class EntryWidget: NSObject {
                 ongoingEngagement = engagement
                 handleQueuesMonitorUpdates(state: environment.queuesMonitor.state, unreadSecureMessagesCount: unreadSecureMessageCount)
             }
+
+        engagementCancellable = environment.currentInteractor()?.$state
```
EgorovEI, line 56: Is it expected to use `engagementCancellable` for two subscriptions?

GliaWidgets/Sources/EntryWidget/EntryWidget.swift
```diff
@@ -51,6 +52,13 @@ public final class EntryWidget: NSObject {
...
+            .sink { [weak self] engagement in
+                guard let self else { return }
+                interactorState = engagement
```
EgorovEI, line 59: Does it make sense if you rename `engagement` param with `state`?

rasmustautsglia: Yes good point. Fixed [2242fbb](https://github.com/salemove/ios-sdk-widgets/pull/1184/commits/2242fbb9a46c5feb768a65a9494eaf123da5dcf5)

GliaWidgets/Sources/Interactor/Interactor.swift
```diff
@@ -65,7 +65,7 @@ class Interactor {
 
     private var observers = [() -> (AnyObject?, EventHandler)]()
 
-    var state: InteractorState = .none {
+    @Published var state: InteractorState = .none {
```
igorkravchenko, line 68: Does this change mean that `didSet` below should be removed and reworked in favour of handling of `state` change by subscription?

ykyivskyi-gl:
!squash

ykyivskyi-gl:
!squash

EgorovEI (approved):

igorkravchenko (approved):

GliaWidgets/Sources/Interactor/Interactor.swift
EgorovEI: I've created a ticket for migrating to subscriptions

PR #1186, opened 2025-01-18
Restore bubble when SC Live Upgrade happens

EgorovEI:
MOB-3951

**What was solved?**
This PR introduces:
- the logic for restoring the bubble for the cases when SC engagement gets:
	- restored from Follow Up;
	- started using Outbound Message feature;
	- started while visitor is on SecureConversation.Confirmation screen
- unit tests

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

### DIFF
--- a/GliaWidgets/Public/Glia/Glia+RestoreEngagement.swift
+++ b/GliaWidgets/Public/Glia/Glia+RestoreEngagement.swift
@@ -1,6 +1,31 @@
 import Foundation
 
 extension Glia {
+    /// Used to restore a bubble for Secure Conversation engagement:
+    /// - started by accepting engagement request;
+    /// - started by Outbound message;
+    /// - restored from Follow Up.
+    func restoreOngoingEngagementIfPresent() {
+        guard let interactor, let configuration else { return }
+
+        guard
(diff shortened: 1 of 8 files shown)


GliaWidgets/Public/Glia/Glia+RestoreEngagement.swift
```diff
@@ -1,6 +1,31 @@
...
+        guard rootCoordinator == nil else { return }
+
+        restoreOngoingEngagement(
+            configuration: configuration,
+            currentEngagement: currentEngagement,
+            interactor: interactor,
+            features: features ?? [],
+            maximize: false
+        )
+        loggerPhase.logger.prefixed(Self.self).info("Engagement was restored")
```
ykyivskyi-gl, line 26: Should it be added to shared sheet with logging events?

ykyivskyi-gl (approved):

GliaWidgets/Public/Glia/Glia+RestoreEngagement.swift
EgorovEI: Thanks! I've added this log into our table

igorkravchenko (approved):

PR #1189, opened 2025-01-20
Restore Video Call Visualizer on enqueuing if it is running

ykyivskyi-gl:
**What was solved?**
This PR restores Video Call Visualizer on enqueuing if it is running

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

### DIFF
--- a/GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
+++ b/GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
@@ -52,7 +52,7 @@ extension Glia {
         features: Features,
         viewFactory: ViewFactory,
         ongoingEngagementMediaStreams: Engagement.Media?
-    ) throws {
+    ) {
         /// If during enqueued state the visitor initiates another engagement, we avoid cancelling the queue
         /// ticket and adding a new one, by monitoring the new engagement kind. If the engagement kind matches
         /// the current enqueued engagement kind, then we resume the old, and do not start a new one
@@ -63,11 +63,11 @@ extension Glia {
 
         if let engagement = environment.coreSdk.getCurrentEngagement() {
             if engagement.source == .callVisualizer {
-                showSnackBar(
-                    with: Localization.EntryWidget.CallVisualizer.description,
-                    style: viewFactory.theme.snackBar
+                handleOngoingCallVisualizer(
+                    from: engagementKind,
+                    engagement: engagement,
+                    snackBarStyle: viewFactory.theme.snackBar
                 )
-                throw GliaError.callVisualizerEngagementExists
             } else {
                 if let rootCoordinator {
                     rootCoordinator.maximize()
@@ -81,8 +81,8 @@ extension Glia {
                     )
                 }
                 loggerPhase.logger.prefixed(Self.self).info("Engagement was restored")
-                return
             }
+            return
         }
 
         // This value can be set to `true` if engagement restoring happened.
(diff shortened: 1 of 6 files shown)


EgorovEI (commented):
From the added logic I see that if there is ongoing CV without video and `resolveEngagementState` is called, snackbar will not be shown. But our diagrams requires this. But maybe I missed something.
It would also be good to add unit test for showing snackbar in this case. If it is not added yet.

ykyivskyi-gl:
@EgorovEI It seems that it's either restoring Video or snackbar. `handleOngoingCallVisualizer` checks if it's video CV so try to restore Video if it is chat/audio/messaging - show snackbar. Here is the test that checks if snackbar is showing https://github.com/salemove/ios-sdk-widgets/blob/86d4e1bae7c91da703f0caa9e67437317c7edba8/GliaWidgetsTests/Sources/Glia/GliaTests%2BEngagementLauncher.swift#L118

Screenshot 2025-01-20 at 17 36 09

EgorovEI:
> @EgorovEI It seems that it's either restoring Video or snackbar. `handleOngoingCallVisualizer` checks if it's video CV so try to restore Video if it is chat/audio/messaging - show snackbar. Here is the test that checks if snackbar is showing
> 
> https://github.com/salemove/ios-sdk-widgets/blob/86d4e1bae7c91da703f0caa9e67437317c7edba8/GliaWidgetsTests/Sources/Glia/GliaTests%2BEngagementLauncher.swift#L118
> 
> Screenshot 2025-01-20 at 17 36 09

`handleOngoingCallVisualizer` checks if `videoCall` engagement kind was initiated by EngagementLauncher, not existing CV video. So If there is non-video CV and `videoCall` is chosen, then it tries to restore CV video screen and fails, because `engagement.mediaStreams.video` is nil.

ykyivskyi-gl:
@EgorovEI basically was right that there is an issue with restoring non video CV engagement with video engagement type selected. During fixing this issue I decided to add a test for every restoring CV engagement case(audio, chat, video engagement) it may be a bit redundant but will keep us safe as there is different behavior for video and non-video engagements.

Thanks Egor!

GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
```diff
@@ -63,11 +63,11 @@ extension Glia {
...
             if engagement.source == .callVisualizer {
-                showSnackBar(
-                    with: Localization.EntryWidget.CallVisualizer.description,
-                    style: viewFactory.theme.snackBar
+                handleOngoingCallVisualizer(
+                    from: engagementKind,
+                    engagement: engagement,
+                    snackBarStyle: viewFactory.theme.snackBar
                 )
-                throw GliaError.callVisualizerEngagementExists
```
ykyivskyi-gl, line 70: As Android team does not throw error here and it there is no sense to do it we decided to align it between platforms.

GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
```diff
@@ -218,6 +218,21 @@ extension Glia {
         }
     }
 
+    private func handleOngoingCallVisualizer(
+        from engagementKind: EngagementKind,
+        engagement: Engagement,
+        snackBarStyle: Theme.SnackBarStyle
+    ) {
+        if engagementKind == .videoCall && engagement.mediaStreams.video != nil {
```
ykyivskyi-gl, line 226: I have not come up with better solution so these checks happen twice: here and in Call Visualizer. It has to be done inside Call Visualizer as it is internal method and may be used somewhere else.

EgorovEI (approved):

GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
igorkravchenko: You can introduce static method on `CallVisualizer` or even free function that will accept  `engagementKind` and `mediaStreamDirection` as parameters, so the logic would be introduced only once and reused for these checks.

igorkravchenko (commented):
One comment about partially sharing logic.

GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
ykyivskyi-gl: The problem here that the logic is similar but not really the same. So Glia checks if restoring was initiated from `videoCall` and there is actually a video media stream, on the other hand CallVisualizer checks if currentEngagement is `callVisualizer` and contains video media stream.

rasmustautsglia (approved):

PR #1196, opened 2025-01-27
Show snack bar for enqueued engagements

ykyivskyi-gl:
**What was solved?**
This PR shows snack bar for enqueued engagements
- Snackbar shown for video/audio enqueueing if chat/SC running
- Snackbar shown for chat/SC enqueueing if video/audio running

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

### DIFF
--- a/GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
+++ b/GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
@@ -61,42 +61,52 @@ extension Glia {
             return
         }
 
-        if let engagement = environment.coreSdk.getCurrentEngagement() {
-            if engagement.source == .callVisualizer {
-                handleOngoingCallVisualizer(
-                    from: engagementKind,
-                    engagement: engagement,
-                    snackBarStyle: viewFactory.theme.snackBar
-                )
-            } else {
-                if let rootCoordinator {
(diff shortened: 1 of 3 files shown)


GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
```diff
@@ -233,6 +243,34 @@ extension Glia {
...
+        shoulShowSnackBar: Bool,
+        ongoingEngagement: Engagement,
+        snackBarStyle: Theme.SnackBarStyle,
+        configuration: Configuration,
+        interactor: Interactor,
+        features: Features
+    ) {
+        if shoulShowSnackBar {
+            showSnackBar(
+                with: Localization.EntryWidget.CallVisualizer.description,
```
EgorovEI, line 256: Now it became not the best localization key, since it's used for other engagement types. Could you create a ticket for future improvement?

EgorovEI (approved):

GliaWidgets/Public/Glia/Glia+EngagementSetup.swift
ykyivskyi-gl: Fair enough https://glia.atlassian.net/browse/MOB-3983

PR #1202, opened 2025-01-29
Transferred SC improvements

EgorovEI:
MOB-3974

**What was solved?**
This PR introduces a logic:
- Don't restore bubble for transferred SC;
- Don't end transferred SC when close button is pressed on Chat screen;
- Showing close button instead of end one in case of transferred SC;
- unit tests;

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**


GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
```diff
@@ -165,6 +165,17 @@ class EngagementViewModel: CommonEngagementModel {
     func setViewAppeared() {
         hasViewAppeared = true
     }
+
+    func handleScreenSharingStatus(_ status: ScreenSharingStatus) {
```
igorkravchenko, line 169: Does this method need separate unit test?

igorkravchenko (approved):
Looks good. Left one question.

rasmustautsglia (approved):

GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
ykyivskyi-gl: this method looks like helping method to move logic from observer closure. I would make this method `private` and test the observing logic so this method would be a black box.

ykyivskyi-gl (approved):

GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
EgorovEI: It can not be `private` because it's called also from Chat and Call viewModels

ykyivskyi-gl: Agree with Igor then 👌

EgorovEI: Added tests bb9b179d62d7ab1d44dbf57034f38686f535d9fe

EgorovEI:
!squash

PR #1208, opened 2025-01-31
Add logs for unknown onEnd engagement action

ykyivskyi-gl:
**What was solved?**
This PR adds logs for unknown onEnd engagement action

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**


igorkravchenko (commented):
Since the log is managed dependency, it would be probably not very hard to add unit test for `case let .unknown(unhandledCase)`. Does it make sense?

ykyivskyi-gl:
!squash

ykyivskyi-gl:
@igorkravchenko I am not a big fan of testing logs but I think you are right in this case, it makes sense to cover `unknown` enum cases logs with tests. Thanks!

EgorovEI (approved):

igorkravchenko (approved):

PR #1215, opened 2025-02-10
Fix Queue update subscription issue

EgorovEI:
MOB-4033

**What was solved?**
This commit introduces:
- single place for subscribing on queue updates - interactor;
- moves setting interactor.queueIds array into getEngagementLauncher method to make subscription as early as possible;
- removes all unnecessary subscription call;
- loading state for QueueMonitor, which prevents subscription duplicates while fetching queues;

ℹ️ Unit tests will be added later in MOB-4046 ticket

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/QueuesMonitor/QueuesMonitor.Live.swift
+++ b/GliaWidgets/Sources/QueuesMonitor/QueuesMonitor.Live.swift
@@ -5,6 +5,7 @@ import Combine
 final class QueuesMonitor {
     enum State {
         case idle
+        case loading
         case updated([Queue])
         case failed(Error)
     }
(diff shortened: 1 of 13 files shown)


GliaWidgets/Sources/QueuesMonitor/QueuesMonitor.Live.swift
```diff
@@ -27,6 +28,34 @@ final class QueuesMonitor {
...
+                environment.logger.error("Setting up queues. Failed to get site queues: \(error)")
+                self.state = .failed(error)
+                completion(.failure(error))
+                return
+            }
+
+            let observedQueues = evaluateQueues(queuesIds: queuesIds, fetchedQueues: queues)
+
+            // Try to prevent affecting state when observeQueuesUpdates is used.
+//            self.state = .updated(observedQueues)
```
ykyivskyi-gl, line 53: Looks like a leftover

ykyivskyi-gl (commented):
👍

rasmustautsglia (approved):

EgorovEI:
!squash

EgorovEI:
!squash

PR #1229, opened 2025-02-13
Improve Transferred SC behaviour

EgorovEI:
MOB-4039

**What was solved?**
This PR adds the logic that skips showing LO indication banner and operator connected view on engaged state if there is transferred SC.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

### DIFF
--- a/GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
+++ b/GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
@@ -179,7 +179,7 @@ class ChatViewModel: EngagementViewModel {
 
             action?(.queue)
             action?(.scrollToBottom(animated: true))
-        case .engaged(let engagedOperator):
+        case .engaged(let engagedOperator) where interactor.currentEngagement?.isTransferredSecureConversation == false:
             environment.log.prefixed(Self.self).info("Operator connected")
             let name = engagedOperator?.firstName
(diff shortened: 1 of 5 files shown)


EgorovEI:
!squash

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -179,7 +179,7 @@ class ChatViewModel: EngagementViewModel {
 
             action?(.queue)
             action?(.scrollToBottom(animated: true))
-        case .engaged(let engagedOperator):
+        case .engaged(let engagedOperator) where interactor.currentEngagement?.isTransferredSecureConversation == false:
```
ykyivskyi-gl, line 182: I suppose that `true` condition is handled somewhere else?

ykyivskyi-gl (approved):
LGTM, just a small question.

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
EgorovEI: If `isTransferredSecureConversation` returns `true`, viewModel should not handle the `.engaged` state. So it falls into `default: break`

Andrii-Horishnii-Glia (approved):

PR #1235, opened 2025-02-18
Fix `replaceExisting` for call media type

EgorovEI:
**What was solved?**
There was the issue that `replaceExisting` was always `false` for Audio/Video enqueueing. Now the value is `true` if Call screen is opened from Leave Current Conversations dialog.

**Release notes:**

 - [ ] Feature
 - [x] Ignore

**Additional info:**

**Screenshots:**


ykyivskyi-gl (approved):

EgorovEI:
!squash

igorkravchenko (approved):

PR #1236, opened 2025-02-19
Fix top banner thread issue

rasmustautsglia:
**What was solved?**
When starting an engagement, then some values were updated from the background thread, which is prohibited. With this PR, UI related changes will happen from the main thread.

MOB-4074
**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
+++ b/GliaWidgets/Sources/EntryWidget/EntryWidget.swift
@@ -42,6 +42,8 @@ public final class EntryWidget: NSObject {
         observeSecureUnreadMessageCount()
 
         Publishers.CombineLatest(environment.queuesMonitor.$state, $unreadSecureMessageCount)
+            // injected combine scheduler will be added here in MOB-4077
+            .receive(on: DispatchQueue.main)
             .sink(receiveValue: { [weak self] in
                 guard let self else { return }
                 handleQueuesMonitorUpdates(state: $0, unreadSecureMessagesCount: $1)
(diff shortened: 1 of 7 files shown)


EgorovEI (approved):

ykyivskyi-gl (approved):

Andrii-Horishnii-Glia (approved):

GliaWidgets/Sources/EntryWidget/EntryWidget.swift
```diff
@@ -42,6 +42,7 @@ public final class EntryWidget: NSObject {
         observeSecureUnreadMessageCount()
 
         Publishers.CombineLatest(environment.queuesMonitor.$state, $unreadSecureMessageCount)
+            .receive(on: DispatchQueue.main)
```
igorkravchenko, line 45: @rasmustautsglia can you add comment here about using here injected combine scheduler?

igorkravchenko (commented):
Receiving value has to be done via injected scheduler. Left suggestion.

GliaWidgets/Sources/EntryWidget/EntryWidget.swift
rasmustautsglia: Added [3126e69](https://github.com/salemove/ios-sdk-widgets/pull/1236/commits/3126e69b109b295c1c17755ab5c5e6705e8c76f1)

igorkravchenko (approved):

PR #1239, opened 2025-02-20
Add Need Live Support banner for Transferred SC

EgorovEI:
MOB-4076

**What was solved?**
Adding Need Live Support banner for Transferred SC

**Release notes:**

 - [x] Feature
 - [ ] Ignore

**Additional info:**

### DIFF
--- a/GliaWidgets/Sources/View/Chat/ChatView.swift
+++ b/GliaWidgets/Sources/View/Chat/ChatView.swift
@@ -1,4 +1,5 @@
 import UIKit
+import Combine
 
 extension ChatView {
     struct Props {
@@ -68,8 +69,9 @@ class ChatView: EngagementView {
     }
 
     // Made internal for Snapshot test purposes
-    @Published var isTopBannerExpanded = false
-    @Published private var isTopBannerHidden = true
+    @Published private(set) var isTopBannerExpanded = false
+    @Published private(set) var isTopBannerHidden = true
+    @Published private(set) var isTopBannerAllowed = false
 
     var props: Props {
         didSet { renderHeaderProps() }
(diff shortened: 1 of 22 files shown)


rasmustautsglia (approved):

EgorovEI:
!squash

GliaWidgets/Sources/View/Chat/ChatView.swift
```diff
@@ -521,11 +523,21 @@ extension ChatView {
         isAnyEngagementTypeAvailable
             .filter { !$0 }
             .assign(to: &$isTopBannerExpanded)
-        isAnyEngagementTypeAvailable
-            .map { !$0 }
+
+        Publishers.CombineLatest(isAnyEngagementTypeAvailable, $isTopBannerAllowed)
+            .map { typesAvailable, isTopBannerAllowed in
+                if !isTopBannerAllowed {
```
igorkravchenko, line 529: I think comment explaining the criteria for top banner to be hidden would be helpful.
```suggestion
                // If top banner is explicitly not allowed, then there's no need to 
                // evaluate available types, and we just early out hiding top banner.
                if !isTopBannerAllowed {
```

GliaWidgets/Sources/View/Chat/ChatView.swift
```diff
@@ -521,11 +523,21 @@ extension ChatView {
...
+                    return true
+                }
+                return !typesAvailable
```
igorkravchenko, line 532:
```suggestion
                // We also hide top banner, if no appropriate media types are available.
                return !typesAvailable
```

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
```diff
@@ -259,6 +268,59 @@ class ChatViewModel: EngagementViewModel {
     }
 }
 
+// MARK: Entry Widget
+extension ChatViewModel {
+    // Set up the Entry Widget configuration inside TranscriptModel,
+    // since it requires passing view model logic to the configuration.
+    private func makeEntryWidgetConfiguration() -> EntryWidget.Configuration {
+        .init(
```
igorkravchenko, line 276: Is this method the same as in `TranscriptModel`? Does it make sense to share its logic in static method of `ChatTranscriptModel`?

igorkravchenko (commented):
Looks good. I wonder if the logic for EW can be shared between the Chat and Transcript models.

GliaWidgets/Sources/View/Chat/ChatView.swift
EgorovEI: Fixed c46bbbd9a62674243927de18305f4c09a895a80b

GliaWidgets/Sources/View/Chat/ChatView.swift
EgorovEI: Fixed c46bbbd9a62674243927de18305f4c09a895a80b

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
igorkravchenko: Can we pass one as parameter?

EgorovEI:
!squash

GliaWidgets/Sources/ViewModel/Chat/ChatViewModel.swift
EgorovEI: Fixed af084838178623075cc04632e58f1824274cbf18

EgorovEI:
!squash

igorkravchenko (approved):

ykyivskyi-gl (approved):

EgorovEI:
!squash

PR #1244, opened 2025-02-27
Update Bitrise stack and steps

EgorovEI:
MOB-4090

**What was solved?**
Bitrise stack was upgraded to macOS 14.5 and Xcode 16.1.x

**Release notes:**

 - [ ] Feature
 - [x] Ignore

### DIFF
--- a/bitrise.yml
+++ b/bitrise.yml
@@ -41,13 +41,12 @@ workflows:
     - certificate-and-profile-installer@1: {}
     - script:
         inputs:
-        - content: |-
-            bundle install
+        - content: bundle install
     - cocoapods-install@2: {}
     - recreate-user-schemes@1:
         inputs:
         - project_path: $BITRISE_PROJECT_PATH
-    - xcode-archive@4:
+    - xcode-archive@5:
         inputs:
         - scheme: $APP_SCHEME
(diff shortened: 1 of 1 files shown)


bitrise.yml
```diff
@@ -238,17 +234,14 @@ workflows:
...
     - cocoapods-install@2: {}
     - bitrise-step-stamp-appicon-with-version-number@1:
         inputs:
         - stamp_version: TestFlight
         - stamp_path_to_icons: TestingApp/Assets.xcassets/AppIcon.appiconset
-    - set-xcode-build-number@1:
-        inputs:
-        - plist_path: TestingApp/Info.plist
-    - xcode-archive@4:
+    - set-xcode-build-number@2: {}
```
Andrii-Horishnii-Glia, line 243: Is it OK not to have a plist path here?

Andrii-Horishnii-Glia (approved):

bitrise.yml
EgorovEI: Yes, it's ok. In new major version `plist_path` was removed completely, so now it extracts info plist under the hood

ykyivskyi-gl (approved):

PR #1263, opened 2025-04-07
Make spi configureLogLevel interface in WidgetSDK

rasmustautsglia:
**What was solved?**
Cortex Financial needs to use configureLogLevel interface, which it is currently accessing directly from CoreSDK. This PR adds a dedicated interface to WidgetSDK, and hidding the interface from other parties.

MOB-4132
**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**


rasmustautsglia:
!squash

EgorovEI (approved):

Andrii-Horishnii-Glia (approved):

PR #1266, opened 2025-04-15
Add all core push notification interfaces to widgetsdk

rasmustautsglia:
**What was solved?**
Additional PushNotification interfaces are needed in WidgetSDK, so that integrators will not have any need to use CoreSDK directly

MOB-4294

**Release notes:**

 - [ ] Feature
 - [ ] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/PushNotifications/PushNotifications.swift
+++ b/GliaWidgets/Sources/PushNotifications/PushNotifications.swift
@@ -1,10 +1,19 @@
 import Foundation
 import GliaCoreSDK
 
+public typealias PushNotificationsType = GliaCoreSDK.PushNotificationsType
+
 public struct PushNotifications {
     let environment: Environment
 
-    /// Application did register for remote notifications with device token
+    /// Notifies the push notifications system that the application has successfully registered for remote notifications.
+    ///
+    /// This method is usually called from the `UIApplicationDelegate`'s registration success callback.
+    ///
+    /// - Parameters:
+    ///   - application: The application instance that registered for notifications.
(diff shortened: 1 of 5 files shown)


GliaWidgets/Sources/PushNotifications/PushNotifications.swift
```diff
@@ -15,17 +24,95 @@ public struct PushNotifications {
...
         environment.coreSdk.pushNotifications.setPushHandler(handler)
     }
 
-    /// The current handler that the SDK is forwarding the
-    /// UNNotificationResponse.actionIdentifier to.
+    /// Retrieves the current push action handler.
+    ///
+    /// - Returns: An optional `PushActionBlock` closure that is currently set to handle push notification actions.
+    ///
     public func pushHandler() -> GliaCoreSDK.PushActionBlock? {
```
EgorovEI, line 60: Do we need to expose `pushHandler` getter?

EgorovEI (approved):

GliaWidgets/Sources/PushNotifications/PushNotifications.swift
rasmustautsglia: You're right, I removed the getter [500bf67](https://github.com/salemove/ios-sdk-widgets/pull/1266/commits/500bf67d08536f45b6eb7bea8f0d78704626cc5e)

rasmustautsglia:
!squash

ykyivskyi-gl (approved):

PR #1277, opened 2025-04-25
Add localization strings for intermediate PN strings

ykyivskyi-gl:
**What was solved?**
Add localization strings for intermediate PN strings.
@1toomas This alert is showed before asking a system alert for Secure Conversation messages Push Notifications during visitor authentication. I deleted the strings that I used as placeholders.

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

### DIFF
--- a/GliaWidgets/Localization.swift
+++ b/GliaWidgets/Localization.swift
@@ -725,10 +725,16 @@ internal enum Localization {
     }
   }
   internal enum PushNotificationsAlert {
-    /// Please allow push notifications to receive new messages alerts.
-    internal static var message: String { Localization.tr("Localizable", "push_notifications_alert.message", fallback: "Please allow push notifications to receive new messages alerts.") }
-    /// Allow Push Notifications
-    internal static var title: String { Localization.tr("Localizable", "push_notifications_alert.title", fallback: "Allow Push Notifications") }
+    /// Allow push notifications so we can message you (and return your messages) in the app.
+    internal static var message: String { Localization.tr("Localizable", "push_notifications_alert.message", fallback: "Allow push notifications so we can message you (and return your messages) in the app.") }
+    /// Stay In Touch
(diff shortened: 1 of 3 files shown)


EgorovEI (approved):

1toomas:
Sorry, I don't get what I should do. Simply approve the deletion?

ykyivskyi-gl:
@1toomas sorry, it was not really clear in first place. We just need 2 strings for intermediate alert(screenshot attached). The strings that were deleted are the ones that I came up with as placeholders. Could you suggest the actual correct strings for this alert?
435172305-cfd06eba-8213-4291-9194-971c1726ee18-3 (1)

ykyivskyi-gl:
@1toomas It should describe that we are asking for pushes to send a notifications about new messages in chat/secure conversation

GliaWidgets/Localization.swift
```diff
@@ -726,9 +726,9 @@ internal enum Localization {
   }
   internal enum PushNotificationsAlert {
     /// Please allow push notifications to receive new messages alerts.
-    internal static var message: String { Localization.tr("Localizable", "push_notifications_alert.message", fallback: "Please allow push notifications to receive new messages alerts.") }
+    internal static var message: String { Localization.tr("Localizable", "push_notifications_alert.message", fallback: "") }
```
1toomas, line 729:
```suggestion
    internal static var message: String { Localization.tr("Localizable", "push_notifications_alert.message", fallback: "Receive notifications about new messages.") }
```

1toomas (commented):
Would it be possible to have the buttons as [Don't Allow] / [Allow] instead of [Cancel] / [Allow]?

1toomas (approved):

ykyivskyi-gl:
!squash

PR #1279, opened 2025-04-28
Configuration.isWhiteLabelApp property not working

dmytrokellerglia:
**What was solved?**

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

 **NOTE**: Modifying this value after the SDK has been configured will have no effect, as the views are retained in memory with the previous configuration. I don't believe there's a valid use case for this workflow, so I haven't made any changes to address it.

### DIFF
--- a/GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration.swift
+++ b/GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration.swift
@@ -13,6 +13,7 @@ public struct RemoteConfiguration: Codable {
     let snackBar: SnackBar?
     let webBrowserScreen: WebView?
     let entryWidget: EntryWidget?
+    let isWhiteLabelApp: Bool?
 }
 
 extension RemoteConfiguration {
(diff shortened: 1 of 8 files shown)


GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration.swift
```diff
@@ -13,6 +13,7 @@ public struct RemoteConfiguration: Codable {
     let snackBar: SnackBar?
     let webBrowserScreen: WebView?
     let entryWidget: EntryWidget?
+    let whiteLabel: Bool?
```
ykyivskyi-gl, line 16: Deprecation message calls it `isWhiteLabelApp`, let's align it.

rasmustautsglia (approved):

dmytrokellerglia:
!squash

ykyivskyi-gl (approved):

EgorovEI (approved):

dmytrokellerglia:
!squash

PR #1283, opened 2025-05-05
IsWhiteLabel not working

dmytrokellerglia:
**What was solved?**

Renamed the RemoteConfig.isWhiteLabelApp property to  whiteLabel as discussed [here](https://github.com/salemove/mobile-sdk-unified/pull/42).

**Release notes:**

 - [] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration.swift
+++ b/GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration.swift
@@ -13,7 +13,7 @@ public struct RemoteConfiguration: Codable {
     let snackBar: SnackBar?
     let webBrowserScreen: WebView?
     let entryWidget: EntryWidget?
-    let isWhiteLabelApp: Bool?
+    let isWhiteLabel: Bool?
 }
 
 extension RemoteConfiguration {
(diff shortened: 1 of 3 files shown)


GliaWidgets/Sources/RemoteConfiguration/RemoteConfiguration.swift
```diff
@@ -13,7 +13,7 @@ public struct RemoteConfiguration: Codable {
     let snackBar: SnackBar?
     let webBrowserScreen: WebView?
     let entryWidget: EntryWidget?
-    let isWhiteLabelApp: Bool?
+    let whiteLabel: Bool?
```
ykyivskyi-gl, line 16: I remember we had a discussion about the naming. Is there a way we can keel isWhiteLabel?

Andrii-Horishnii-Glia: I don't like "App". `isWhiteLabel` looks better to me. But let's discuss it with both mobile teams in Slack.
@dmytrokellerglia please ask both team members to review questions related to both teams in the future.

DavDo: For me there is no difference between `isWhiteLable` and `whiteLabel`, I also was against `App` prefix.

Andrii-Horishnii-Glia (approved):

EgorovEI (approved):

PR #1285, opened 2025-05-07
Add VisitorInfo object to WidgetSDK

rasmustautsglia:
**What was solved?**
To avoid a situation where the integrator needs to import CoreSDK, a widgetSDK VisitorInfo object was created

MOB-4335

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**


GliaWidgets/Public/Glia/Glia.Deprecated.swift
```diff
@@ -4,7 +4,7 @@ import GliaCoreSDK
 extension Glia {
     /// Deprecated, use ``Glia.getVisitorInfo(completion:)`` instead.
     @available(*, deprecated, message: "Deprecated, use ``Glia.getVisitorInfo(completion:)`` instead. ")
-    public func fetchVisitorInfo(completion: @escaping (Result<GliaCore.VisitorInfo, Error>) -> Void) {
+    public func fetchVisitorInfo(completion: @escaping (Result<VisitorInfo, Error>) -> Void) {
```
Andrii-Horishnii-Glia, line 7: If this method is deprecated, why is the type changed? Is it something that has already been released?

Andrii-Horishnii-Glia (approved):

ykyivskyi-gl (approved):

PR #1304, opened 2025-06-04
Change alignment of vertical action buttons in Alert

rasmustautsglia:
**What was solved?**
To align with Android, vertical buttons stack in alert was reordered so that positive button is up top.

Also the alignment of title and message was not dynamic, allowing only central placement. This was also fixed.

MOB-4427

**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**


rasmustautsglia:
Simulator Screenshot - iPhone 15 Pro Max - 2025-06-04 at 11 24 48

dmytrokellerglia (approved):

PR #1308, opened 2025-06-16
Fix End button visibility bug

rasmustautsglia:
**What was solved?**
When a visitor exits enqueued engagement early, the endbutton is briefly shown, which is not intended. This was caused by screensharing handler, that started a flow, in which part of it mistakenly changed the endbutton visibility.

To fix it, we need to check if there is an active engagement, to continue with that part of the flow.

2 scenarios:

1. If there is active engagement, and screensharing is stopped/cleared - show endbutton (expected, and is currently working)
2. If there is no active engagement, and screensharing is stopped/cleared - skip showing endbutton

MOB-4439
**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/ViewController/Call/CallViewController.Mock.swift
+++ b/GliaWidgets/Sources/ViewController/Call/CallViewController.Mock.swift
@@ -115,6 +115,7 @@ extension CallViewController {
             unreadMessages: unreadMessages,
             startWith: startAction
         )
+        viewModel.activeEngagement = .mock()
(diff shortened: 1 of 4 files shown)


EgorovEI (approved):

ykyivskyi-gl (approved):

dmytrokellerglia (approved):

rasmustautsglia:
!squash

PR #1311, opened 2025-06-18
Fix endbutton missing during engagement

rasmustautsglia:
**What was solved?**
Endbutton was missing from live engagement. The new logic will display endbutton appropriately, only during engagement
**Release notes:**

 - [ ] Feature
 - [X] Ignore

**Additional info:**

**Screenshots:**

### DIFF
--- a/GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
+++ b/GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
@@ -182,7 +182,10 @@ class EngagementViewModel: CommonEngagementModel {
         case .stopped where interactor.currentEngagement?.isTransferredSecureConversation == true:
             engagementAction?(.showCloseButton)
         case .stopped:
-            if activeEngagement != nil {
+            switch interactor.state {
+            case .none, .enqueueing, .enqueued, .ended:
+                break
+            case .engaged:
                 engagementAction?(.showEndButton)
             }
         }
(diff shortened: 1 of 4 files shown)


EgorovEI (approved):

GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
```diff
@@ -182,7 +182,10 @@ class EngagementViewModel: CommonEngagementModel {
         case .stopped where interactor.currentEngagement?.isTransferredSecureConversation == true:
             engagementAction?(.showCloseButton)
         case .stopped:
-            if activeEngagement != nil {
+            switch interactor.state {
+            case .none, .enqueueing, .enqueued, .ended:
+                break
```
ykyivskyi-gl, line 187: What if visitor is engaged and then hits end button, isn't there a glitch on a screen between the engagement ended and chat screen disappeared?

rasmustautsglia: Short answer: no

Longer:
Yesterday I merged [this PR](https://github.com/salemove/ios-sdk-widgets/pull/1308), which removed a glitch where, after clicking X and exiting chat, for a brief second, you could see the end button. However, this fix introduced a problem, in which the end button would not be present during an engagement. 

This fix right now sets both scenarios right:
1. If you enqueue and click X, then you exit the chat, and you don't see end button
2. If you are engaged, you see end button, unless screensharing is present

ykyivskyi-gl (approved):
LGTM.

GliaWidgets/Sources/ViewModel/EngagementViewModel.swift
rasmustautsglia: Essentially, for now, there is an inherit issue where screensharing controls the end button in some ways. Once we remove screen sharing, we can rework this logic to make more sense. For example, introducing some computed property or method that controls what is shown in the trailing end of the header.

rasmustautsglia:
!squash

PR #1318, opened 2025-06-25
Remove screen sharing from Call Visualizer

rasmustautsglia:
**What was solved?**
This PR removes any mention, with one exception (marked as TODO), of screen sharing from Call Visualizer.

**Additional info:**

**Screenshots:**


dmytrokellerglia (approved):

rasmustautsglia:
!squash

PR #1322, opened 2025-06-26
Remove screen sharing from alerts

rasmustautsglia:
**What was solved?**
This PR removes screen sharing from alerts.

**Additional info:**

**Screenshots:**


GliaWidgets/Sources/ViewController/Call/CallViewController.swift
```diff
@@ -25,7 +25,7 @@ final class CallViewController: EngagementViewController {
         let view = viewFactory.makeCallView(
             endCmd: .init { [weak self] in self?.viewModel.event(.closeTapped) },
             closeCmd: .init { [weak self] in self?.viewModel.event(.closeTapped) },
-            endScreenshareCmd: .init { [weak self] in self?.viewModel.event(.endScreenSharingTapped) },
+            endScreenshareCmd: .init {},
```
EgorovEI, line 28: `endScreenshareCmd` can be removed from `makeCallView` method. You can also use `.nop` for empty Command (Cmd)

EgorovEI (approved):

rasmustautsglia:
!squash

PR #1339, opened 2025-07-23
Fix automatic enqueueing on Chat screen

EgorovEI:
MOB-4081

**What was solved?**
This PR fixes the issue when visitor does not get enqueued automatically if they press Leave button on Leave Current Conversation dialog.

**Additional info:**


Andrii-Horishnii-Glia (approved):
LGTM

dmytrokellerglia (approved):

PR #1365, opened 2025-08-29
Add shadow to the header in the chatview

rasmustautsglia:
From the user's perspective, the title is accurate. However, from a technical perspective, the shadow is added to the chat scrollview instead, because of the Entry Widget container in Secure Conversations. Innser shadow provides the desired look, so that all the scrolled items disappear underneath the header or header+entryWidget.

MOB-4601


rasmustautsglia:
IMG_7025
IMG_7024

Andrii-Horishnii-Glia (approved):

PR #1367, opened 2025-09-01
Convert secure conversations get unread message count interface to async

rasmustautsglia:
This PR converts Secure Conversations getUnreadMessageCount interface to async. In tests, couple of waitUntils were introduced, and will be removed once fetchChatHistory interface is converted.

MOB-4611

### DIFF
--- a/GliaWidgets/SecureConversations/SecureConversations.MessagesWithUnreadCountLoader.swift
+++ b/GliaWidgets/SecureConversations/SecureConversations.MessagesWithUnreadCountLoader.swift
@@ -17,64 +17,91 @@ extension SecureConversations {
         var environment: Environment
 
         func loadMessagesWithUnreadCount(callback: @escaping (Result<MessagesWithUnreadCount, Error>) -> Void) {
-            typealias ReactiveSwift = CoreSdkClient.ReactiveSwift
-            let unreadCountTimeoutProducer = ReactiveSwift.SignalProducer<Void, Never>(value: ())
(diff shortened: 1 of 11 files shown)


GliaWidgets/SecureConversations/SecureConversations.MessagesWithUnreadCountLoader.swift
```diff
@@ -17,64 +17,91 @@ extension SecureConversations {
...
-
-            unreadCountTimeoutProducer
-                .delay(Self.unreadCountFallbackTimeoutSeconds, on: environment.scheduler)
-                .startWithCompleted {
-                    unreadCountTimeoutSource.input.send(value: .failure(TimeoutError()))
+            func finishIfReady() {
+                stateQueue.async {
+                    guard !didCallback,
+                          let m = messagesResult,
+                          let u = unreadResult
```
EgorovEI, line 29: Can it be done like this?
```suggestion
                          let messagesResult,
                          let unreadResult
```

rasmustautsglia: If it is okay with you, I'd keep it because a PR for fetchChatHistory, that is already ready, removes this workaround logic. `MessagesWithUnreadCountLoader` is much simpler after that. with 2 concurrent async tasks running, not one with async and other with callback.

EgorovEI (approved):

PR #1369, opened 2025-09-02
Scroll to input field when it becomes active

rasmustautsglia:
Scroll to input field when it becomes active
This PR fixes the issue where the keyboard would hide the active input field. There is also additional logic, to bring half-way out of bounds input field from the top to the view, together with the question title.

MOB-4615


EgorovEI:
!squash

PR #1383, opened 2025-09-17
Fix for chat button state after authentication during ongoing engagement

dmytrokellerglia:
**What was solved?**
https://glia.atlassian.net/browse/MOB-4678

Fixed the bug and refactored the hard dependency for Glia.sharedInstance to make it testable.

**Screenshots:**


EgorovEI (commented):
Can you add unit tests for the changes?

PR #1396, opened 2025-09-26
Increase minimal deployment version to iOS 15

rasmustautsglia:
This PR increases the minimum deployment target to iOS 15

MOB-4737


rasmustautsglia:
!squash

ykyivskyi-gl (approved):

PR #1397, opened 2025-09-26
Add flipCamera button telemetry event

EgorovEI:
MOB-4738

**What was solved?**
Added flipCamera button clicked telemetry event

**Additional info:**


GliaWidgets/Sources/CallVisualizer/VideoCall/ViewModel/VideoCallViewModel.Environment.swift
```diff
@@ -16,7 +16,7 @@ extension CallVisualizer.VideoCallViewModel {
         var log: CoreSdkClient.Logger
         var cameraDeviceManager: CoreSdkClient.GetCameraDeviceManageable
         var flipCameraButtonStyle: FlipCameraButtonStyle
-        @Dependency(\.widgets.openTelemetry) var openTelemetry
+        @Dependency(\.widgets.openTelemetry) var openTelemetry: OpenTelemetry
```
ykyivskyi-gl, line 19: Good catch. @dmytrokellerglia @rasmustautsglia please do not forget to explicitly add a type, because it increases the compilation time and missing that in may places will lead to super long compile time. However it could be me who missed it 😆

ykyivskyi-gl (approved):

PR #1539, opened 2026-04-21
Fix operator message containing link not clickable in i os 26

rasmustautsglia:
in iOS 17+, the operator message, containing URL, didn't open. This PR fixes it.


ykyivskyi-gl (approved):
LGTM. As I understand we need this fix ASAP but for improvement we can define those conditions in one place and re-use, also add unit tests. But I am good to merge it now and make those enhancements later

rasmustautsglia:
> LGTM. As I understand we need this fix ASAP but for improvement we can define those conditions in one place and re-use, also add unit tests. But I am good to merge it now and make those enhancements later

I agree

PR #1541, opened 2026-04-23
Update info plist to allow file injection

rasmustautsglia:
Adds Files app document sharing support to GliaTestApp by enabling LSSupportsOpeningDocumentsInPlace and UIFileSharingEnabled.

This lets files pushed into the app’s Documents container, such as sample.pdf used by iOS acceptance tests, appear in the iOS Files picker like they already do for the old TestingApp.


ykyivskyi-gl (approved):

Andrii-Horishnii-Glia (approved):
Could you add more info to the PR description?

PR #1543, opened 2026-05-06
Fix wrong operator header behavior after engagement transfers

Andrii-Horishnii-Glia:
MOB-5159
MOB-5182

**What was solved?**
Each history cell for `.operatorConnected` and `.transferring` now owns its own `ChatConnectViewHost` instance. Previously a single shared instance was re-parented on every `cellForRow` call, causing **UIKit** to silently remove it from the previous cell and leave empty gaps while scrolling.

Also adds name/imageUrl associated values to `ChatItem.Kind.operatorConnected` so each historical cell correctly reflects the operator active at that point in the session, and a regression test covering the multi-transfer scenario.

**Additional info:**

**Screenshots:**


rasmustautsglia (approved):
Screenshot 2026-05-07 at 09 54 24

Other than that, looks great!

Andrii-Horishnii-Glia:
!squash

PR #1565, opened 2026-06-25
Split sendMessage() to send text and each file as separate messages

Andrii-Horishnii-Glia:
MOB-5352

**What was solved?**
- Plumb sendFile through CoreSDKClient and Interactor in Widgets SDK
- sendMessage() emits one text OutgoingMessage (if non-empty) and one per
  uploaded file, calling sendFile() per file instead of bundling everything
  into a single sendMessagePayload
- pending-queue flush and retryMessageSending updated to route file-only
  items through sendFile() and text/legacy items through sendMessagePayload()
- sendOutgoingMessage() helper added to centralise the routing logic
- Updated test to assert both text content and attachment survive separate
  failure paths, using real SendMessagePayload init to avoid messageId collisions

**Additional info:**

**Screenshots:**


ykyivskyi-gl (approved):
Good job on abstracting WCC feature flag in Core

PR #1579, opened 2026-08-28
iOS Chat Attachments Blank Screen

rasmustautsglia:
This PR fixes a blank File attachments screen that could appear after sending a message while an engagement was still being established. Attachment availability is now derived from the complete chat or secure messaging state, including site configuration, engagement connectivity, and the upload limit. The attachment state is refreshed when chat and transcript models are rebound, while empty attachment option lists are no longer presented as a defensive fallback.

MOB-5451


ykyivskyi-gl (approved):

rasmustautsglia:
!squash

